PR #11, opened 2015-03-02
add search engine based on a ilike sql query and a filterable result page...

lebouquetin:
See associated bug - http://tracim.org/workspaces/4/folders/6/threads/54


tracim:
Please fix the tests before to pull request

PR #28, opened 2015-11-05
Update README.md: specific install for >= python3.4

buxx:
Little update of install process for python3.4. 

PS: I'm looking at docker usage for test installation process in "one docker command line": You can take a look [here](https://github.com/buxx/tracim/blob/dev/install/scripts/install/scripts/ubuntu_trusty.sh).


lebouquetin:
Thanks for the pull request. The update is good.
For your information, we'll later update the documentation with python 3.4 as default.

PR #275, opened 2017-07-10
Parameters the storage path of files depot.

algooapy:
3 parameters require careful review:
- `cache_dir`
- `preview_cache_dir`
- `depot_storage_dir`

### DIFF
--- a/tracim/development.ini.base
+++ b/tracim/development.ini.base
@@ -34,8 +34,8 @@ use = egg:tracim
 full_stack = true
 # You can set french as default language by uncommenting next line
 # lang = fr
-cache_dir = %(here)s/data
-preview_cache_dir = /tmp/tracim/cache/previews/
+cache_dir = %(here)s/data/cache
+preview_cache_dir = %(here)s/data/cache/preview/
 beaker.session.key = tracim
 beaker.session.secret = 3283411b-1904-4554-b0e1-883863b53080
 
@@ -221,6 +221,9 @@ wsgidav.config_path = wsgidav.conf
 ## in this case, you have to create your own proxy behind this url.
 ## Do not set http:// prefix.
 # wsgidav.client.base_url = 127.0.0.1:<WSGIDAV_PORT>
+
+## DEPOT
+depot_storage_path = %(here)s/data/depot/
 
 #####
 #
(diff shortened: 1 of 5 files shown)


tracim/development.ini.base
```diff
@@ -34,8 +34,13 @@ use = egg:tracim
 full_stack = true
 # You can set french as default language by uncommenting next line
 # lang = fr
-cache_dir = %(here)s/data
```
tracim: I believe this parameter is a turbogears core parameter. It is probably a bad idea to re-use it four our use except if you know exactly what you are doing. You may broke something without to know it.

tracim/tracim/config/app_cfg.py
```diff
@@ -116,24 +126,25 @@ def start_daemons(manager: DaemonsManager):
...
-duplicated_email_body = l_('''
-We've received a request to reset the password for this account.
-Please click this link to reset your password:
+def configure_depot():
+    """Configure Depot."""
+    depot_storage_name = 'default'
+    depot_storage_path = CFG.get_instance().DEPOT_STORAGE_DIR
+    depot_storage_settings = {'depot.storage_path': depot_storage_path}
+    DepotManager.configure(depot_storage_name,
+                           depot_storage_settings)
```
tracim: The standard indentation we use is with no parameter at parenthesis level : 
```
    DepotManager.configure(
        depot_storage_name,
        depot_storage_settings
    )
```

This gives a presentation robust to refactoring

tracim/tracim/config/app_cfg.py
```diff
@@ -274,12 +347,19 @@ def __init__(self):
             # ContentType.Folder -- Folder is skipped
         ]
 
-        self.RADICALE_SERVER_HOST = tg.config.get('radicale.server.host', '0.0.0.0')
-        self.RADICALE_SERVER_PORT = int(
-            tg.config.get('radicale.server.port', 5232)
+        self.RADICALE_SERVER_HOST = tg.config.get(
+            'radicale.server.host',
+            '0.0.0.0',
```
tracim: listening by default on '0.0.0.0' is dangerous. Please update this with 127.0.0.1 before merge (this is a "by-the-way bugfix")

PR #298, opened 2017-08-02
Fix/249/preview generator stabilization

algooapy:


tracim/tracim/controllers/content.py
```diff
@@ -216,7 +221,11 @@ def get_one(self, file_id, revision_id=None):
             revision_id = file.revision_id
 
         file_path = content_api.get_one_revision_filepath(revision_id)
-        nb_page = preview_manager.get_nb_page(file_path=file_path)
+        nb_page = 0
+        try:
+            nb_page = preview_manager.get_page_nb(file_path=file_path)
+        except PreviewGeneratorException:
+            pass
```
tracim: when try/except/pass, it is required to explain why the exception is silenced. Here you must explain that preview_manager is a "bonus" feature, so if something goes wrong, it should not make tracim fail

tracim/tracim/controllers/content.py
```diff
@@ -225,6 +234,15 @@ def get_one(self, file_id, revision_id=None):
                                  revision_id)
             preview_urls.append(url)
 
+        pdf_ready_mimetypes = []  # type: List[str]
```
tracim: I did move this decision-making part of code into preview-generator. See https://github.com/algoo/preview-generator/commit/205cb460fdbee33d9e701f5024be6a5c22319b5c

So instead, use manager.has_pdf_preview(file_path) which returns True or False according to the availability of PDF previewing"

tracim/tracim/controllers/content.py
```diff
@@ -8,7 +8,11 @@
 
 from cgi import FieldStorage
 from depot.manager import DepotManager
+from preview_generator.exception import PreviewGeneratorException
```
tracim: as soon as you use has_pdf_preview() method of the preview-generator manager, you can remove the imports (see comment below)

tracim/tracim/controllers/content.py
```diff
@@ -235,18 +234,15 @@ def get_one(self, file_id, revision_id=None):
...
-            pdf_available = str(enable_pdf_buttons).lower()
+                preview_manager.has_pdf_preview(file_path=file_path)
         except PreviewGeneratorException as e:
+            # INFO - A.P - Silently intercepts preview exception
+            # As preview generation isn't mandatory, just register it
             logger.debug(self, 'Exception: {}'.format(e.__str__))
-            msg_str = _('Sorry... No PDF downloads for {}: {}')  # type: str
-            msg = msg_str.format(file.file_name, str(e))  # type: str
-            tg.flash(msg, CST.STATUS_ERROR)
+        pdf_available = str(enable_pdf_buttons).lower()
```
tracim: in this case, prefer to write explicitly what you want to have : 

```pdf_available = 'true' if enable_pdf_buttons else 'false'```

It's less dynamic but totally explicity (for example, if the API of preview_generator changes, your code may be broken)

PR #303, opened 2017-08-07
Tests getting notifiable roles

### LINKED ISSUE #168: Notifications are sent to disabled users
When notifications are sent, they are sent to disabled user to. Fix it: disabled user should not be notified.

algooapy:
Function tested:
tracim/tracim/lib/workspace.py:WorkspaceApi.get_notifiable_roles

lebouquetin@8c23f0a has closed #168

### DIFF
--- a/tracim/tracim/tests/library/test_workspace.py
+++ b/tracim/tracim/tests/library/test_workspace.py
@@ -1,11 +1,17 @@
 # -*- coding: utf-8 -*-
-import transaction
 from nose.tools import eq_
 
 from tracim.lib.content import ContentApi
-from tracim.model import DBSession, Content, User
+from tracim.lib.user import UserApi
+from tracim.lib.userworkspace import RoleApi
+from tracim.lib.workspace import WorkspaceApi
+from tracim.model import Content
+from tracim.model import DBSession
+from tracim.model import User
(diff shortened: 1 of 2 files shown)


tracim/tracim/tests/library/test_workspace.py
```diff
@@ -21,3 +27,17 @@ def test_children(self):
         workspace = DBSession.query(Workspace).filter(Workspace.label == 'workspace_1').one()
         folder = ContentApi.get_canonical_query().filter(Content.label == 'folder_1').one()
         eq_([folder, ], list(workspace.get_valid_children()))
+
+    def test_get_notifiable_roles(self):
```
tracim: Generally, its a good reflex to make two tests for testing nominal case ok and nominal case error.

In the current case, this would be named like this: 

```
def test_get_notifiable_roles__ok__nominal_case(self): 
    [...]

def test_get_notifiable_roles__err__nominal_case(self): 
    [...]
```

By the way we have test naming conventions. See http://algoo.trac.im/workspaces/11/folders/1194/pages/2043 section "Tests et nommage des méthodes de test"

This is not for fix now but for information (for future developments).

You can merge the PR

PR #333, opened 2017-08-30
Fix/328/admin workspaces access

algooapy:


tracim/tracim/lib/workspace.py
```diff
@@ -3,6 +3,7 @@
 
 from sqlalchemy.orm import Query
 from tg.i18n import ugettext as _
+from typing import List
```
lebouquetin: common usage is to import typing, so that there is only one import for all typing stuff (and then in code you write `typing.List[Workspace]` instead of `List[Workspace]`

tracim/tracim/tests/library/test_workspace.py
```diff
@@ -41,3 +43,33 @@ def test_get_notifiable_roles(self):
...
+        rapi.create_one(u, w4, UserRoleInWorkspace.READER, off)
+        rapi.create_one(u, w3, UserRoleInWorkspace.CONTRIBUTOR, off)
+        rapi.create_one(u, w2, UserRoleInWorkspace.CONTENT_MANAGER, off)
+        rapi.create_one(u, w1, UserRoleInWorkspace.WORKSPACE_MANAGER, off)
+        eq_([], wapi.get_all_manageable_for_user(user=u))
+        # Checks a manager gets only its own workspaces.
+        u.groups.append(gapi.get_one(Group.TIM_MANAGER))
+        rapi.delete_one(u.user_id, w2.workspace_id)
+        rapi.create_one(u, w2, UserRoleInWorkspace.WORKSPACE_MANAGER, off)
+        eq_([w1, w2], wapi.get_all_manageable_for_user(user=u))
```
lebouquetin: are you sure the items will always be returned in the same order?

algooapy:
Use the workspace api current user instead of a user parameter in function:
```
tracim/tracim/lib/workspace.py:WorkspaceApi.get_all_manageable_for_user()
```

PR #334, opened 2017-09-01
Logs notification metadata

### LINKED ISSUE #278: notification worker must write all sent mail into a file
Currently : we need to investigate large log files with unclear content in order to known if a notification has been sent

Expected : the worker must write to `/tmp/mail-notifications.log` all sent mail. The generated format will be text : `<datetime>|<recipient>|<subject>`

algooapy:
Closes #278


tracim/tracim/lib/notifications.py
```diff
@@ -210,6 +211,24 @@ def _get_sender(self, user: User=None) -> str:
...
+    @staticmethod
+    def _log_notification(
+            recipient: str,
+            subject: str
+    ) -> None:
+        """Log notification metadata."""
+        from tracim.config.app_cfg import CFG
+        log_path = CFG.get_instance().EMAIL_NOTIFICATION_LOG_FILE_PATH
+        if log_path:
+            with open(log_path, 'a') as log_file:
```
lebouquetin: This method works but it is not efficient : file is open on each notification. Example: you update a document which will be notified to 100 users, the file will be open 100 times and closed 100 times.

Add a TODO in the code explaining this efficientless way of work.
(note: the way you fix the bug is enough for the moment)

tracim/tracim/lib/notifications.py
```diff
@@ -210,6 +211,24 @@ def _get_sender(self, user: User=None) -> str:
             email_address = email_address
         )
 
+    @staticmethod
+    def _log_notification(
+            recipient: str,
+            subject: str
+    ) -> None:
```
lebouquetin: According to our live-talk, add a parameter called `action` which would be "CREATED" or "SENT".

Then, this parameter will be set to "CREATED" in the log where you already log data, and add a call to `_log_notification` with action "SENT" in `async_email_sender.send_mail` function.

Let's talk if its not clear for you.

PR #344, opened 2017-09-06
Fix/342/t9n files cleaning

algooapy:


tracim/development.ini.base
```diff
@@ -32,8 +32,6 @@ cookie_secret = 3283411b-1904-4554-b0e1-883863b53080
 [app:main]
 use = egg:tracim
 full_stack = true
-# You can set french as default language by uncommenting next line
-# i18n.lang = fr
```
lebouquetin: Why did you remove the example of setting a default language?

tracim/setup.py
```diff
@@ -1,65 +1,87 @@
...
 except ImportError:
     from ez_setup import use_setuptools
     use_setuptools()
     from setuptools import setup, find_packages
 
+py_version = sys.version_info[:2]
+
+DESCRIPTION = ('Tracim is a plateform software designed to improve '
+               'traceability and productivity in collaborative work.')
+
```
lebouquetin: You moved the description on a global var which is used only once. This is overkill.

By the way, it is a good habit to write 
```
DESCRIPTION = ('Tracim is a plateform software designed to improve '
               'traceability and productivity in collaborative work.')
```

as the following:

```
DESCRIPTION = (
    'Tracim is a plateform software designed to improve '
    'traceability and productivity in collaborative work.'
)
```

The reason is that the first way to write will be broken on name-refactoring (which is a relatively common operation)

tracim/setup.py
```diff
@@ -1,65 +1,87 @@
...
+# quickstarted Options:
+#  sqlalchemy: True
+#  auth:       sqlalchemy
+#  mako:       True
 
-#This is just a work-around for a Python2.7 issue causing
-#interpreter crash at exit when trying to log an info message.
+import sys
+
+from setuptools.command.develop import develop
```
lebouquetin: by the way, maybe you can remove this part of code which is related to python 2.7 (not supported by tracim)

tracim/development.ini.base
algooapy: This is the point also talked about by @buxx in #114:
`i18n.lang` is not a default but a fallback language when not using language detection from browser request header. The [documentation]( http://turbogears.readthedocs.io/en/latest/reference/config-options.html#internationalization) is quite unclear on this parameter, stating two opposite behaviors:
- "works both when language detection is enabled or disabled." followed by
- "If this is set and language detection is dislabled, the application will consider that all gettext wrapped strings must be translated to this language."

Tests shows the later prevails on the former: `i18n.lang` seems to set the language for an application not designed to be internationalized.

As we:
- have an internationalized application,
- built on language detection,
I choosed to simply drop this misleading configuration parameter.

But I may have done wrong :)

lebouquetin: Could you open an issue on github about the documentation and ask for confirmation of what you explain here (that i18n.fr is intended for not-internationalized applications)? https://github.com/TurboGears/tg2/issues

Then, we'll get response to the current question:
- if you're right we merge your PR as it is right now and maybe we contribute to the TG2 documentation
- if the behavior is different, then we'll see what alessandro molina explain to us.
-

algooapy: Given the answer received from @amol-, I was wrong:
https://github.com/TurboGears/tg2docs/issues/42
So i'm going to not only insert back this parameter but uncomment it for good and set it to English.

PR #351, opened 2017-09-12
Sending mail without authentication

gehasia:
Hi,

we should be able to connect to an smtp server even if no login/password are set. This is possible when smtp server allows relay without authentication for some ip adresses/range for example.

I just moved the call to connect() into the try block executed only if login is set. If login is not set, no auth is attempted.


tracim/tracim/lib/email.py
```diff
@@ -73,11 +73,10 @@ def connect(self):
                 try:
                     starttls_result = self._smtp_connection.starttls()
                     logger.debug(self, 'SMTP start TLS result: {}'.format(starttls_result))
+                    login_res = self._smtp_connection.login(self._smtp_config.login, self._smtp_config.password)
```
lebouquetin: Your fix is not working in case we do not want TLS but want login (starttls will fail, then login will never be checked).

Could you put this in a if statement after the try/except so that the login is not executed if empty
something like that:
```
if self._smtp_config.login:
    login_res = self._smtp_connection.login(self._smtp_config.login, self._smtp_config.password)
    logger.debug(self, 'SMTP login result: {}'.format(login_res))
```

lebouquetin:
The Tracim licence is about to change from AGPL to MIT. Are you ok with this ? If you are, I'll merge the PR

gehasia:
@lebouquetin : Ok, i'll do it this evening !
I was not thinking about starttls() raising an exception if it could not start tls... My bad.

tracim/tracim/lib/email.py
gehasia: i pushed the changes.

From my point of view there is a remaining problem : 
It can log at debug level the starttls() or login() exceptions, but this is not handled at the next logger.info() call which log "Connection OK" whatever happens.
I don't know what's your policy for exceptions handling/logging in this kind of case, but it can be pretty disturbing if you don't have debug enabled or if you think that the debug info which is logged is just an info but everything went fine...as it's not true.

In this example my login/password pair was not valid, and the mail was never sent :
```
22:15:01,501 INFO  [tracim] [EmailSender] Connecting from SMTP server srv06.gecoh-regence.local
22:15:06,510 INFO  [tracim] [EmailSender] Connection OK
22:15:06,511 INFO  [tracim] [EmailSender] Sending email to xxx@xxx.xxx
```

gehasia:
i pushed the changes.

From my point of view there is a remaining problem : 
It can log at debug level the starttls() or login() exceptions, but this is not handled at the next logger.info() call which log "Connection OK" whatever happens.
I don't know what's your policy for exceptions handling/logging in this kind of case, but it can be pretty disturbing if you don't have debug enabled or if you think that the debug info which is logged is just an info but everything went fine...as it's not true.

In this example my login/password pair was not valid, and the mail was never sent :
```
22:15:01,501 INFO  [tracim] [EmailSender] Connecting from SMTP server srv06.gecoh-regence.local
22:15:06,510 INFO  [tracim] [EmailSender] Connection OK
22:15:06,511 INFO  [tracim] [EmailSender] Sending email to xxx@xxx.xxx
```

lebouquetin:
You're right, the logging is buggy. Actually, the best would to get a boolean status from starttls and another from login. The log message "connexion ok" would show if both are True, otherwise a "connexion failure" log should appear. Do you want to fix it? In this PR or in another one?

gehasia:
Ok, sounds like i thought to implement it : I'll fix it in this PR this evening or tomorrow !

lebouquetin:
Ok, thanks a lot!

gehasia:
I committed some changes (sorry for the double commit to reset the tracim.mo i added accidentally in first commit). The good one is a31597127e140f5dcb3b5f5a543fce43884ef15c 

Finally i choose to raise the exception if connect() fail because it can't go anywhere else for sending email
The exception is catch in the controller to show a flash warning instead of a flash status_ok. 
I choose to except on Exception() instead of specific smtplib Exceptions because being specific here was not adding any value to the flash message...

Tell me if you want me to revert some changes.

tracim/tracim/lib/email.py
```diff
@@ -67,17 +67,26 @@ def connect(self):
...
                 except Exception as e:
                     logger.debug(self, 'SMTP start TLS error: {}'.format(e.__str__()))
+                    
+            if self._smtp_config.login:
+                try:
+                    login_res = self._smtp_connection.login(self._smtp_config.login, self._smtp_config.password)
+                    logger.debug(self, 'SMTP login result: {}'.format(login_res))
+                except Exception as e:
+                    logger.debug(self, 'SMTP login error: {}'.format(e.__str__()))
+                    raise e
```
lebouquetin: Do not raise again the exception: the email send process is in a background task so raised exceptions would kill the thread.

xandercagexxx:
Its not realy verified.
We don't have tool for that.

Issue #276, opened 2017-07-10
Simplify/Refactor mail notification sending code


tracim:
Currently the code is parameterized : 

- notifications can/may be built synchronously or asynchronously
- built emails can be sent synchronously or asynchronously

Expected : 

- remove useless parameters and complexity
- build emails synchrounously (as it is currently done by default)
- send emails asynchronously (as it is currently done by default)
- make the code more robust (ie manage return values, do not trap exceptions silently)
- write a notification log file which logs:
  - current datetime
  - sender email
  - recipient email
  - subject
  - content id if there is one

PR #365, opened 2017-09-20
Sending mail without authentication (continuation of PR #351)

algooapy:
This is @gehasia work, about connecting to `SMTP` server without authentication, but I was not able to push my few modifications on his repo branch... So this is a twin pull request of #351


gehasia:
Hi sorry i left the pr in the previous state because of being too busy for my own work.

I was raising exception to intercept it in frontend and manage a Warning Flash Message if email was not sent. Did you merge the code too ? If yes, it's not necessary anymore because the exception is not raised anymore.
https://github.com/tracim/tracim/pull/365/commits/a31597127e140f5dcb3b5f5a543fce43884ef15c

I don't see any proper solution to manage a flash message in this case (i'd like to implement it because in the current state everything seems ok but it's not true, if mail has not been sent, user will never be notified of his account creation, and nobody can know it). 
One solution i see, if exception can't be raised at all, would be to return a value (That's mean 3 or 4 methods to change....).
The other would be to catch properly a more specific exception and to do the treatment in the background task too.

algooapy:
Hi @gehasia,
Please, do not be sorry to contribute to `Tracim` :) Your commits are greatly welcomed !
While waiting for the opinion of @lebouquetin, here is my point of view:
- I consider the issue you report us - `SMTP` connection w/o authentication - as solved,
- The use case you describe about notifying user is legit, but:
  - right now, this involves quite a work,
  - we may be able to address it easier once issue #276 fixed.

lebouquetin:
Thanks a lot, @gehasia for the job. I asked @algooapy to fix the last exception raised because the fix was quite simple (just removing one line).

xandercagexxx:
Its not realy verified.
We don't have tool for that.

PR #370, opened 2017-09-27
Fix for Apache to find the path of mako compiled templates.

gehasia:
Apache can't find compiled template path because it will try to serve those files after already appending the basedir to this variable (will be equivalent to %(here)s/%(here)s/data/templates)


gehasia:
Strange : i'm going back to %here%s/data/templates like in master and everything is going fine without any other modification !... I can not reproduce what was leaded me to this modification....

It was a modification i made 2 weeks ago (because nothing was rendering in browser : there were error_log full ...) and forgot about it. 
I just recalling it today and decided to pr it. So i close it because it's not reproducable even for me, even with the previous apache configuration documented (i didn't try the new one)

PR #375, opened 2017-10-19
Documents running the docker image

### LINKED ISSUE #260: Tracim installation with docker image
Expected : 
- have a docker image allowing end users to use tracim
- have a "one line installation" process in the documentation

algooapy:
Closes #260

Do not merge for the moment, the [algoo/tracim docker image](https://hub.docker.com/r/algoo/tracim/) is to be updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,12 @@ The easiest way to test Tracim is to test it through the online demo:
 
 If you want your own dedicated instance but do not want to manage it by yourself, let's contact me at damien.accorsi@free.fr
 
+## Docker ##
+
+In case you prefer using Docker:
+
+    docker run -e DATABASE_TYPE=sqlite -p 80:80 -v /tmp/tracim:/etc/tracimetc -v /tmp/tracimvar:/var/tracim algoo/tracim
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -98,6 +98,12 @@ The easiest way to test Tracim is to test it through the online demo:
 
 If you want your own dedicated instance but do not want to manage it by yourself, let's contact me at damien.accorsi@free.fr
 
+## Docker ##
+
+In case you prefer using Docker:
+
+    docker run -e DATABASE_TYPE=sqlite -p 80:80 -v /tmp/tracim:/etc/tracimetc -v /tmp/tracimvar:/var/tracim algoo/tracim
```
lebouquetin: @algooapy `/tmp/tracimvar` is where files will be stored ? If so, could you change the example in order not to use `/tmp` as root folder ? The reason for that: (at least) on debian, /tmp is purged on every (re)boot... so following the example may result in loosing data

algooapy: Yes of course, I plan to work on the Docker issue today.

Issue #371, opened 2017-09-28
Links are not underlined / not colored in pages


lebouquetin:
Currently:

- The HTML wysiwyg editor detects url and transform them in links.
- When a HTML page is rendered, the link is working but it is presented like normal text

Expected:

- links must be emphasised (ie underlined + colored)

Issue #413, opened 2017-11-02
frontend: Bug wih translation in english of preview generator


PhilippeAccorsi:
**At present:**
(When you are in english version)

![capture d ecran de 2017-11-02 10-42-53](https://user-images.githubusercontent.com/25848257/32319397-b33fb6ca-bfba-11e7-9203-66f6e3bf6ade.png)

**Expected:**

All information need to be translated.

PhilippeAccorsi:
the translation not working, text stay in french.

Issue #398, opened 2017-10-31
Frontend enhancement: change button aspect in workspace homepage


PhilippeAccorsi:
**At present:**

here: http://algoo.trac.im/workspaces/49

When we see the page, we don't know that "members", "an associated calendar" and "in your file explorer (webdav)" are button (**See screenshot bellow**).

![capture d ecran de 2017-10-31 11-56-28](https://user-images.githubusercontent.com/25848257/32232757-0f19307e-be59-11e7-8e1f-beb0408c0224.png)

**Expected:**

changed aspect of button

PR #415, opened 2017-11-02
[#398 #413] more visible link in workspace details + preview generator with better languages handling

Skylsmoi:


Skylsmoi:
also fix #371

PR #494, opened 2017-11-15
Better escaping for mail header with formataddr method

### LINKED ISSUE #490: No email notification send when active user has "@" character in his display name.
Currently :  
-  Notification by email not send when the active user (the user who do a notifiable thing in Tracim) contain a **@** character in his display name.
- Exception catched in RealNotifier about `Sender address rejected: need fully-qualified address`

Expected :
- Notification by email should work regardless of display name of users.

This bug happen because display name appears in "From" header of sended mail. 
The "From" header look like : `a@b via Tracim <noreply@tracim>` which is incorrect.
**@** is a special character and should not appear in atext elem like this one [according to RFC 5322 ](https://tools.ietf.org/html/rfc5322#section-3.2.3).

[One standard way](https://tools.ietf.org/html/rfc5322#section-3.2.4) to support **@** character is by escaping display name part of "From" header with double-quote character: **"** .

inkhey:
Use _formataddr_ from standard lib instead of homemade format for _To_ and _From_ Header in both notification mail and new user mail. Allow to use utf8 character and special character (like @) in "label" part of these header. 
double-quote (**"**) character is also escaped when needed.
Closes #490

### DIFF
--- a/tracim/tracim/lib/email.py
+++ b/tracim/tracim/lib/email.py
@@ -159,9 +159,9 @@ class EmailManager(object):
         message['Subject'] = subject
         message['From'] = formataddr((
             self._global_config.EMAIL_NOTIFICATION_FROM_DEFAULT_LABEL,
-            self._global_config.EMAIL_NOTIFICATION_FROM_EMAIL)
-        )
-        message['To'] = formataddr(('',user.email))
+            self._global_config.EMAIL_NOTIFICATION_FROM_EMAIL,
+        ))
+        message['To'] = formataddr(('', user.email))
 
         text_template_file_path = self._global_config.EMAIL_NOTIFICATION_CREATED_ACCOUNT_TEMPLATE_TEXT  # nopep8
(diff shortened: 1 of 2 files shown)


tracim/tracim/lib/email.py
```diff
@@ -156,11 +157,11 @@ def notify_created_account(
...
         message = MIMEMultipart('alternative')
         message['Subject'] = subject
-        message['From'] = '{0} <{1}>'.format(
+        message['From'] = formataddr((
             self._global_config.EMAIL_NOTIFICATION_FROM_DEFAULT_LABEL,
-            self._global_config.EMAIL_NOTIFICATION_FROM_EMAIL,
+            self._global_config.EMAIL_NOTIFICATION_FROM_EMAIL)
         )
-        message['To'] = user.email
+        message['To'] = formataddr(('',user.email))
```
lebouquetin: you can show the real user name instead of showing email. this is better for user experience, don't you think? (ie `message['To'] = formataddr((user.get_display_name(),user.email)))`

inkhey: Done.

PR #498, opened 2017-11-22
Add Reply with mail Feature

inkhey:
* Add specific header to notification mail in order to allow reply
* Add MailFetcher Daemon to check for new mail and find body (html is preferred)
   * Parse and sanitize body of html mail : drop quotes and remove all unneeded tag/attributes.
   * Parse and sanitize body of text mail : drop quotes and cites, then consider as markdown and them returned as sanitized html.
* Add simple controller to transmit mails to tracim and create new comment to a specific content. 
* All the previous elements working together in order to have a working reply with mail feature.


tracim/tracim/lib/email_fetcher.py
```diff
@@ -0,0 +1,296 @@
...
+        :return: key
+        """
+        username = mail_address.split('@')[0]
+        username_data = username.split('+')
+        if len(username_data) == 2:
+            key = username_data[1]
+        else:
+            key = None
+        return key
+
```
lebouquetin: Code easier to read/understand. Instead of :

```
       if len(username_data) == 2:
           key = username_data[1]
       else:
           key = None
       return key
```

write this:

```
       if len(username_data) == 2:
           return username_data[1]
       return None
```

tracim/tracim/lib/email_fetcher.py
```diff
@@ -0,0 +1,296 @@
...
+
+class MailFetcher(object):
+
+    def __init__(self,
+                 host: str, port: str, user: str, password: str, folder: str,
```
lebouquetin: prefer one parameter on each line

tracim/tracim/lib/email_fetcher.py
```diff
@@ -0,0 +1,296 @@
...
+            time.sleep(self.delay)
+            self._connect()
+            self._fetch()
+            self._notify_tracim()
+            self._disconnect()
+
+    def stop(self) -> None:
+        self._is_active = False
+
+    def _connect(self) -> None:
```
lebouquetin: you should add a parameter `connection_life_duration` which does the following : 

```
       now = datetime.now()
       expiration_datetime = self._connection_creation_datetime() + connection_life_duration
       if self._connection:
           if now < expiration_datetime:
               return
           self._disconnect()
      [...]
```

This allow not to disconnect if _connect() is called twice in the same time

tracim/tracim/lib/email_fetcher.py
```diff
@@ -0,0 +1,296 @@
...
+        self._is_active = True
+
+    def run(self) -> None:
+        while self._is_active:
+            time.sleep(self.delay)
+            self._connect()
+            self._fetch()
+            self._notify_tracim()
+            self._disconnect()
+
```
lebouquetin: The following code: 
```
           self._connect()
           self._fetch()
           self._notify_tracim()
           self._disconnect()
```

would be better written like this:

```
           self._connect()
           emails = self._fetch()
           self._notify_tracim(emails)
           self._disconnect()
```

Let's talk together for explanation about this

tracim/tracim/lib/email_fetcher.py
```diff
@@ -0,0 +1,296 @@
...
+        self.delay = delay
+        self.endpoint = endpoint
+        self.token = token
+
+        self._is_active = True
+
+    def run(self) -> None:
+        while self._is_active:
+            time.sleep(self.delay)
+            self._connect()
```
lebouquetin: self._connect(connection_life_duration=120)  # connection lives 2 minutes

tracim/tracim/lib/email_fetcher.py
```diff
@@ -0,0 +1,296 @@
...
+            # deleted from mailbox during this process) ?
+            rv, data = self._connection.search(None, "(UNSEEN)")
+            if rv == 'OK':
+                # get mail content
+                for num in data[0].split():
+                    rv, data = self._connection.fetch(num, '(RFC822)')
+                    if rv == 'OK':
+                        msg = message_from_bytes(data[0][1])
+                        decodedmsg = DecodedMail(msg)
+                        self._mails.append(decodedmsg)
```
lebouquetin: instead of storing mails in self._mails, why not returning them at the end of _fetch() call ?

tracim/tracim/lib/email_fetcher.py
```diff
@@ -0,0 +1,296 @@
...
+            else:
+                # FIXME : Distinct error from empty mailbox ?
+                pass
+        else:
+            log = 'IMAP : Unable to open mailbox : {}'
+            logger.debug(self, log.format(str(rv)))
+
+    def _notify_tracim(self) -> None:
+        unsended_mail = []
+        while self._mails:
```
lebouquetin: does it really work to do while self._mails and in the loop pop some items?

tracim/tracim/lib/email_fetcher.py
```diff
@@ -0,0 +1,296 @@
...
+                       "content": mail.get_body(),
+                   }}
+            try:
+                r = requests.post(self.endpoint, json=msg)
+                response = r.json()
+                if 'status' not in response:
+                    log = 'bad response: {}'
+                    logger.error(self, log.format(str(response)))
+                else:
+                    if response['status'] == 'ok':
```
lebouquetin: use response.status_code instead of content of the response dictionnary

tracim/tracim/lib/notifications.py
```diff
@@ -285,7 +294,13 @@ def notify_content_update(self, event_actor_id: int, event_content_id: int):
             message['Subject'] = subject
             message['From'] = self._get_sender(user)
             message['To'] = to_addr
-
+            message['Reply-to'] = formataddr(('',replyto_addr))
```
lebouquetin: The good way to reply would be for example to reply to "Damien Accorsi dans l'espace de travail "Intranet" <some@email.address>

inkhey:
Effectively, Candy should not appear. It's seems that my last merging with upstream isn't correct...

tracim/development.ini.base
```diff
@@ -212,6 +214,18 @@ email.processing_mode = sync
 # email.async.redis.port = 6379
 # email.async.redis.db = 0
 
+# Email reply configuration
+email.reply.activated = False
+email.reply.imap.server = your_imap_server
+email.reply.imap.port = 993
```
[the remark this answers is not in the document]
inkhey: You are right, there is not such option in this commit, imaplib allow it so, it should be simple to add it but it's not yet ready.
see : https://github.com/inkhey/tracim/blob/feature/replywithmail/tracim/tracim/lib/email_fetcher.py#L214 .
Same problem with keyfile,certfile which are not yet supported.

tracim/tracim/lib/email_fetcher.py
```diff
@@ -0,0 +1,296 @@
...
+    def get_to_address(self)-> typing.Optional[str]:
+        return parseaddr(self._message['To'])[1]
+
+    def get_first_ref(self) -> typing.Optional[str]:
+        return parseaddr(self._message['References'])[1]
+
+    def get_special_key(self) -> typing.Optional[str]:
+        return self._decode_header(TRACIM_SPECIAL_KEY_HEADER)
+
+    def get_body(self) -> typing.Optional[str]:
```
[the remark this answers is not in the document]
inkhey: Allow to return Str or None object.

tracim/tracim/lib/email_fetcher.py
```diff
@@ -0,0 +1,296 @@
...
+                html_body = body_part.get_payload(decode=True).decode(
+                    charset)
+                body = DecodedMail._parse_html_body(html_body)
+
+        return body
+
+    @staticmethod
+    def _parse_txt_body(txt_body: str):
+        txt_body = EmailReplyParser.parse_reply(txt_body)
+        html_body = markdown.markdown(txt_body)
```
buxx: You convert txt content into html content to be able to use your function `_parse_html_body` or for another reason ? If yes a comment should be great : )

tracim/tracim/lib/email_fetcher.py
```diff
@@ -0,0 +1,300 @@
...
+        if rv == 'OK':
+            # get mails
+            # FIXME - G.M -  2017-11-15 Which files to select as new file ?
+            # Unseen file or All file from a directory (old one should be moved/
+            # deleted from mailbox during this process) ?
+            rv, data = self._connection.search(None, "(UNSEEN)")
+            if rv == 'OK':
+                # get mail content
+                for num in data[0].split():
+                    rv, data = self._connection.fetch(num, '(RFC822)')
```
lebouquetin: Explain here why you use RFC822

inkhey: Example in python doc use it to retrieve all mail: https://docs.python.org/fr/3.5/library/imaplib.html#imap4-example .

lebouquetin: Actually, I expected you to put the answer as a comment in the source code ^ ^

tracim/tracim/lib/daemons.py
```diff
@@ -151,6 +152,42 @@ def append_thread_callback(self, callback: collections.Callable) -> None:
...
+            endpoint=cfg.WEBSITE_BASE_URL + "/events",
+            token=cfg.EMAIL_REPLY_TOKEN,
+        )
+        self._fetcher.run()
+
+    def stop(self):
+        if self._fetcher:
+            self._fetcher.stop()
+
+    def append_thread_callback(self, callback: collections.Callable) -> None:
```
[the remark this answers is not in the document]
inkhey: Done https://github.com/tracim/tracim/issues/500

tracim/tracim/lib/email_fetcher.py
```diff
@@ -0,0 +1,323 @@
...
+            content_dispo = str(part.get('Content-Disposition'))
+            if content_type == CONTENT_TYPE_TEXT_HTML \
+                    and 'attachment' not in content_dispo:
+                return part
+        # check for plain text
+        for part in self._message.walk():
+            content_type = part.get_content_type()
+            content_dispo = str(part.get('Content-Disposition'))
+            if content_type == CONTENT_TYPE_TEXT_PLAIN and 'attachment' \
+                    not in content_dispo:
```
lebouquetin: Instead of

```
         if content_type == CONTENT_TYPE_TEXT_PLAIN and 'attachment' \
                 not in content_dispo:
```

It's more readable to split line on end of condition:

```
         if content_type == CONTENT_TYPE_TEXT_PLAIN \
                  and 'attachment' not in content_dispo:
```

What do you think?

tracim/tracim/lib/email_fetcher.py
```diff
@@ -0,0 +1,323 @@
...
+    def run(self) -> None:
+        while self._is_active:
+            time.sleep(self.delay)
+            try:
+                self._connect()
+                mails = self._fetch()
+                # TODO - G.M -  2017-11-22 retry sending unsended mail
+                # These mails are return by _notify_tracim, flag them with "unseen"
+                # or store them until new _notify_tracim call
+                self._notify_tracim(mails)
```
lebouquetin: Actually, I just saw that notifying tracim and getting "cleaned mail" are done in the same task, which is not a good way to work (it's difficult to understand where is the part of the algorithm in charge of "purging emails" from trash stuff.

Instead of

```
self._connect()
mails = self._fetch()
self._notify_tracim(mails)
self._disconnect()
```

You should have something like this:

```
self._connect()
messages = self._fetch()  # returns Message objects, not DecodedMails
cleaned_mails = [DecodedMail(msg) for msg in messages]
self._notify_tracim(cleaned_mails)
self._disconnect()
```

Why ? Writing the code this way allow to insert other operations without to modify _fetch() (which is in charge of fetching, not in charge of cleaning emails)

Could you add a ticket for a future refactoring? (or fix this before we merge)

PR #506, opened 2017-12-01
Feature/better html email parsing

inkhey:
Add "intelligent" parser who parse html mail body depending on main text, cite and signature parts order.
Now support thunderbird, gmail and outlook.com (non-beta) webmail.


tracim/tracim/tests/library/test_email_body_parser.py
```diff
@@ -0,0 +1,757 @@
+from bs4 import BeautifulSoup
+from nose.tools import raises
+from tracim.lib.email_processing.checkers import HtmlMailQuoteChecker, \
+    HtmlMailSignatureChecker
```
buxx: We require one import per line (i will do the modification)

PR #514, opened 2017-12-12
Filelock in mail fetching + Better IMAP flagging

inkhey:
- Add a lock into MailFetcher when checking new mail to avoid concurrent trouble.
- Flag  as 'Seen' after retrieving element.
- Flag 'Flagged when tracim rest request is ok or  error 400, unflag 'Seen' otherwise.


tracim/tracim/config/app_cfg.py
```diff
@@ -392,6 +392,9 @@ def __init__(self):
             'email.reply.use_txt_parsing',
             True,
         ))
+        self.EMAIL_REPLY_FILELOCK_PATH = tg.config.get(
+            'email.reply.filelock_path',
+        )
```
buxx: default value: `"%(here)s/email_fetcher.lock"` ;)

tracim/tracim/lib/email_fetcher.py
```diff
@@ -185,11 +198,12 @@ def run(self) -> None:
             time.sleep(self.delay)
             try:
                 self._connect()
-                messages = self._fetch()
-                # TODO - G.M -  2017-11-22 retry sending unsended mail
-                # These mails are return by _notify_tracim, flag them with "unseen" # nopep8
-                # or store them until new _notify_tracim call
-                cleaned_mails = [DecodedMail(msg) for msg in messages]
+                with self.lock.acquire(
```
buxx: What appens if lock acquire fail ? filelock raise something ?

tracim/tracim/lib/email_fetcher.py
```diff
@@ -199,6 +213,7 @@ def run(self) -> None:
 
     def stop(self) -> None:
         self._is_active = False
+        del self.lock
```
buxx: You make this del for filelock specific reason ?

tracim/tracim/lib/email_fetcher.py
inkhey: You're right, because of timeout, lock may fail (TimeoutError exception).
I use timeout but i don't think about it the correct way.
I should correct that.

tracim/tracim/lib/email_fetcher.py
inkhey: Just to be sure lock is released, del is same as release(force = True) according to 
https://github.com/benediktschmitt/py-filelock/blob/master/filelock.py .

tracim/development.ini.base
```diff
@@ -228,6 +228,9 @@ email.reply.token = mysecuretoken
 email.reply.check.heartbeat = 60
 email.reply.use_html_parsing = true
 email.reply.use_txt_parsing = true
+# Lockfile path is required for email_reply feature,
+# it's just an empty file use to prevent concurrent access to imap unseen mail
+email.reply.filelock_path = %(here)s/email_fetcher.lock
```
lebouquetin: default path is probably not good. @buxx @inkhey look at other similar parameters (it's similar to cache stuff) and update the default location

tracim/tracim/config/app_cfg.py
lebouquetin: good point, but probably not in this location

tracim/development.ini.base
lebouquetin: by the way, replace `email.reply.filelock_path` by `email.reply.fetcher_lockfile_path`.

tracim/tracim/lib/email_fetcher.py
buxx: I think timeout usage is correct. You have an except after that, it's ok.

tracim/tracim/lib/email_fetcher.py
```diff
@@ -146,6 +158,7 @@ def __init__(
         token: str,
         use_html_parsing: bool,
         use_txt_parsing: bool,
+        filelock_path: str,
```
lebouquetin: rename to `lockfile_path` (the feature is filelock because its a lock based on a file, but the filepath is the lock file path)

tracim/tracim/lib/email_fetcher.py
buxx: I guess the lock is free when no "with lock.acquire" is done. So i guess it is not necessary to delete it when you stop the daemon.

lebouquetin: so add a comment for explaining this.

tracim/tracim/config/app_cfg.py
buxx: Yes after  face to face discuss inkhey suggest same thing.

tracim/development.ini.base
buxx: @lebouquetin You suggest value like ` %(here)s/data` or somethng like `/tmp/lock` ? When inkhey start to code the lock system, i suggest to him to use something like `%(here)s/email_fetcher.lock` insteand something like `/tmp/lock` to prevent conflict between multiple tracim instances on same host.

lebouquetin: `/tmp/lock` is probably conflict-ready ;) we have `%(here)/cache` which is for cache, we have `%(here)/data` which is for data, so probably something like `%(here)/tmp` for temporary content...

tracim/tracim/config/app_cfg.py
buxx: @inkhey Take care with  ̀'/tmp/tracim/email_fetcher.lock'` this default value: Be sure where it is used to create folders in the path. When the default value contain a possible non created folder, we should take care of it.

tracim/tracim/config/app_cfg.py
```diff
@@ -392,8 +392,9 @@ def __init__(self):
             'email.reply.use_txt_parsing',
             True,
         ))
-        self.EMAIL_REPLY_FILELOCK_PATH = tg.config.get(
-            'email.reply.filelock_path',
+        self.EMAIL_REPLY_LOCKFILE_PATH = tg.config.get(
+            'email.reply.lockfile_path',
+            '/tmp/tracim/email_fetcher.lock'
```
lebouquetin: why not use `%(here)s/tmp/email_fetcher.lock` ? `/tmp/tracim/email_fetcher.lock` may result in conflict from different instances

tracim/tracim/lib/email_fetcher.py
```diff
@@ -188,9 +189,15 @@ def __init__(
...
         self.use_html_parsing = use_html_parsing
         self.use_txt_parsing = use_txt_parsing
-        self.lock = filelock.FileLock(filelock_path)
+        self._set_lock(lockfile_path)
         self._is_active = True
 
+    def _set_lock(self, lockfile_path):
+        dir_path = os.path.dirname(lockfile_path)
+        if dir_path:
+            os.makedirs(dir_path, exist_ok=True)
```
lebouquetin: I would not do this because if you make a typo when writing the file path, it may result in creating some folders that you didn't expect to have. For example, if I write `/home/Damien/tracimcache` instead of `/home/damien/tracim/cache` it will potentially create folders that should not exist.

tracim/tracim/config/app_cfg.py
inkhey: %(here)s isn't interpreted correctly here. It will create a '%(here)s' folder...

PR #522, opened 2018-01-10
Imapclient instead of imaplib as imap lib

inkhey:
- Use imapclient instead of imaplib : more pythonic lib
- IDLE Imap feature option
- Don't recreate connection each time, stay in connection up to 'connection_max_lifetime'
- renaming : delay->heartbeat
- Better exceptions : more precise.

### DIFF
--- a/tracim/tracim/lib/email_fetcher.py
+++ b/tracim/tracim/lib/email_fetcher.py
@@ -1,7 +1,6 @@
 # -*- coding: utf-8 -*-
 
 import time
-import imaplib
 import json
 import typing
 from email import message_from_bytes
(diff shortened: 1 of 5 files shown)


tracim/tracim/lib/email_fetcher.py
```diff
@@ -194,137 +209,172 @@ def __init__(
...
+                    if not self._is_active:
+                        logger.warning(self, 'Mail Fetcher process aborted')
+                        sleep_after_connection = False
+                        break
+
+                    if time.time() > deadline:
+                        logger.debug(
+                            self,
+                            "MailFetcher Connection Lifetime limit excess"
+                            ", Try Re-new connection")
```
buxx: I d'ont see where is the re-connection. Where is it ?

inkhey: look at the first while. ;-)

PR #536, opened 2018-02-12
Fix 2 typos in French t10n

feth:
membres (members) -> propose both genders
travaille -> typo (noun is travail)

my first PR :-)


buxx:
Thank you very much !

PR #542, opened 2018-03-08
add copy method into tracim contentAPI and COPY support for webdav

### LINKED ISSUE #540: "COPY" method not implemented in webdav -> file copy not working on some webdav clients.
COPY method for webdav is now not implemented in tracim webdav feature.
 The consequence is that file copy (copying a file from webdav to webdav into the same or a different folder) doesn't work on some client like nautilus 3.22. (Problem appear somewhere after nautilus 3.14.1)
Some other client like thunar are working because they don't rely on COPY to copy file, they do a PUT like for creating a new file.

Solution to fix this bug: 
-  implement "copy" method into content API, like the "move" one
- Create copyMoveSingle method in File (lib/webdav/sql_ressource.py) and  check move_file method.

### LINKED ISSUE #543: Copy comments when using copy feature of contentAPI
New copy feature of content (see https://github.com/tracim/tracim/pull/542 ) doesn't copy comments related to the content. They should be added to copy file.

### LINKED ISSUE #552: When copy is done, 2 new revisions are created.
Using #542 code  up to https://github.com/tracim/tracim/pull/542/commits/29f7860358a03c2afb285457c390bf701aa7dbfe commit.

Currently :

-  Code use "Content.new_revision()" to create new revision (see 'copy' method in content api )
-  When creating this revision, content.new_revision() create 2 new revision instead of one, autoflush change nothing.
- Problem is related to ContentRevisionRO.new_from(content) method, when copying "node" attribute, a new revisions is created, which is not expected.

- This bug doesn't look to affect UI, seems that UI is intelligent enough to know that those 2 revisions are only one.

Expected: 
- Having only 1 more revision added to Content when copying content through copy method of content_api.

inkhey:
Fix https://github.com/tracim/tracim/issues/540 .

**About contentAPI copy feature into tracim :**
Copy feature support copy of all type of document, copy are not exact copy, history and comments are not added to the copy. Copy is same as a special case of "creation" .

**About COPY and webdav**

Webdav support now COPY feature using copy method of tracim contentAPI.
In nautilus 3.26 , recursive COPY of folder work.

closes #540
closes #543
closes #552

### DIFF
--- a/tracim/tracim/lib/content.py
+++ b/tracim/tracim/lib/content.py
@@ -906,9 +906,13 @@ class ContentApi(object):
         content.parent = parent
         content.workspace = workspace
         content.label = label
-        content.revision_type = ActionDescription.MOVE
+        content.revision_type = ActionDescription.COPY
+        content.properties['origin'] = {
+            'content': item.id,
+            'revision' : item.last_revision.revision_id,
+        }
         DBSession.add(content)
-        self.save(content, ActionDescription.MOVE, do_notify=do_notify)
+        self.save(content, ActionDescription.COPY, do_notify=do_notify)
 
     def move_recursively(self, item: Content,
                          new_parent: Content, new_workspace: Workspace):
--- a/tracim/tracim/lib/content.py
+++ b/tracim/tracim/lib/content.py
@@ -860,6 +860,47 @@ class ContentApi(object):
 
         item.revision_type = ActionDescription.MOVE
 
+    def copy(
+        self,
+        item: Content,
+        new_parent: Content=None,
+        new_label: str=None,
+        do_save: bool=True,
+    ) -> None:
+        if not new_parent and not new_label:
+            # TODO - G.M - 08-03-2018 - Use something else than value error
+            raise ValueError("You can't copy file into itself")
+        if new_parent:
+            workspace = new_parent.workspace
+            parent = new_parent
+        else:
+            workspace = item.workspace
+            parent = item.parent
+
+        if new_label:
+            label = new_label
+        else:
+            label = item.label
+
+        with DBSession.no_autoflush:
+            file = self.create(
+                content_type=item.type,
+                workspace=workspace,
+                parent=parent,
+                label=label,
+                do_save=False,
+            )
+            file.description = item.description
+            if item.depot_file:
+                self.update_file_data(
+                    file,
+                    item.file_name,
+                    item.file_mimetype,
+                    item.depot_file.file
+                )
+        if do_save:
+            self.save(file, ActionDescription.CREATION, do_notify=True)
+
     def move_recursively(self, item: Content,
                          new_parent: Content, new_workspace: Workspace):
         self.move(item, new_parent, False, new_workspace)
--- a/tracim/tracim/lib/content.py
+++ b/tracim/tracim/lib/content.py
@@ -865,7 +865,6 @@ class ContentApi(object):
         item: Content,
         new_parent: Content=None,
         new_label: str=None,
-        do_save: bool=True,
         do_notify: bool=True,
     ) -> None:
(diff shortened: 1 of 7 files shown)


tracim/tracim/lib/content.py
```diff
@@ -860,6 +860,47 @@ def move(self, item: Content,
...
+    def copy(
+        self,
+        item: Content,
+        new_parent: Content=None,
+        new_label: str=None,
+        do_save: bool=True,
+    ) -> None:
+        if not new_parent and not new_label:
+            # TODO - G.M - 08-03-2018 - Use something else than value error
+            raise ValueError("You can't copy file into itself")
```
buxx: If `new_parent` and `new_label` are filled but with same parent and label than source the copy should fail too.

tracim/tracim/lib/content.py
```diff
@@ -860,6 +860,47 @@ def move(self, item: Content,
...
+            file.description = item.description
+            if item.depot_file:
+                self.update_file_data(
+                    file,
+                    item.file_name,
+                    item.file_mimetype,
+                    item.depot_file.file
+                )
+        if do_save:
+            self.save(file, ActionDescription.CREATION, do_notify=True)
```
buxx: add `do_notify` parameter in this method and use it in `self.save` call.

tracim/tracim/lib/webdav/sql_resources.py
```diff
@@ -1005,6 +1005,40 @@ def move_file(self, destpath):
 
         transaction.commit()
 
+    def copyMoveSingle(self, destpath, isMove):
+        if isMove:
+            raise NotImplemented
```
buxx: Do you know why wsgidan provide a "copy move" function ? it's weird. No risk than that method called for move something ?

tracim/tracim/lib/webdav/sql_resources.py
```diff
@@ -1005,6 +1005,40 @@ def move_file(self, destpath):
...
+
+        new_file_name = None
+        new_file_extension = None
+
+        # Inspect destpath
+        if basename(destpath) != self.getDisplayName():
+            new_given_file_name = transform_to_bdd(basename(destpath))
+            new_file_name, new_file_extension = \
+                os.path.splitext(new_given_file_name)
```
buxx: This should be refactored. Maybe the "move" method use a similar code block ?

tracim/tracim/lib/webdav/sql_resources.py
inkhey: I do think it's because move and copy are really similar. As far as we override move_recursive method, there should not be case where this method is used for move. I should check this. Normally instead of NotImplemented, call existing move_file method should correct this case if it exist, but as far as i didn't see this case, i prefer to do it.

tracim/tracim/lib/webdav/sql_resources.py
inkhey: Yes, i do think a big refactor is needed, but i prefer to do this in the other PR.

inkhey: #548

tracim/tracim/lib/content.py
```diff
@@ -860,6 +860,57 @@ def move(self, item: Content,
...
+            # TODO - G.M - 08-03-2018 - Use something else than value error
+            raise ValueError("You can't copy file into itself")
+        if new_parent:
+            workspace = new_parent.workspace
+            parent = new_parent
+        else:
+            workspace = item.workspace
+            parent = item.parent
+
+        if new_label:
```
lebouquetin: can be written label = new_label or item.label

inkhey: Done

PR #555, opened 2018-03-21
Fix/535/do not create empty revision

### LINKED ISSUE #535: Create a new version of the file during an empty modification
When you switch to the edit mode of the file and validate a modification without having made any changes in the file, a new version of the file will still create it based on the previous version. This may eventually overload the page with "useless" version.

inkhey:
Do not create new empty revision when content do not change.
Fix #535

### DIFF
--- a/tracim/tracim/lib/webdav/__init__.py
+++ b/tracim/tracim/lib/webdav/__init__.py
@@ -12,7 +12,8 @@ from tracim.model.data import ActionDescription
 from tracim.model.data import ContentType
 from tracim.model.data import Content
 from tracim.model.data import Workspace
-from wsgidav.dav_error import DAVError, HTTP_FORBIDDEN
+from wsgidav.dav_error import DAVError
+from wsgidav.dav_error import HTTP_FORBIDDEN
 
 
 class HistoryType(object):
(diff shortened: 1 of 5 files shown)


tracim/tracim/lib/webdav/__init__.py
```diff
@@ -5,11 +5,14 @@
...
 
 from tracim.lib.content import ContentApi
+from tracim.lib.utils import SameValueError
+from tracim.lib.base import logger
 from tracim.model import new_revision
 from tracim.model.data import ActionDescription
 from tracim.model.data import ContentType
 from tracim.model.data import Content
 from tracim.model.data import Workspace
+from wsgidav.dav_error import DAVError, HTTP_FORBIDDEN, HTTP_NOT_MODIFIED
```
buxx: `import` code style to follow: "Imports" chapter in http://algoo.trac.im/workspaces/11/folders/1194/pages/2043

buxx:
Does exist a unit test who check the case (do not create an empty revision) ?

tracim/tracim/lib/webdav/__init__.py
inkhey: HTTP_NOT_MODIFIED is not used anymore, i drop it.

inkhey:
@buxx Now, there is !

PR #558, opened 2018-03-21
WIP: Fix webdav for davfs2

inkhey:
davfs2 need a working "LOCK" (Doing a lock create automatically an empty file) . As method to create empty file doesn't return a 'File' type, automatic creation of new file when "LOCK" inexistent file in wsgidav failed. 

Return a valid File in method to create empty file resolve this issue.

### DIFF
--- a/tracim/tracim/lib/webdav/sql_resources.py
+++ b/tracim/tracim/lib/webdav/sql_resources.py
@@ -215,14 +215,24 @@ class Workspace(DAVCollection):
         if resource:
             content = resource.content
 
-        return FakeFileStream(
-            file_name=file_name,
-            content_api=self.content_api,
-            workspace=self.workspace,
-            content=content,
-            parent=self.content,
-            path=self.path + '/' + file_name
-        )
+        if not content:
+            # INFO - G.M - 21-03-2018 create new empty file and commit it.
+            new_empty_file = FakeFileStream(
+                file_name=file_name,
+                content_api=self.content_api,
+                workspace=self.workspace,
+                content=content,
+                parent=self.content,
+                path=path,
+            )
+            new_empty_file.close()
+
+        # INFO - G.M - 21-03-2018 - Obtain true davFile from provider
+        resource = self.provider.getResourceInst(path, self.environ)
+        if not resource:
+            raise DAVError(HTTP_FORBIDDEN)
+        content = resource.content
+        return File(path=path, content=content,environ=self.environ)
 
     def createCollection(self, label: str) -> 'Folder':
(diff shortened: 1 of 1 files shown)


tracim/tracim/lib/webdav/sql_resources.py
```diff
@@ -215,14 +215,24 @@ def createEmptyResource(self, file_name: str):
...
+            # INFO - G.M - 21-03-2018 create new empty file and commit it.
+            new_empty_file = FakeFileStream(
+                file_name=file_name,
+                content_api=self.content_api,
+                workspace=self.workspace,
+                content=content,
+                parent=self.content,
+                path=path,
+            )
+            new_empty_file.close()
```
buxx: Is it necessary to use FakeFileStream ? If the goal is to create a Content, maybe just use ContentLib ?

tracim/tracim/lib/webdav/sql_resources.py
```diff
@@ -215,14 +215,24 @@ def createEmptyResource(self, file_name: str):
...
+
+        # INFO - G.M - 21-03-2018 - Obtain true davFile from provider
+        resource = self.provider.getResourceInst(path, self.environ)
+        if not resource:
+            raise DAVError(HTTP_FORBIDDEN)
+        content = resource.content
+        return File(path=path, content=content,environ=self.environ)
```
buxx: missing space here: `content=content,Xenviron=`

tracim/tracim/lib/webdav/sql_resources.py
```diff
@@ -215,14 +215,24 @@ def createEmptyResource(self, file_name: str):
...
+                content=content,
+                parent=self.content,
+                path=path,
+            )
+            new_empty_file.close()
+
+        # INFO - G.M - 21-03-2018 - Obtain true davFile from provider
+        resource = self.provider.getResourceInst(path, self.environ)
+        if not resource:
+            raise DAVError(HTTP_FORBIDDEN)
```
buxx: Why return `HTTP_FORBIDDEN` instead 404 ?

tracim/tracim/lib/webdav/sql_resources.py
inkhey: I'm don't think so, this MR is actually more a proof of concept, i will refactor this. i'm thinking about removing all or most FakeFileStream class  and add his code directly into File (dav_resource).

buxx: Ok. So open an issue for that.

PR #1135, opened 2018-11-15
add legal stuff

inkhey:
- add mit license
- add dco

### DIFF
--- a/DCO
+++ b/DCO
@@ -1,0 +1,37 @@
+Developer Certificate of Origin
+Version 1.1
+
+Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
+1 Letterman Drive
+Suite D4700
+San Francisco, CA, 94129
+
+Everyone is permitted to copy and distribute verbatim copies of this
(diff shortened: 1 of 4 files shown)


DCO
```diff
@@ -0,0 +1,37 @@
+Developer Certificate of Origin
+Version 1.1
+
+Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
```
tracim: You should put Algoo, here... no?
by the way, copyright should be 2013, 2018

inkhey: I don't think we should put it there, DCO v1.1 copyright owner are Linux Foundation.
Gitlab itself, let linux foundation copyright : https://gitlab.com/gitlab-org/dco/blob/master/README.md .

But you're right for date in MIT license.

tracim: ok, that's fine

inkhey:
@tracim I do not really like to force an invalid extension in this type of file. LICENSE and DCO file contains plain text, not markdown.

Issue #1144, opened 2018-11-16
FIx unwanted error 500 on standard endpoint when depot file is corrupted


inkhey:
**Describe the bug**
Currently when using a corrupted depot file (does not match with link in database) we get error 500 on endpoint who are not directly related to depot generic info content endpoint or revisions endpoints.

**Expected behavior**
Return 200 for those endpoint with logged stacktrace about the issue.

PhilippeAccorsi:
If its done yet in v2.0 please change Milestone and project to 2.0

PhilippeAccorsi:
For test: rename depot folder and try to use tracim.
Everything work but not preview of file and its not  possible to download file from old depot.

PhilippeAccorsi:
Verified.

PR #1145, opened 2018-11-16
Fix/1144 better deal with corrupted depot folder

inkhey:
related to #1144 

- Better support of corrupted depot folder case: More explicit exception and does not breake anymore standard endpoints.

### DIFF
--- a/backend/tracim_backend/models/context_models.py
+++ b/backend/tracim_backend/models/context_models.py
@@ -909,6 +909,12 @@ class ContentInContext(object):
             return None
         try:
             return self.content.depot_file.file.content_length
+        except IOError as e:
+            logger.warning(
+                self,
+                "IO Exception Occured when trying to get content size  : {}".format(str(e))
+            )
+            logger.warning(self, traceback.format_exc())
         except Exception as e:
             logger.warning(
                 self,
--- a/backend/tracim_backend/models/context_models.py
+++ b/backend/tracim_backend/models/context_models.py
@@ -1,5 +1,6 @@
 # coding=utf-8
 import cgi
+import traceback
 import typing
 from datetime import datetime
 from enum import Enum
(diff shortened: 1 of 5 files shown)


backend/tracim_backend/models/data.py
```diff
@@ -716,11 +723,17 @@ def copy(
...
-                revision.file_mimetype,
-            )
+            try:
+                copy_rev.depot_file = FileIntent(
+                    revision.depot_file.file.read(),
+                    revision.file_name,
+                    revision.file_mimetype,
+                )
+            except IOError as exc:
+                raise NewRevisionAbortedDepotCorrupted(
```
tracim: `NewRevisionAbortedDepotCorrupted` => `CopyRevisionAbortedDepotCorrupted` (this allow to distinguish the two cases)

backend/tracim_backend/models/data.py
```diff
@@ -716,11 +723,17 @@ def copy(
...
+                    "Can't create new revision of content if"
+                    " last revision file is not accessible."
```
tracim: I suggest not to interpret the reason for IOError. (in most cases, "last revision file is not accessible." is the right interpretation, but you can't be sure - you are not the developer of FileIntent. I suggest you to reformulate like "Can't create new revision by copying another one. May be related to original revision file not being available"

backend/tracim_backend/models/data.py
```diff
@@ -686,11 +687,17 @@ def new_from(cls, revision: 'ContentRevisionRO') -> 'ContentRevisionRO':
...
+            try:
+                new_rev.depot_file = FileIntent(
+                    revision.depot_file.file.read(),
+                    revision.file_name,
+                    revision.file_mimetype,
+                )
+            except IOError as exc:
+                raise NewRevisionAbortedDepotCorrupted(
+                    "Can't create new revision of content"
+                    " if last revision file is not accessible."
```
tracim: Some remark. You can suggest to check that the file is available, but do not say "the file is not available" - you did not check it)

inkhey: done

backend/tracim_backend/models/data.py
inkhey: done

backend/tracim_backend/models/data.py
inkhey: done

backend/tracim_backend/models/context_models.py
```diff
@@ -903,10 +905,18 @@ def size(self) -> typing.Optional[int]:
...
         :return: size of content if available, None if unavailable
         """
-        if self.content.depot_file:
-            return self.content.depot_file.file.content_length
-        else:
+        if not self.content.depot_file:
             return None
+        try:
+            return self.content.depot_file.file.content_length
+        except Exception as e:
```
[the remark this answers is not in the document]
inkhey: done

Issue #1050, opened 2018-10-19
add minimum number of letter for public name, password and email


PhilippeAccorsi:
At present

You can hace user with just one letter in public name but if you need to add this user in one workspace you cant find this user because frontedn send request to backend when we type minimum 2 letters.

Expected

- [ ] define a minimum/max of necessary letter for the public name.
- [ ] define a minimum/max of necessary letter for the email.
- [ ] define a minimum/max of necessary letter for the password.

tracim:
We must allow 3 chars minimum. This is to implement on backend side first.

PhilippeAccorsi:
@inkhey please assign @Skylsmoi when you change label. This issue is available for frontend since 7 hour but frontend dont know that.

PhilippeAccorsi:
@Skylsmoi in my account if I tried to change name with less than 3 caracter i can see this message: `Warning
Error while changing name`

Expected

- [ ] Have same flash message that we have in add user `Warning
Full name must be at least 3 characters`

Skylsmoi:
added in my account and admin account

PhilippeAccorsi:
Verified.

PR #1201, opened 2018-11-28
Fix/1050 add length limitation to field in endpoint

inkhey:
related to https://github.com/tracim/tracim/issues/1050,

- [X] add password,email, public_name length limit in endpoint


lebouquetin:
@inkhey I do not see the tests in charge of validating that small email / public_name / password result in a HTTP 400. You should at least test these cases

lebouquetin:
By the way, these validations MUST be implemented at core lib level, not on serialization level. Otherwise, a user can be created through CLI with a display_name of 2 chars for example, but everything else may fail because this case is not expected.

It's not required to add this limitations at the database level

inkhey:
@lebouquetin I don't really like the idea of having 2 level for validation of almost same things and If we drop validation at serialisation level, we lost documentation too.
I'm checking if we can do both but using same validation code.

inkhey:
poc about reusing validator from marshmallow: https://github.com/tracim/tracim/commit/b5e1506e8b5f3480c82ddf323d1ca1d6f607bf24

backend/tracim_backend/app_models/validator.py
```diff
@@ -1,11 +1,93 @@
...
+from tracim_backend.models.data import UserRoleInWorkspace
 
-all_content_types_validator = OneOf(choices=[])
+class TracimValidator(object):
+
+    def __init__(self):
+        self.validators = {}
+        self.values = {}
+
+    def add_validator(self, field:str, value:str, validator:typing.Callable) -> bool:  # nopep8
```
buxx: You can be more precise on Callable type:

    validator:typing.Callable[[typing.Any], None]

backend/tracim_backend/app_models/validator.py
```diff
@@ -1,11 +1,93 @@
...
 # TODO - G.M - 2018-08-08 - [GlobalVar] Refactor Global var
 # of tracim_backend, Be careful all_content_types_validator is a global_var !
+from tracim_backend.exceptions import TracimValidationFailed
+from tracim_backend.models.auth import Profile
+from tracim_backend.models.auth import User
+from tracim_backend.models.data import ActionDescription
+from tracim_backend.models.data import UserRoleInWorkspace
 
-all_content_types_validator = OneOf(choices=[])
+class TracimValidator(object):
```
buxx: Can you add a doctring ?

backend/tracim_backend/app_models/validator.py
buxx: Can you add a doctsring ? One of my interrogation is for example: why when value is "false", the method return False ?

backend/tracim_backend/app_models/validator.py
```diff
@@ -1,11 +1,93 @@
...
+
+    def __init__(self):
+        self.validators = {}
+        self.values = {}
```
buxx: Can you add type of these variables ?

backend/tracim_backend/app_models/validator.py
```diff
@@ -1,11 +1,93 @@
...
+
+    def add_validator(self, field:str, value:str, validator:typing.Callable) -> bool:  # nopep8
+        if not value:
+            return False
+
+        self.validators[field] = validator
+        self.values[field] = value
+        return True
 
+    def validate_all(self):
```
buxx: Missing type and doctrsing. Additional question: Do we really need return a True when it is ok ? I don't think so.

backend/tracim_backend/app_models/validator.py
```diff
@@ -1,11 +1,93 @@
...
+
+## TODO - G.M - 2018-11-29 - Refactor validator system
+# static validator #
+# int
+bool_as_int_validator = Range(min=0, max=1, error="Value must be 0 or 1")
+strictly_positive_int_validator = Range(min=1, error="Value must be positive")
+positive_int_validator = Range(min=0, error="Value must be positive or 0")
+
+# string
+regex_string_as_list_of_int = Regexp(regex=(re.compile('^(\d+(,\d+)*)?$')))
```
buxx: Can you add a comment about what do the regex ?

inkhey:
poc was added to this branch.

backend/tracim_backend/tests/functional/test_account.py
```diff
@@ -121,7 +121,7 @@ def test_api__get_recently_active_content__ok__200__nominal(self):
             'Basic',
             (
                 'test@test.test',
-                'pass'
+                'password'
```
buxx: haha ;)

backend/tracim_backend/app_models/validator.py
inkhey: @buxx 
That's because None case is not handled by validator for two main reason:
-> None case is not handled by validate in marshmallow (there is an "allow_none" param)
-> We want to test easily if all field are correct in user update, but update method allow none value, if value is none, param is not modified. It become really complex to deal with this if we handle None case in validator itself.

backend/tracim_backend/app_models/validator.py
inkhey: That's just more convenient for testing.

backend/tracim_backend/tests/functional/test_account.py
inkhey: :smile:

backend/tracim_backend/app_models/validator.py
inkhey: done

backend/tracim_backend/app_models/validator.py
inkhey: done.

backend/tracim_backend/app_models/validator.py
inkhey: done.

Issue #1222, opened 2018-12-04
remove code from models/__init__.py


inkhey:
**Describe the bug**
Models code in models/__init__.py create many issue related to import. We should remove all code from this place and replace it somewhere else.

Issue #1210, opened 2018-11-29
app folder: right about create folder or/and content in folder is not applied correctly


PhilippeAccorsi:
at present:

User with role contributor can create folder but he cant access to app folder advanced and he cant create content in this folder.

Expected:

- [ ] Apply right about role for all content (look here https://github.com/tracim/tracim/blob/develop/backend/doc/roles.md   **warning**: role about folder is missing, issue about that is open).
- [ ] contributor cant create folder but he can create content in folder
- [ ] update documentation in backend too about this new rule

tracim:
@PhilippeAccorsi all authorization/authentification related stuff MUST be first fixed in backend. What I understand here is "maybe the problem is frontend, maybe backend". Si first step is to validate/fix it in backend (maybe the problem does not appear in backend... if so, then add a comment in the issue telling that "checks have been done in backend and backend is ok")

PhilippeAccorsi:
@lebouquetin ok. I have discuss with back and front before create this issue, and idea is frontend issue.

tracim:
I understand. What I want is before assigning the issue to frontend, you *must be sure* that there are no problem in backend. It means "all checks have been done in backend", then if everything is fine (and you are sure about it) you can assign the issue to frontend.

PhilippeAccorsi:
@Skylsmoi backend is done and merged. 
@inkhey please check all message on issue and change label and reassign when its necessary. This issue need frontend job before testing.

PhilippeAccorsi:
Verified.

PR #1224, opened 2018-12-04
Fix/1210 better folder right role

inkhey:
related to #1210
:warning: require #1222 and #1223


buxx:
closed in favor of #1238.

PR #1238, opened 2018-12-07
Fix/1210 better folder right role 

buxx:
related to #1210 

Reopen of #1224 (to regenerate github diff)


backend/tracim_backend/lib/utils/authorization.py
```diff
@@ -181,6 +182,40 @@ def check(
...
+    Check if user can create content of this type
+    """
+    def __init__(
+            self,
+            content_type_list: ContentTypeList,
+            content_type_slug: typing.Optional[str] = None,
+    ):
+        """
+        :param content_type: force to check a content_type, if not provided,
+        check is done with tracim_context.candidate_content_type
```
buxx: docstring seems outdated

buxx:
Ok for me. i wait travis confirmation and i merge.

backend/tracim_backend/lib/utils/authorization.py
inkhey: you're right.

Issue #1187, opened 2018-11-26
webdav: some issues


PhilippeAccorsi:
**At present**

- All user can create folder with webdav (no verification about role)
- ~~all user cant create share space with webdav~~ => this is the normal behavior
- authorized user can copy file from local to tracim but an error appears (and if you refresh the screen, the file has been copied)
~~~
11:43:28 ERROR [waitress:341][waitress] Exception when serving /'fr'/test1/bug2.jpg
Traceback (most recent call last):
  File "/home/algoo/repo/tracim/backend/env/lib/python3.6/site-packages/waitress/channel.py", line 338, in service
    task.service()
  File "/home/algoo/repo/tracim/backend/env/lib/python3.6/site-packages/waitress/task.py", line 169, in service
    self.execute()
  File "/home/algoo/repo/tracim/backend/env/lib/python3.6/site-packages/waitress/task.py", line 421, in execute
    for chunk in app_iter:
  File "/home/algoo/repo/tracim/backend/env/lib/python3.6/site-packages/wsgidav/wsgidav_app.py", line 428, in __call__
    for v in app_iter:
  File "/home/algoo/repo/tracim/backend/tracim_backend/lib/webdav/middlewares.py", line 130, in __call__
    for v in app_iter:
  File "/home/algoo/repo/tracim/backend/env/lib/python3.6/site-packages/wsgidav/error_printer.py", line 47, in __call__
    for v in app_iter:
  File "/home/algoo/repo/tracim/backend/env/lib/python3.6/site-packages/wsgidav/request_resolver.py", line 199, in __call__
    for v in app_iter:
  File "/home/algoo/repo/tracim/backend/env/lib/python3.6/site-packages/wsgidav/request_server.py", line 125, in __call__
    app_iter = provider.customRequestHandler(environ, start_response, method)
  File "/home/algoo/repo/tracim/backend/env/lib/python3.6/site-packages/wsgidav/dav_provider.py", line 1471, in customRequestHandler
    return defaultHandler(environ, start_response)
  File "/home/algoo/repo/tracim/backend/env/lib/python3.6/site-packages/wsgidav/request_server.py", line 781, in doPUT
    if res.supportEtag():
AttributeError: 'FakeFileStream' object has no attribute 'supportEtag'
~~~
~~~
Message => erreur lors de la copie dav://192.168.1.150:3030/'fr'/test1
internal error
~~~

- copy file from tracim to local is not working (html-document)
~~~
[...]

PR #1241, opened 2018-12-07
Fix/1187 repair webdav

inkhey:
:warning:  Windows manual test success, we can merge this.
related to #1187 .

included #1139 

- Webdav config in same config file as tracim
- Fix many webdav issue related to breaking change in tracim_v2
- Fix webdav issue about right not correctly handled (add right level check in webdav code
- drop dead code about webdav deleted/archived
- Add few webdav functional test


inkhey:
:warning: We can merge this.
modification of code in frontend break cypress test, as #1264 was not yet merged, cypress test are not yet tested in frontend. Coverage issue will not be fixed yet, we need more functionalm test for webdav, see : https://github.com/tracim/tracim/issues/1247

Issue #1249, opened 2018-12-07
migration of bdd not working


PhilippeAccorsi:
At present:

~~~
Traceback (most recent call last):
  File "/home/come/repo/tracim/backend/env/bin/alembic", line 11, in <module>
    load_entry_point('alembic==1.0.5', 'console_scripts', 'alembic')()
  File "/home/come/repo/tracim/backend/env/lib/python3.5/site-packages/alembic/config.py", line 503, in main
    CommandLine(prog=prog).main(argv=argv)
  File "/home/come/repo/tracim/backend/env/lib/python3.5/site-packages/alembic/config.py", line 497, in main
    self.run_cmd(cfg, options)
  File "/home/come/repo/tracim/backend/env/lib/python3.5/site-packages/alembic/config.py", line 480, in run_cmd
    **dict((k, getattr(options, k, None)) for k in kwarg)
  File "/home/come/repo/tracim/backend/env/lib/python3.5/site-packages/alembic/command.py", line 254, in upgrade
    script.run_env()
  File "/home/come/repo/tracim/backend/env/lib/python3.5/site-packages/alembic/script/base.py", line 427, in run_env
    util.load_python_file(self.dir, 'env.py')
  File "/home/come/repo/tracim/backend/env/lib/python3.5/site-packages/alembic/util/pyfiles.py", line 81, in load_python_file
    module = load_module_py(module_id, path)
  File "/home/come/repo/tracim/backend/env/lib/python3.5/site-packages/alembic/util/compat.py", line 145, in load_module_py
    spec.loader.exec_module(module)
  File "<frozen importlib._bootstrap_external>", line 673, in exec_module
  File "<frozen importlib._bootstrap>", line 222, in _call_with_frames_removed
  File "tracim_backend/migration/env.py", line 80, in <module>
    run_migrations_online()
  File "tracim_backend/migration/env.py", line 71, in run_migrations_online
    context.run_migrations()
  File "<string>", line 8, in run_migrations
  File "/home/come/repo/tracim/backend/env/lib/python3.5/site-packages/alembic/runtime/environment.py", line 836, in run_migrations
    self.get_context().run_migrations(**kw)
  File "/home/come/repo/tracim/backend/env/lib/python3.5/site-packages/alembic/runtime/migration.py", line 330, in run_migrations
    step.migration_fn(**kw)
[...]

PR #1254, opened 2018-12-10
Fix/1249_fix migration script+tests

inkhey:
related to #1249


inkhey:
:warning: We can merge this.
Coverage fail because coverage doesn't cover all migration code in one build (each specific code running in each bdd specific build)

Issue #1005, opened 2018-10-11
Write a migration documentation


buxx:
- [ ] Write a migration documentation (v1 -> v2)

PhilippeAccorsi:
Here you find documentation.

https://github.com/tracim/tracim/blob/develop/backend/doc/migrate_from_v1.md

PhilippeAccorsi:
need some updated information.

PhilippeAccorsi:
Verified.

PR #1271, opened 2018-12-13
fix doc about ldap and migration

PhilippeAccorsi:
Modification about LDAP.


PhilippeAccorsi:
Related to https://github.com/tracim/tracim/issues/1005

Issue #1300, opened 2019-01-08
docker image: if database exist, no upgrade is make when start docker image. Webdav parameter "worker" is missing.


PhilippeAccorsi:
At present

If you start tracim image with existing database, the upgrade of this database is not start.
When you use webdav with docker, Request have lock error. Something missing in wsgi conf.
Expected

- [x] upgrade database when you start docker image if its not an initialization of new databse.
- [x] Add "worker = 1" in tracim_webdav.ini about lock and multiprocess.

PhilippeAccorsi:
Verified.

PR #1301, opened 2019-01-08
#1300 fix missing condition in entrypoint

PhilippeAccorsi:
Add condition to upgrade database if initialization of database is not necessary.
Add parameter in wsgi file for lock with webddav.


PhilippeAccorsi:
related to https://github.com/tracim/tracim/issues/1300

Issue #1318, opened 2019-01-15
Docker: missing mail_fetcher in entrypoint and also mail_notifier


PhilippeAccorsi:
At present

Reply by email not working in docker images because some parameter is not correctly added.
Daemon about mail_fetcher (sync) and mail_notifier (async) is not started.

Expected

- [x] add possibility to start mail_fetcher or mail_notifier
=> its necessary to export 'export TRACIM_CONF_PATH="$(pwd)/development.ini"' before starting daemon.

PhilippeAccorsi:
Idea for running mail_fetcher daemon:

`export TRACIM_CONF_PATH="/etc/tracim/development.ini"`
`nohup python3 /tracim/backend/daemons/mail_fetcher.py > /var/tracim/logs/mail_fetcher.log 2>&1 </dev/null &`

PhilippeAccorsi:
@lebouquetin do you think its necessary to save the mail_fetcher log? Because inside we can see raw_content of reply.

PhilippeAccorsi:
Idea to entrypoint:

Use env variable to start email_fetcher or mail_notifier (but what is going on if you dont use reply by email?).
EMAIL_MODE=sync 
EMAIL_MODE=async

PhilippeAccorsi:
First test with daemon is working good for reply_by_email (after install lib requests see #1317 )

Now I try to use supervisord for better solution.

buxx:
@PhilippeAccorsi 

> Idea to entrypoint:
>
>Use env variable to start email_fetcher or mail_notifier (but what is going on if you dont use reply by email?).
>EMAIL_MODE=sync
>EMAIL_MODE=async

Take care with `EMAIL_MODE` env variable: there are config in development.ini to configure that in Tracim. Is docker image user use this env variable, he must update tracim config too (or config must be updated automatically).

PhilippeAccorsi:
@buxx my answer is just an idea, but thanks for your information.

I have used more explicit variable:
- REPLY_BY_EMAIL
- EMAIL_MODE_ASYNC

PhilippeAccorsi:
Verified.

Issue #1317, opened 2019-01-15
reply_by_email: lib 'requests' is not defined in production (just in test)


PhilippeAccorsi:
**at present**

In `/tracim/backend/tracim_backend/lib/mail_fetcher/email_fetcher.py` the lib 'requests' is just install for test.

**Expected**

- [ ] Remove this lib from 'tests_require' and put in 'requires' (make sur test working all the time)
- [ ] Update requirements.txt

PhilippeAccorsi:
@inkhey Its necessary to make this change today.
We need to merge this change very quickly.

PhilippeAccorsi:
Related to #1318

PhilippeAccorsi:
Verified.

PR #1328, opened 2019-01-16
Fix/1318 fix docker missing reply email daemon

PhilippeAccorsi:
Add:
- supervisord to start python service
- parameter in entrypoint
- new config file for supervisord

related to #1318


PhilippeAccorsi:
This fix need also resolution of #1317
MR #1320

PR #1330, opened 2019-01-16
Some fix to make rc3

PhilippeAccorsi:
Fix:
- doc
- lib missing in prod
- fix service mail_fetcher and mail_notifier


PhilippeAccorsi:
Ready to create release_02.01.00-rc3

Issue #1183, opened 2018-11-26
log message not good about async or sync when sending email


PhilippeAccorsi:
At present

When one email is send by tracim, log message is "Sending asynchronous email to" but this message is the same with config async or sync.

Todo :
- Change the log to something that match to reality.
- Explain with log if emails are send synchronously or asynchronously in log.
- One other nice thing to do is to replace log file for notification (email.notification.log_file_path) to a log handle the same way as normal log.

inkhey:
In fact, log message is incorrect, at this level, it's not possible to know if message is send synchronously or asynchronously. The correct log should be like "Generating content X notification email for X user(s)"

inkhey:
https://github.com/tracim/tracim/pull/1344  resolve all 3 issue : 
- Log for building notification correctly named
- Log for sending asynchronously/synchronously email
- Log for notification are converted as pastedeploy format but default config allow us to use almost same way as before, in a separate log file.
This log file is different from main log of tracim as it log only some information about email in a specific format.

PhilippeAccorsi:
@inkhey More information about this issue (after test):

~~~
FROM users 
WHERE users.is_deleted = 0 AND users.email = ?
16:28:38 INFO  [sqlalchemy.engine.base.Engine:1154][waitress] ('admin@admin.admin',)
16:28:38 DEBUG [tracim:24][waitress] [EmailManager] user: 1
16:28:38 INFO  [tracim:38][waitress] [EmailManager] Sending asynchronous email to 1 user (admin@admin.admin)
16:28:38 INFO  [tracim:38][waitress] [send_email_through] send email to Global manager <admin@admin.admin> synchronously
16:28:38 INFO  [tracim:38][waitress] [EmailSender] Connecting from SMTP server mail.gandi.net
16:28:39 INFO  [tracim:38][waitress] [EmailSender] SMTP Start TLS OK
16:28:39 DEBUG [tracim:24][waitress] [EmailSender] SMTP Start TLS return code: 220 with message: 2.0.0 Ready to start TLS
16:28:39 INFO  [tracim:38][waitress] [EmailSender] SMTP Authentication Successful
16:28:39 DEBUG [tracim:24][waitress] [EmailSender] SMTP login return code: 235 with message: 2.7.0 Authentication successful
16:28:39 INFO  [tracim:38][waitress] [EmailSender] Sending email to Global manager <admin@admin.admin>
[...]

PR #1344, opened 2019-01-21
Fix/1183 better log for sync/async mail

inkhey:
related to #1183 

- [X] do not return "async email" when this is not true.
- [X] refactor logger for mail logger, using pastedeploy log instead of custom code.
- [X] more explict log : we are able to know when a synchronous/asynchronous mail is send to EmailSender.

### DIFF
--- a/backend/development.ini.sample
+++ b/backend/development.ini.sample
@@ -207,8 +207,7 @@ email.processing_mode = sync
 
 ## available choice: True or False
 email.notification.activated = False
-## You can change file_path for mail-notifications.log
-; email.notification.log_file_path = /tmp/mail-notifications.log
+# You can activate for mail-notifications.log using 'tracim_notification' logger.
 
 ### SMTP ###
 email.notification.smtp.server = your_smtp_server
@@ -430,6 +429,9 @@ script_location = tracim_backend/migration
 ### LOGGERS ###
 [loggers]
 keys = root, tracim, sqlalchemy, alembic, hapic
+# add Tracim_notification logfile
+# this need to add tracim_notification handler and tracim_notification formatter
+; keys = root, tracim, sqlalchemy, alembic, hapic, tracim_notification
 
 [logger_root]
 level = INFO
@@ -455,6 +457,12 @@ level = DEBUG
 handlers =
 qualname = hapic
 
+[logger_tracim_notification]
+level = INFO
+handlers = tracim_notification
+qualname = tracim_notification
+propagate = 0
+
 ### FORMATTER ###
 # colored log formatter for console (do not work properly with file)
 # this need coloredlog python package, you can install it in venv
@@ -474,11 +482,17 @@ datefmt = %H:%M:%S
 format = %(asctime)s %(levelname)-5.5s [%(name)s:%(lineno)s][%(threadName)s] %(message)s
 datefmt = %H:%M:%S
 
+[formatter_tracim_notification]
+format = %(asctime)s %(network)s %(action)s %(recipient)s %(subject)s %(message)s
+datefmt = %H:%M:%S
+
 ### HANDLERS ###
 [handlers]
 keys = console
 # use file log instead
 ; keys = logfile
+# add tracim_notification_handler
+; keys = console, tracim_notification
 
 ### Console log ###
 [handler_console]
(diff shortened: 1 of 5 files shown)


backend/development.ini.sample
```diff
@@ -430,6 +429,9 @@ script_location = tracim_backend/migration
 ### LOGGERS ###
 [loggers]
 keys = root, tracim, sqlalchemy, alembic, hapic
+# add Tracim_notification logfile
```
lebouquetin: why not to add tracim_notification to default loggers ?

lebouquetin: this will allow to remove this comment by the way

backend/development.ini.sample
```diff
@@ -474,11 +482,17 @@ datefmt = %H:%M:%S
 format = %(asctime)s %(levelname)-5.5s [%(name)s:%(lineno)s][%(threadName)s] %(message)s
 datefmt = %H:%M:%S
 
+[formatter_tracim_notification]
```
lebouquetin: is there a reason for not using default formatter?

backend/development.ini.sample
```diff
@@ -498,3 +512,10 @@ level=NOTSET
 # number of max log file = 5
 args=('tracim.log','a',10*1024*1024,5)
 formatter=generic
+
+# file log for tracim_notification
```
lebouquetin: same remark: why not to use default handler by default (the admin sys may change this if he wants to - but default configuration would be simplier)

backend/development.ini.sample
lebouquetin: You should explain here that tracim_notification is used for logging email prepared and send from tracim. this allow to investigate cases where mails are not received by users

backend/development.ini.sample
```diff
@@ -498,3 +512,10 @@ level=NOTSET
...
+[handler_tracim_notification]
+class=FileHandler
+level=INFO
+args = ('%(here)s/tracim_notification.log','a')
```
lebouquetin: we do not use `%(here)` variable for `handler_logfile`, I suggest to use a similar configuration

backend/tracim_backend/lib/mail_notifier/notifier.py
```diff
@@ -20,6 +20,7 @@
 from tracim_backend.lib.mail_notifier.utils import EST
 from tracim_backend.lib.mail_notifier.utils import SmtpConfiguration
 from tracim_backend.lib.utils.logger import logger
+from tracim_backend.lib.utils.logger import notification_logger
```
lebouquetin: I suggest to move the notification_logger to `tracim_backend/lib/mail_notifier`. what do you think? This logger is more "functionnal" than "technical". What do you think?

backend/tracim_backend/lib/mail_notifier/notifier.py
```diff
@@ -187,25 +188,23 @@ def _get_sender(self, user: User=None) -> str:
     @staticmethod
     def log_notification(
             config: CFG,
+            msg: str,
```
lebouquetin: adding `msg` may be interpreted as "msg is the email message" (because there are also `recipient` and `subject` parameters). I suggest to keep msg and to rename r`ecipient` and `subject` to `mail_recipient` and `mail_subject`

backend/development.ini.sample
inkhey: Yes, We can write directly tracim_notification.log but we should be careful with relative path. I've got  some trouble without finding were the log was stored, so i set default to this working value.

backend/tracim_backend/lib/mail_notifier/notifier.py
inkhey: That's a correct choice as soon as we don't want to log another type of notifcation than email one.
I'm thinking about this logger as more generic than just a "email notification logger".

backend/tracim_backend/lib/mail_notifier/notifier.py
inkhey: done, i've use **email** instead of mail for better code coherency and  also change log_notification (in EmailManager) to **log_email_notification** in  to be more explicit.

backend/tracim_backend/lib/mail_notifier/notifier.py
inkhey: @lebouquetin so should we consider the notifier as email notifier or should we be more global ?
I do think if we decide to consider it as email-only, that we should change the naming.

lebouquetin: as we talked, move the logger to mail module

Issue #1161, opened 2018-11-21
backend documentation: missing information about async 


PhilippeAccorsi:
At present:

No information about what is the solution to use async configuration. This parameter email.processing_mode is not documented.

Expected:

- [ ] Add better documentation about application you need to start to use async mode.

inkhey:
@PhilippeAccorsi to avoid confusing stuff, as email.notification.processing_mode as only one valid value, i did a modification to disable the feature : https://github.com/tracim/tracim/pull/1346 . see #1345 for issue about reenabling it correctly with meaningfull stuff.

lebouquetin:
@inkhey I agree with your suggestion.

Issue #1345, opened 2019-01-22
Add feature to process email (generating template, etc...) asynchronously


inkhey:
Actually only sending part of email process can be done asynchronously, we can change this to have also email creation done in async.
Entry to this feature already exist in tracim_v1, using email.notification.processing_mode, but async was not at all implemented.
This feature is now disable by commit

PR #1346, opened 2019-01-22
disable email.notification.processing_mode feature, set it always as sync

inkhey:
related to #1161


buxx:
Why disable asynchronus email sending feature @inkhey @lebouquetin ?

buxx:
> Why disable asynchronus email sending feature @inkhey @lebouquetin ?

If it is related to #1345 , please make a reference into #1161 .

inkhey:
> Why disable asynchronus email sending feature @inkhey @lebouquetin ?

This is not asynchronous mail sending but another feature, there is actually 2 level of asynchronous for email notification but this one (asynchronous for mail notification building) isn't implemented and create only trouble when reading configuration file.

buxx:
> but this one (asynchronous for mail notification building) isn't implemented and create only trouble when reading configuration file.

Ok, thank's for precisions !

inkhey:
@lebouquetin It's not never read, but it exist only one valid value. As this issue is to not show the parameter, i didn't wanted to add modification to the way it was handled internally.

Issue #993, opened 2018-10-06
How does the auth mechanism work ?


buxx:
I was reading auto generated tracim v2 backend documentation and i asked myself: How give auth information during api discussions ?

I think it is missing in documentation. It should be added in (possible locations):

* backend README.md
* backend auto generated doc in global description
* backend auto generated doc in each api path request headers + returned headers in `/api/v2/sessions/login`.

tracim:
This documentation must be part of the basic documentation "Howto interact with tracim API"

inkhey:
@buxx @lebouquetin can I've done a documentation about using api. Not sure if this is enough, not sure about what the meaning of the issue (more technical stuff about auth mecanism, a figure ?)

Also this documentation do not explain setup of auth mecanism, but another PR add some doc about it, see here : https://github.com/tracim/tracim/pull/1338.

buxx:
The goal of issue is to enhance documentation about auth mechanisms. Wrote documentation about api is ok for me. Additional: 

We can add a link to this documentation into generated api doc.

PhilippeAccorsi:
Verified.

PR #1348, opened 2019-01-22
Add documentation for api

inkhey:
related to #993


buxx:
@Skylsmoi Can you make an english quality review (with direct change when you can) ?

backend/doc/api.md
```diff
@@ -0,0 +1,188 @@
...
+    "user_id": 1
+}
+
+```
+
+Then you just need to resend cookie using `Cookie` header like any browser do
+automatically to have a temporary access.
+
+```
+$ http --session=/tmp/session.json GET  http://127.0.0.1:6543/api/v2/auth/whoami
```
buxx: Instead using `--session` parameter, use directly header in `http` command. It will be more excplicit.

inkhey: done.

Issue #1299, opened 2019-01-07
Old contents "page" are still pages (which make some notification not working)


tracim:
**Description**

When editing label of an old page content, no notification is send. The reason is source code is looking at content type and search for html-document (and the content has page)

Todo : add a data upgrade which convert all page into htmldocument

inkhey:
@lebouquetin should i also correct the code to work with alias mecanism (as that the main reason of the problem) or should i just create a migration script ?

tracim:
Hm... what seems to be the simplest for you? My feeling is the following one:

- if migration is done (and working well), the alias code will become useless
- currently, the alias code is working almost everywhere...

So, the options are:
- make a migration and remove the alias code
- use the alias code everywhere (and do not make the migration)

I suggest you evaluate the option of writing a migration (particularly with the json column) and if you are confident, then go for the first option, otherwise go for the second one.

What do you think?

lebouquetin:
@PhilippeAccorsi test this fix with data from our production

PhilippeAccorsi:
Maybe I have missed to update database.

PhilippeAccorsi:
@lebouquetin After migration of bdd with command `alembic -c development.ini upgrade head` Thats working good.

FI: in first step, I have misse to upgrade bdd.

PhilippeAccorsi:
Verified.

PR #1349, opened 2019-01-22
add migrations script to replace all page slug to html-document

inkhey:
related to #1299.
need to be verified on all databases supported (actually check is done only on postgres) and with better database sample.

### DIFF
--- a/backend/tracim_backend/migration/versions/f889c2b59759_remove_page_slug_from_content_revision_.py
+++ b/backend/tracim_backend/migration/versions/f889c2b59759_remove_page_slug_from_content_revision_.py
@@ -15,11 +15,13 @@ down_revision = '354d62d490ad'
 
 OLD_SLUG = 'page'
 NEW_SLUG = 'html-document'
+NEW_SLUG_FILE_EXTENSION = '.document.html'
 
 content_revisions = sa.Table(
     'content_revisions',
     sa.MetaData(),
-    sa.Column('type')
+    sa.Column('type'),
+    sa.Column('file_extension'),
 )
 
 
@@ -30,6 +32,7 @@ def upgrade():
             content_revisions.c.type == OLD_SLUG
         ).values(
             type=NEW_SLUG,
+            file_extension=NEW_SLUG_FILE_EXTENSION
         )
     )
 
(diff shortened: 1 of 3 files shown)


backend/tracim_backend/migration/versions/f889c2b59759_remove_page_slug_from_content_revision_.py
```diff
@@ -0,0 +1,46 @@
...
+            file_extension=NEW_SLUG_FILE_EXTENSION
+        )
+    )
+
+def downgrade():
+    """ Downgrade is not able to retrieve content with old slug and content
+    with new as both are valid before this. Stay with NEW_SLUG is better, so
+    don't do anything
+    """
+    pass
```
lebouquetin: Do you realize that you break downgrade-capacity of tracim, here? You can't decide as a developer to break this, this kind of decision must be discussed.

inkhey: In fact, downgrading problem is not that new. New/Updated html-document are not supported for downgrading.
 Note:  Downgrade issue happened only with tracim code older than https://github.com/tracim/tracim/commit/a05c27a7869b311755590359b2ae587f9df84d64 as old not updated page can't be read anymore in old version of tracim.

We can converted all content to page in this migration to avoid issue as downgrading. This may work but that's not really correct. In fact, migration tree lack a "downgrading only" migration to revert all html-document to page.

lebouquetin: I did not see any documentation or warning about impossibility to downgrade.

> We can converted all html-document to page in this migration to avoid issue as downgrading. This may work but that's not really correct. In fact, migration tree lack a "downgrading only" migration to revert all html-document to page at the correct level.

This is your work to identify this kind of issue when you implement features / refactoring. You should have added a migration, or at least document that downgrade is not possible in this specific case

lebouquetin: So, please add a documentation information which explain that downgrade from tracim v2 to tracim v1 (if it's this case) is not supported

lebouquetin: then the fix will be ok (to be tester by @PhilippeAccorsi with our production dump)

Issue #1359, opened 2019-01-23
Cant reset password if never logged in


lebouquetin:
Currently: 
- someone created an account for me on tracim
- i receive an email with login and password
- i lost the email
- i go to tracim in order to reset my password
- i get an error saying that auth type is unknown

(which seems logical with the current strategy of logging / managing several authentication backend)

Expected:

- find a way to allow this. When creating an account through admin interface, the authentication type should be "internal", doesn't it ? (in ldap or apache auth cases, the account is created on the fly in tracim, no?)

inkhey:
@lebouquetin When creating an account through admin interface, default type is unknown. Account type is setted to any other value as soon as the user log in first time.

 This allow to create account for other case than internal. If we don't do this you will have trouble if you want to preset right and access for a ldap/apache account before any user log in.

The solution i see to fix this issue was to accept reset password for 'unknown' account. Not sure we should set user to direcly set user to internal in that case or wait for first loggin to fix account type.

lebouquetin:
> @lebouquetin When creating an account through admin interface, default type is unknown. Account type is setted to any other value as soon as the user log in first time.

You mean auth type is unknown, right ? I understand that auth type is identified as soon as the user first log in.

> This allow to create account for other case than internal. If we don't do this you will have trouble if you want to preset right and access for a ldap/apache account before any user log in.

Nice. I understand you take into account advanced use cases, but first, we must address basic use cases. Basic use cases are:

- internal authentication
- auto-activation of an account based on ldap or apache authentication

Then, presets are a good case, but this is the third case in priority.

1. first case must allow user to reset his password (this is not working)
2. second case is when a user has dynamically created his account, he should not be able to reset the password (this is already ok, I think)
3. the presets case is difficult to manage correctly, so we can accept border cases and a "not totally nice fix" (I mean, if there are some drawbacks, its ok.

> The solution i see to fix this issue was to accept reset password for 'unknown' account. 
Good.
> Not sure if we should set user auth type directly to internal or wait for first loggin to fix account type.
I think waiting is better for now. We'll see in the future if something else is better (users will tell us).
>This solution was done in PR #1366 without modification of auth_type.
Ok, lets see

tracim:
@lebouquetin @inkhey 

I have tested this 5 cases:

- [x] Internal auth: create account in tracim => if I'm going directly to reset password email is accepted. (new bdd for this test)
[...]

PR #1366, opened 2019-01-28
Fix/1359 cant reset password if never logged in

inkhey:
related to #1359

### DIFF
--- a/backend/tracim_backend/lib/core/user.py
+++ b/backend/tracim_backend/lib/core/user.py
@@ -932,7 +932,9 @@ class UserApi(object):
             raise UserAuthTypeDisabled('user {} auth type {} is disabled'.format(user.email, user.auth_type.value))
 
     def _user_can_authenticate(self, user: User) -> bool:
-        return user.auth_type and user.auth_type in self._config.AUTH_TYPES
+        valid_auth_types = list(self._config.AUTH_TYPES)
+        valid_auth_types.append(AuthType.UNKNOWN)
+        return user.auth_type and user.auth_type in valid_auth_types
 
     def allowed_to_invite_new_user(self, email: str) -> bool:
         # INFO - G.M - 2018-10-25 - disallow account creation if no
--- a/backend/tracim_backend/lib/core/user.py
+++ b/backend/tracim_backend/lib/core/user.py
@@ -933,6 +933,10 @@ class UserApi(object):
 
     def _user_can_authenticate(self, user: User) -> bool:
         valid_auth_types = list(self._config.AUTH_TYPES)
+        # INFO - G.M - 2019-01-29 - we need to add Unknown as config doesn't
+        # list it for some reason and as unknown is a valid auth method.
+        # this fix issue 1359 :
+        # https://github.com/tracim/tracim/issues/1359
         valid_auth_types.append(AuthType.UNKNOWN)
         return user.auth_type and user.auth_type in valid_auth_types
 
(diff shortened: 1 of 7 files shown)


backend/tracim_backend/lib/core/user.py
```diff
@@ -932,7 +932,9 @@ def _check_user_auth_validity(self, user:User) -> None:
             raise UserAuthTypeDisabled('user {} auth type {} is disabled'.format(user.email, user.auth_type.value))
 
     def _user_can_authenticate(self, user: User) -> bool:
-        return user.auth_type and user.auth_type in self._config.AUTH_TYPES
+        valid_auth_types = list(self._config.AUTH_TYPES)
+        valid_auth_types.append(AuthType.UNKNOWN)
```
lebouquetin: please add a comment and a link to the issue

backend/tracim_backend/tests/functional/test_reset_password.py
```diff
@@ -36,6 +37,55 @@ def test_api__reset_password_request__ok__nominal_case(self):
...
+            params=params,
+        )
+        response = requests.get('http://127.0.0.1:8025/api/v1/messages')
+        response = response.json()
+        assert len(response) == 1
+        headers = response[0]['Content']['Headers']
+        assert headers['From'][0] == 'Tracim Notifications <test_user_from+0@localhost>'  # nopep8
+        assert headers['To'][0] == 'test user <test@test.test>'
+        assert headers['Subject'][0] == '[TRACIM] Reset Password Request'
+        requests.delete('http://127.0.0.1:8025/api/v1/messages')
```
lebouquetin: I think you do not have to manage data in your test - data must be reset between every test case

backend/tracim_backend/tests/functional/test_reset_password.py
```diff
@@ -76,6 +125,48 @@ def test_api__reset_password_request__ok__nominal_case(self):
         assert 'code' in res.json.keys()
         assert res.json_body['code'] == error.NOTIFICATION_DISABLED_CANT_RESET_PASSWORD  # nopep8
 
+    @pytest.mark.unknown_auth
```
lebouquetin: I think this test case is enough... it's almost a duplicate of functionnal test case, doesnt it ? You do not need to implement the two of them every time.

backend/tracim_backend/tests/functional/test_reset_password.py
inkhey: You're right that mailhog data as not properly managed now.

backend/tracim_backend/tests/functional/test_reset_password.py
inkhey: There is 3 different endpoints for email reset and we should be sure that unknown account can use all of them, so i'm not sure if that a good idea to not set those tests, best solution to avoid code duplication in this case can be using parameter : https://docs.pytest.org/en/latest/parametrize.html but to use this, we should probably reconvert test to real pytest first (there are unittest-like actually)

backend/tracim_backend/lib/core/user.py
inkhey: done.

backend/tracim_backend/tests/functional/test_reset_password.py
lebouquetin: two possibilities:

1. if easy, move the `requests.delete` in `tearDown()`
2. if not easy, add a TODO to the code

inkhey: done, cleanup all mailhog test.

Issue #1358, opened 2019-01-23
Workspace invitation must result in notification subscription by default


lebouquetin:
Currently:

if I invite a member to the current workspace, the member is invited, but the notification subscription is set to false by default

Expected: new member of a workspace must be subscripted to notifications by default.

inkhey:
@lebouquetin if user was just created with invitation, i'm not sure that set notification subscription to true as user was not connected yet is a great idea. 

I propose to set notification by default to true when adding workspace to user **but** disable notification for "unknown" auth user.

PhilippeAccorsi:
Verfiied.

PR #1367, opened 2019-01-28
Fix/1358 workspace subscription by default on invitation

inkhey:
related to #1358


backend/tracim_backend/tests/functional/test_workspaces.py
```diff
@@ -2627,6 +2626,155 @@ def test_api__create_workspace_member_role__err_400__user_not_found_as_simple_us
...
+        user_role_found = res.json_body
+        assert user_role_found['role'] == 'content-manager'
+        assert user_role_found['user_id']
+        user_id = user_role_found['user_id']
+        assert user_role_found['workspace_id'] == workspace.workspace_id
+        assert user_role_found['newly_created'] is True
+        assert user_role_found['email_sent'] is True
+        assert user_role_found['do_notify'] is True
+
+        self.testapp.authorization = (
```
lebouquetin: this is a new test case, please split your code

backend/tests_configs.ini
```diff
@@ -197,6 +197,7 @@ website.base_url = http://localhost:6543
 color.config_file_path = %(here)s/color-test.json
 user.reset_password.validity = 5
 frontend.serve = False
+email.notification.enabled_on_invitation = False
```
lebouquetin: @inkhey 
In the issue, it is requested to set notifications to true by default. In your MR you add a parameter which is False by default. It is the opposite which is asked.

backend/tests_configs.ini
```diff
@@ -342,6 +343,36 @@ email.notification.smtp.password = just_a_password
 website.base_url = http://localhost:6543
 color.config_file_path = %(here)s/color-test.json
 user.reset_password.validity = 5
+email.notification.enabled_on_invitation = False
+
+[functional_test_with_mail_test_sync_with_auto_notif]
```
lebouquetin: we do not need to test here if notifications are sent, we need to check "when a user is invited, the boolean value is True instead of False")

backend/tracim_backend/tests/functional/test_workspaces.py
```diff
@@ -2627,6 +2626,155 @@ def test_api__create_workspace_member_role__err_400__user_not_found_as_simple_us
...
+        workspace = workspace_api.create_workspace('test', save_now=True)  # nopep8
+        rapi = RoleApi(
+            current_user=admin,
+            session=dbsession,
+            config=self.app_config,
+        )
+        rapi.create_one(user, workspace, UserRoleInWorkspace.WORKSPACE_MANAGER, False)  # nopep8
+        transaction.commit()
+
+        requests.delete('http://127.0.0.1:8025/api/v1/messages')
```
lebouquetin: what is /api/v1 ?

backend/tracim_backend/tests/functional/test_workspaces.py
```diff
@@ -2627,6 +2626,155 @@ def test_api__create_workspace_member_role__err_400__user_not_found_as_simple_us
...
+        )
+        user_role_found = res.json_body
+        assert user_role_found['role'] == 'content-manager'
+        assert user_role_found['user_id']
+        user_id = user_role_found['user_id']
+        assert user_role_found['workspace_id'] == workspace.workspace_id
+        assert user_role_found['newly_created'] is True
+        assert user_role_found['email_sent'] is True
+        assert user_role_found['do_notify'] is True
+
```
lebouquetin: All these asserts should be simplified:

        assert user_role_found['role'] == 'content-manager'
        assert user_role_found['user_id'] == user.user_id
        assert user_role_found['workspace_id'] == workspace.workspace_id
        assert user_role_found['do_notify'] is True 

Other stuff is not required for the test case.

backend/tracim_backend/tests/functional/test_workspaces.py
```diff
@@ -2627,6 +2626,155 @@ def test_api__create_workspace_member_role__err_400__user_not_found_as_simple_us
...
+        )
+        res = self.testapp.get(
+            '/api/v2/users/{}'.format(user_id),
+            status=200,
+        )
+        res = res.json_body
+        assert res['profile'] == 'users'
+
+        # check mail received
+        response = requests.get('http://127.0.0.1:8025/api/v1/messages')
```
lebouquetin: what is /api/v1?

backend/tests_configs.ini
lebouquetin: I understand that this is a test file, but, what is the interest to put this here ? By the way, you did not update the reference configuration file (development.ini.sample), so adding this parameter is not visible by user, so he will never configure it, so it is useless to implement a parameterizable behavior.

If you put the parameter in one place only, this is in the development.ini.sample, not in test stuff.

backend/tracim_backend/tests/functional/test_workspaces.py
inkhey: that's mailhog endpoint to get messages.

backend/tracim_backend/tests/functional/test_workspaces.py
inkhey: mailhog endpoint.

inkhey:
> Note : we do not need to test here that notifications are sent. We need to test "user is correctly configured".

Goal of this test is to be sure that "unknown" user don't receive notification. That's not the same goal. User can be correctly configured, if he received mail before being connected a first time, that's not a correct behaviour.

At this kind of case is complex and take time to check manually, i prefer to have tests for them.

backend/tests_configs.ini
inkhey: User can be invited but not connected a first time yet. that the goal of this test.

backend/tracim_backend/tests/functional/test_workspaces.py
lebouquetin: ok

lebouquetin:
- add a unit test for function `get_notifiable_roles()` if user has not yet logged in
- add `email.notification.enabled_on_invitation = True` to development.ini.sample

Issue #1376, opened 2019-01-30
WEBDAV: move folder inside another folder in same workspace not working


PhilippeAccorsi:
At present

If you try to move folder in another folder in same woorkspace, you have an error.
Tested with windows webdav and heavy client (WinSCP, CyberDuck)

Error 500 visible in winscp.
In tracim_webdav.log:
~~~
11:30:48 INFO  [sqlalchemy.engine.base.Engine:1154][b'uWSGIWorker1Core0'] {'param_1': 8186}
11:30:48 DEBUG [sqlalchemy.engine.base.Engine:680][b'uWSGIWorker1Core0'] Col ('content_revisions_revision_id', 'content_revisions_content_id', 'content_revisions_owner_id', 'content_revisions_label', 'content_revisions_description', 'content_revisions_file_extension', 'content_revisions_file_mimetype', 'content_revisions_depot_file', 'content_revisions_properties', 'content_revisions_type', 'content_revisions_status', 'content_revisions_created', 'content_revisions_updated', 'content_revisions_is_deleted', 'content_revisions_is_archived', 'content_revisions_is_temporary', 'content_revisions_revision_type', 'content_revisions_workspace_id', 'content_revisions_parent_id')
11:30:48 DEBUG [sqlalchemy.engine.base.Engine:1105][b'uWSGIWorker1Core0'] Row (17880, 8186, 233, '3', '', '', '', None, '', 'folder', 'open', datetime.datetime(2019, 1, 30, 11, 22, 33, 235643), datetime.datetime(2019, 1, 30, 11, 22, 33, 235657), False, False, False, 'creation', 85, None)
<139872224270208> --- MOVE Response(500 Internal Server Error): ---
Content-Type: 'text/html'
Content-Length: '484'
Date: 'Wed, 30 Jan 2019 11:30:48 GMT'

[...]

PR #1385, opened 2019-01-31
Fix move of folder in webdav (doesn't fix char issue)

inkhey:
related to #1376 
todo : 
- [X] add webdav functional test

this does not fix move in specific case when special characters are in use, see https://github.com/tracim/tracim/issues/1399


backend/tracim_backend/lib/core/content.py
```diff
@@ -1477,6 +1477,9 @@ def move(self,
             if new_parent and new_parent.workspace_id != item.workspace_id:
                 raise ValueError('the item should stay in the same workspace')
 
+        if not new_workspace:
+            new_workspace = new_parent.workspace
+
         if (new_workspace and new_parent) and \
```
lebouquetin: with the 3 lines above, you can remove the check about new_workspace here:

```
         if (new_workspace and new_parent) and \ 
             new_parent.workspace_id != new_workspace.workspace_id: 
```

Can be written:

```
         if new_parent and new_parent.workspace_id != new_workspace.workspace_id: 
```

backend/tracim_backend/lib/webdav/resources.py
```diff
@@ -587,14 +587,13 @@ def move_folder(self, destpath):
...
                     self.content_api.update_content(self.content, webdav_convert_file_name_to_bdd(basename(destpath)))
                     self.content_api.save(self.content)
-                else:
-                    if destination_workspace.workspace_id == self.content.workspace.workspace_id:
-                        self.content_api.move(self.content, destination_parent)
-                    else:
-                        self.content_api.move_recursively(self.content, destination_parent, destination_workspace)
+                # move file if needed
+                if destination_workspace != self.content.workspace or destination_parent != self.content.parent :
+                    self.content_api.move_recursively(self.content, destination_parent, destination_workspace)
```
lebouquetin: As we talked, the final code should be something like this:

```
if basename(destpath) != self.getDisplayName(): 
    [do some stuff]

if some_move_is_in_progress:
    if destination_workspace.workspace_id == self.content.workspace.workspace_id:
        self.content_api.move(self.content, destination_parent) 
    else: 
        self.content_api.move_recursively(self.content, destination_parent, destination_workspace) 
```

lebouquetin: + add `move_recursively()` docstring explaining that it is required for moving content from a workspace to another one (if stay in the same workspace, child items do not move as their parent_id and workspace_id stay the same)

inkhey: After some search it seems that there is also a bug in normal move. Api doesn't use move_recursive when needed.

I do think if we keep this behaviour, we should refactor this better. i've made a quick poc (with a test): https://github.com/tracim/tracim/commit/60d77ec2ed4165d079088821d0038baad2ee248a

@buxx seems to not like the behaviour because it create new revisions, so he want we talk about this.

buxx: @lebouquetin Inkhey was talking with me about some thing: When Content is moved to an other workspace, make a new_revision with news workspace_id is not enough: Comments must be updated with update oh their workspace_id.

Do we have architecture problem ? I think it is not normal to update all comments "workspace_id" when parent content is moved to an other workspace. Comment should only related to the related content no ?

backend/tracim_backend/lib/core/content.py

backend/tracim_backend/lib/webdav/resources.py
inkhey: see https://github.com/tracim/tracim/issues/1421 for architecture problem and https://github.com/tracim/tracim/issues/1419 for repairing of recursive move in api.

Issue #1295, opened 2018-12-20
on admin user app, remove message "email notif activated"


Skylsmoi:
the email notification activated is the default behavior, there is no need to remind the user that it is on that page

PhilippeAccorsi:
Verified.

Issue #1327, opened 2019-01-16
Cypress test: test not working good all the time tiny_mce__can_type_content_spec.js


PhilippeAccorsi:
**At present**

The test "app_html-document/tiny_mce__can_type_content_spec.js" is not working correctly each time is starting. Sometime you just need to restart this test and its working

~~~
Running: app_html-document/tiny_mce__can_type_content_spec.js...                       (19 of 53) 
  Known users as a workspace-manager
  1) Type into tiny mce and save from dashboard
  ✓ Type into tiny mce and save (14127ms)
  2) Check if TinyMCE open content of the document
  ✓ Check TinyMCE is active if document is empty when opening (13888ms)
  ✓ Check TinyMCE is not active when opening if document has been modified (17101ms)
  3 passing (1m)
  2 failing
  1) Known users as a workspace-manager Type into tiny mce and save from dashboard:
     CypressError: Timed out retrying: cy.click() failed because this element is disabled:
<button type="button" data-cy="editionmode__button__submit" class="html-document__editionmode__submit editionmode__button__submit btn highlightBtn" disabled="" data-radium="true" style="background-color: rgb(0, 204, 0); color: rgb(253, 253, 253); border-color: rgb(0, 204, 0);">Validate</button>
Fix this problem, or use {force: true} to disable error checking.
https://on.cypress.io/element-cannot-be-interacted-with
      at Object.cypressErr (http://localhost:1337/__cypress/runner/cypress_runner.js:65238:11)
      at Object.throwErr (http://localhost:1337/__cypress/runner/cypress_runner.js:65203:18)
      at Object.throwErrByPath (http://localhost:1337/__cypress/runner/cypress_runner.js:65230:17)
      at Object.retry (http://localhost:1337/__cypress/runner/cypress_runner.js:58775:16)
      at retryActionability (http://localhost:1337/__cypress/runner/cypress_runner.js:50564:19)
      at tryCatcher (http://localhost:1337/__cypress/runner/cypress_runner.js:127090:23)
      at Function.Promise.attempt.Promise.try (http://localhost:1337/__cypress/runner/cypress_runner.js:124464:29)
      at tryFn (http://localhost:1337/__cypress/runner/cypress_runner.js:59011:24)
      at whenStable (http://localhost:1337/__cypress/runner/cypress_runner.js:59039:12)
      at http://localhost:1337/__cypress/runner/cypress_runner.js:58799:16
[...]

PR #1408, opened 2019-02-07
[#1295] removed msg 'email notif activated' on app admin + fixed tran…

Skylsmoi:
…slation

https://github.com/tracim/tracim/issues/1295


Skylsmoi:
the CI that failed is due to the tinymce test that randomly fails.
I have pushed a commit to disable it until proper fix (in https://github.com/tracim/tracim/issues/1327)

Issue #1427, opened 2019-02-08
fix children function in content model


inkhey:
#### Currently
Children method in content model doesn't work as expected, it get previous children (content that are been moved somewhere else) too.

#### Expected:
Return only current children.

PhilippeAccorsi:
Verified.

PR #1428, opened 2019-02-08
fix children function in content model

inkhey:
related to #1427

### DIFF
--- a/backend/tracim_backend/models/data.py
+++ b/backend/tracim_backend/models/data.py
@@ -1160,7 +1160,7 @@ class Content(DeclarativeBase):
         valid_children = []
         for revision in all_childrens_revisions:
             if revision.revision_id in all_up_to_date_revisions_ids:
-               valid_children.append(revision)
+               valid_children.append(revision.node)
 
         return valid_children
 
--- a/backend/tracim_backend/models/data.py
+++ b/backend/tracim_backend/models/data.py
@@ -1150,12 +1150,13 @@ class Content(DeclarativeBase):
         """
         # TODO - G.M - 2019-02-08 - Refactor this code to be more efficient
         all_childrens_revisions = object_session(self).query(ContentRevisionRO).filter( ContentRevisionRO.parent_id == self.content_id).all()
+        children_content_ids = set([cr.content_id for cr in all_childrens_revisions])
         all_up_to_date_revisions_result = object_session(self).query(
             func.max(ContentRevisionRO.revision_id)
-            ).group_by(ContentRevisionRO.content_id).all()
+        ).filter(ContentRevisionRO.content_id.in_(children_content_ids))\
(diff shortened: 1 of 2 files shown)


backend/tracim_backend/models/data.py
```diff
@@ -1139,19 +1142,26 @@ def owner(self, value: User) -> None:
...
-        # check if possible to do this better, probably using database sorting
-        # and sql distinct.
-        revisions = set([revision.node for revision in self.children_revisions])
-        revisions = sorted(revisions, key=lambda r: r.revision_id)
-        return revisions
+        # TODO - G.M - 2019-02-08 - Refactor this code to be more efficient
+        all_childrens_revisions = object_session(self).query(ContentRevisionRO).filter( ContentRevisionRO.parent_id == self.content_id).all()
+        all_up_to_date_revisions_result = object_session(self).query(
+            func.max(ContentRevisionRO.revision_id)
+            ).group_by(ContentRevisionRO.content_id).all()
```
lebouquetin: you take the full database, here. What about at least filtering by content_id found in all_children_revisions ?

Here is the code

```
all_childrens_revisions = object_sessio  # unchanged
children_content_ids = set([cr.content_id for cr in all_childrens_revisions])
all_up_to_date_revisions_result = object_session(self).query(
    func.max(ContentRevisionRO.revision_id)
    ) \
    .filter(ContentRevisionRO.content_id.in_(children_content_ids) \
    .group_by(ContentRevisionRO.content_id).all()

File: backend/tracim_backend/models/data.py [hunk 2]
```diff
@@ -1139,19 +1142,26 @@ def owner(self, value: User) -> None:
...
+
+        all_up_to_date_revisions_ids = [value[0] for value in all_up_to_date_revisions_result]
+
+        valid_children = []
+        for revision in all_childrens_revisions:
+            if revision.revision_id in all_up_to_date_revisions_ids:
+               valid_children.append(revision)
```
lebouquetin: note : in the original code you returned revision.node (which is a Content), here you seems to return ContentRevisionRO

backend/tracim_backend/models/data.py
inkhey: nice idea. object_session is a method to retrieve session context from sqlalchemy object, so it's really all database. Thats really different from using query object from libContent where there are filter.

backend/tracim_backend/models/data.py
inkhey: You're right. I have to fix this.

backend/tracim_backend/models/data.py
inkhey: done.

backend/tracim_backend/models/data.py
inkhey: done;

PR #1476, opened 2019-03-13
Fixed some typos in french translation

Nioub:
Just fixed some typos in the french translation.

Best regards,


buxx:
Hello @Nioub , thank's for this patch.

@Skylsmoi or @PhilippeAccorsi Can you take a look about cypress test fail ? We have to determinate if test fail because the changes or if it is a "random fail". If it is a "random fail", please disable the test and open an issue to fix it. If it is a fail because changes, please fix the test.

PhilippeAccorsi:
Issue exist here: #1327 
We need to discuss with @lebouquetin to define if we need to solve this issue for v2.2 or if we can disable this test.

lebouquetin:
What happens if we restart the tests?

PhilippeAccorsi:
Generally it works. Sometime we need to restart 2 or 3 time more.

lebouquetin:
So you can accept the MR if you are pretty sure the failure comes from our unstable test case.

Issue #1495, opened 2019-03-21
Allow to setup default lang in backend


inkhey:
currently default lang in backend is not settable, we should add a parameter "default_lang" in config to allow force default lang.

PhilippeAccorsi:
**For testing:**
- uncomment `default_lang = en`
- change "en" by "fr"
- create one notification
=> you can see email in french
- change "fr" by "en"
- create one notification
=> you can see email in english

PhilippeAccorsi:
Verified.

PR #1496, opened 2019-03-21
Fix/1495 add default lang param in config

inkhey:
related to #1495


backend/tracim_backend/lib/utils/translation.py
```diff
@@ -26,7 +26,10 @@ class Translator(object):
     def __init__(self, app_config: 'CFG', default_lang: str = None, fallback_lang: str = None):  # nopep8
         self.config = app_config
         if not fallback_lang:
-            fallback_lang = DEFAULT_FALLBACK_LANG
+            if app_config.DEFAULT_LANG:
+                fallback_lang = app_config.DEFAULT_LANG
+            else:
+                fallback_lang = DEFAULT_FALLBACK_LANG
```
buxx: It seems useless to manage case where `app_config.DEFAULT_LANG` is not set because this config is always set...

inkhey: done.

backend/tracim_backend/lib/utils/translation.py
```diff
@@ -21,12 +21,15 @@ def translator_marker(string: str) -> str:
 class Translator(object):
     """
     Get translation from json file
+    you should provide either valid fallback_lang(true value) or valid
```
lebouquetin: this comment should be the docstring of the __init__, not the docstring of the class, I think

inkhey: done.

Issue #1475, opened 2019-03-13
Some parameter in development.ini.sample are missing (since november 2018)


PhilippeAccorsi:
## Bug description & symptom

Some parameter for preview generator are missing in development.ini.sample so you cant change resolution for preview.

## Diagnostic

When we have refactor this file, we have lost some necessary parameter.

## Expected

- [ ] All necessary parameter are comeback

## Impacted components

- *backend*

PhilippeAccorsi:
This issue is related to https://github.com/tracim/tracim/pull/1208
Missing parameter come from this commit https://github.com/tracim/tracim/pull/1208/commits/d8483aece3890dae3ee73dd251b0eecb2bf1c192

inkhey:
@PhilippeAccorsi #1479 
i just restore params for restricted/allowed dims in config file.
but frontend doesn't support restricted size well : https://github.com/tracim/tracim/issues/1480

PhilippeAccorsi:
Syntaxe problem:

## You can parametrized allowed jpg preview dimension list, if not set, default
## is 256x256. First {width}x{length} items is default preview dimensions.
## all items should be separated by ',' and you should be really careful to do
## set anything else than '{int}x{int}' item and ', ' separator
; preview.jpg.allowed_dims = 256x256,512x512,1024x1024
## Preview dimensions can be set as restricted, if set as restricted, access
## endpoint to  to get any other preview dimensions than allowed_dims will
## return error
; preview.jpg.restricted_dims = True

PhilippeAccorsi:
Verified.

PR #1543, opened 2019-04-03
fix syntaxe about size preview information

PhilippeAccorsi:
Change some wrong syntaxe in development.ini.sample . Write with @lebouquetin .


PhilippeAccorsi:
related to #1475

Issue #1524, opened 2019-03-27
Refuse to launch tracim/webdav/caldav if paths given does not exist


inkhey:
## Currently

- preview_generator dir is created automatically
- caldav path existence is mandatory (radicale branch)
- session data dir is created automatically (needed for beaker)
- depot dir is created automatically

## Expected : 
- processus (tracim,caldav,webdav, ...) return error if one of this file doesn't exist as file and is not writable. Raise error probably during config settings.
- Explicit error : propose user to change config or create dir/ change permission

inkhey:
#1544 allow to fix this issue, but cause lot of trouble for tests.
Currently: 
- [X] Backend test  : we force a code which default those path parameters in test with autocreated file and default value which can be overrided by environnement var, this means that we can more easily configure test behaviour. 
- [X] Command Backend test: we change the behaviour of command and command test to allow override default param, so we can use same env var like for other backend tests.
- [ ] Cypress functionnal test : We launch tracim as 'normal' with a specific conf file, so env var doesn't work, launch failed because of not finding params or folder existence, like normal tracim behaviour.

inkhey:
after discussion, we should do #1550 before finished work on this.

PhilippeAccorsi:
@inkhey is it testable or not?

inkhey:
#1687 does:
- makes paths mandatory for running tracim.
- add CALDAV__RADICALE__STORAGE__FILESYSTEM_FOLDER as tracim param.
- update travis file to user env variables
- add some log for travis debugging.
- add TEST_MIGRATION_SCRIPT_LOCATION env var and TEST_CONFIG_FILE_PATH env var mecanism that allow to run test from anywhere (not just /backend) if env var are correctly setted. This allow to fix https://github.com/tracim/tracim/issues/1083.
- fix https://github.com/tracim/tracim/issues/1083 by doing coverage and test from depot root dir instead of /backend.
- add one working default dir tree for test in backend/ and one other for run in backend/ + add fix related to allow not handled file when deleting depot in tests.

inkhey:
@lebouquetin Current behavior was to have default dir tree for both test and run to have something running "by default", is this behavior ok, or should we do in script somewhere ?

inkhey:
@lebouquetin https://github.com/tracim/tracim/issues/1525 is about doing this by script. For running that's ok, but what about tests ? should a test "fixture"(we have not yet pytest-style test) create those dir ? 
[...]

PR #1544, opened 2019-04-03
WIP: Fix/1524 make paths mandatory

inkhey:
related to #1524
:warning:  Deprecated, redo same mecanism with new config file.


inkhey:
deprecated PR, check #1687 instead.

PR #1552, opened 2019-04-03
Fix/1539  caldavzap save timezone

Skylsmoi:
added the timezone fixes for initialization, saving and design


lebouquetin:
Ok for me. Can be merge as soon as CI is ok

PR #1459, opened 2019-02-27
Fix/1154  check empty htlml tags in comments

basilelegal:
check empty htlml tags in comments

Ref #1154

Issue #1154, opened 2018-11-19
Check for empty comment does not verify html markup case


inkhey:
**Describe the bug**
Currently check for empty comment verify only case when content given is completely empty. As comment content can be html, we should not allow empty comment contaning only html tag without any meaning of consistent content (like `<p>`)
This issue will probably need https://github.com/tracim/tracim/issues/1101 to be resolved first in order to simplify case.

PhilippeAccorsi:
Verified.

inkhey:
@basilelegal your fix doesn't have the correct behaviour, so, we reverted all you commit for this fix. Can you fix this ?

PhilippeAccorsi:
We have reverted also about big regression for image or media link with tinymce. If this kind of content is send, the backend don't save value and you cant see image or media link in timeline.

gbonaspetti:
Still not doing:

- [ ]  show explicit message with error code 2003

- [ ]  translate this error message

(Linked to #1553)

lebouquetin:
I mark this "need test" because #1553 is need test too

gbonaspetti:
Verified.

PR #1557, opened 2019-04-04
Fix/1154  revert modification from pr 1459

inkhey:
#1459 cause trouble fixing #1154 , we revert those commit until a correct fix.


PhilippeAccorsi:
Tested and working.

Issue #1556, opened 2019-04-04
Some improvements to calendar (agenda)


Skylsmoi:
## Feature description

- [x] Rename calendar into agenda in all tracim
  - [x] frontend
  - [x] backend
- [x] set default padding to calendar to match all other pages
- [x] add button fullscreen to agenda
- [x] remove btn "scroll to top" (see https://github.com/tracim/tracim/issues/1554)
- [x] define a color for calendar to set in color.json.sample
- [x] rename title of agenda pages to differentiate all calendar from one workspace's calendar
- [x] add .ics url of own calendar in my account page
- [x] add .ics url of workspace in it's dashboard
- [x] button agenda in header should be have some intelligence (hide when not logged and when calendars are disabled through backend conf (get info from list in app list))
- ~disable btn to edit calendar it user doesn't have enough permissions~ This task has been moved to this issue https://github.com/tracim/tracim/issues/1581

## Context

Some improvements to calendar

## Impacted components
app agenda
app frontend (core)

Skylsmoi:
adding @inkhey and backend tag so that he can include the modifications for renaming agenda in backend in the same branch

Skylsmoi:
set default agenda color to #f63434

PhilippeAccorsi:
Verified.

PhilippeAccorsi:
@Skylsmoi (after discuss with you) I reopen this issue because I have detected a bug with agenda in my account page.

The problem is when you deactivate radicale in development.ini, when you go in My account page, you can see an error 404 :
"Error while loading your agenda"
`{"message": "/api/v2/users/me/agenda is not a valid path", "details": {"error_detail": {}}, "code": null}`

I think it's because the link is not show in this case but still exist in dom.

Skylsmoi:
account page wont ask api for agenda list if app is deactivated in backend

PhilippeAccorsi:
Verified.

PR #1561, opened 2019-04-04
Fix/1556 rename calendar to agenda

inkhey:
related to #1556 
:warning: merge on frontend #1556 branch to check full compat of new code.


backend/tracim_backend/config.py
```diff
@@ -483,16 +483,16 @@ def __init__(self, settings: typing.Dict[str, typing.Any]):
             'caldav.radicale_proxy.base_url',
             None
         )
-        self.CALDAV_RADICALE_CALENDAR_DIR = 'calendar'
+        self.CALDAV_RADICALE_AGENDA_DIR = 'agenda'
```
lebouquetin: is this `calendar` folder name is standard for radicale, it may stay `calendar` (if no specific value, keep agenda)

backend/tracim_backend/models/context_models.py
```diff
@@ -668,11 +668,11 @@ def is_deleted(self) -> bool:
         return self.workspace.is_deleted
 
     @property
-    def calendar_enabled(self) -> bool:
+    def agenda_enabled(self) -> bool:
```
lebouquetin: Please add an an explanation (a comment) about why the property is not renamed

backend/tracim_backend/config.py
inkhey: it is not that just a config to set first level dir for both proxy and radicale :
`/agenda` in tracim should match `/agenda` in radicale.

backend/tracim_backend/models/context_models.py
inkhey: it is .

Issue #1550, opened 2019-04-03
Refactoring CFG file to separate setting from check and allow overriding by env var


inkhey:
**Currently**:
- No way to set config through env var in tracim
- Code for checking validity of variable and code for setting them is mixed.

Allow setting easily env var will make tests easier to custom.
CFG object will be cleaner.

**Expected**:
- Separated get value from config file/env var and check validity in CFG.
- add generic mecanism to override config through env var.
- Added: set var using get_raw_config() method, see https://github.com/tracim/tracim/issues/1550#issuecomment-480172784
- Added: better logging for config: source of config applied, etc...
- Rename all parameter with similar pattern: to upper and  "." -> "__" .
- Verify all parameters setted with get_raw_config exist as parameter of CFG object .

inkhey:
@lebouquetin There is an issue with making env var mecanism generic
We have in entry always string but parameter should be something converted sometime to int, list or other type. We should know for all param how to convert them in both situation: when setting CFG from settings dict (from config file) and from env var value, currently we set to int explicitly for all parameters, so we can't make it generic for env var.

@buxx propose to change CFG object with a list of spec:
```
    available_configs = {
      'APP_ENABLED': ConfigProps(type_=ConfigList, setting_param='app.enabled', legacy_setting_param='app_enabled', env_var_name='TRACIM_APP_ENABLED'),
    }
```
type_ is callable to cast str setting,
setting_param is name of the param is settings dict
legacy_setting_param is to allow some legacy param for retrocompat with logging warning message,
env_var_name is just an example, we can this way decide to force a specific env_var name for parameter, not sure we need it (use convention like 'TRACIM_{}' prefixed env var is ok).

lebouquetin:
could you give a concrete example of current config parameter which is a problem?

inkhey:
@lebouquetin A simple example : 

in code we have : 
```
        self.EMAIL_REPLY_CONNECTION_MAX_LIFETIME = int(settings.get(
            'email.reply.connection.max_lifetime',
            600,  # 10 minutes
        ))
```
If we want to support env_var, we should do something like:
```
self.EMAIL_REPLY_CONNECTION_MAX_LIFETIME = int(os.env.get('TRACIM_EMAIL_REPLY_CONNECTION_MAX_LIFETIME'))
```
[...]

PR #1570, opened 2019-04-05
Fix/1550  refactoring cfg separate get value from check and env var

inkhey:
related to #1550


backend/tracim_backend/lib/utils/utils.py
```diff
@@ -218,7 +218,8 @@ def lighten(self):
 def string_to_list(
         base_string: str,
         separator: str,
-        cast_func: typing.Callable
+        cast_func: typing.Callable,
+        stripped: bool=False,
```
lebouquetin: stripped means "les espaces ont déjà été enlevés". what you want is `do_strip` : if true then string items will be stripped, else they won't). So rename it to `do_strip`

backend/tracim_backend/lib/utils/utils.py
```diff
@@ -228,12 +229,16 @@ def string_to_list(
     :param base_string: entry string which should be converted.
     :param separator: string separator,
     :param cast_func: all item should be casted to this function, this help
+    :param stripped: strip string (remove beginning and ending whitespace) of
```
lebouquetin: if the parameter is a boolean, then the explanation must be : "if True xxx else xxx". Your explanation is not clear here, that if true then it will strip otherwise it wont

inkhey: done.

backend/tracim_backend/lib/utils/utils.py
inkhey: done.

backend/tracim_backend/lib/utils/utils.py
```diff
@@ -218,7 +218,8 @@ def lighten(self):
 def string_to_list(
         base_string: str,
         separator: str,
-        cast_func: typing.Callable
+        cast_func: typing.Callable,
```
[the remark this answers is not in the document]
inkhey: done with typing.

backend/tracim_backend/config.py
```diff
@@ -758,12 +855,33 @@ def _set_default_app(self, enabled_app_list: typing.List[str]):
         # app_list is updated.
         update_validators()
 
+    # INFO - G.M - 2019-04-05 - Others methods
+    def _check_consistency(self):
```
inkhey: @buxx Can you check this method, its one of the important point. I do not do a specific error because Attribute Error return seems to me clear, but i want feedback about it

buxx: I think you can add `try except` where `_check_consistency` is used to be able to print an obvious message for user.

Issue #1573, opened 2019-04-08
Radicale proxy issue about anonymous access: Thunderbird caldav mod doesn't work


inkhey:
## Bug description & symptom

Thunderbird doesn't allow to be set up with "caldav" parameter (ics file work).

Radicale itself return error:
```
17:14:27 INFO  [radicale:409][waitress] PROPFIND request for '/calendar/workspace/2/' with depth '0' received from 127.0.0.1 using 'Mozilla/5.0 (X11; Linux x86_64; rv:60.0) Gecko/20100101 Thunderbird/60.3.0 Lightning/6.2.3'
17:14:27 INFO  [radicale:508][waitress] Access to '/calendar/workspace/2/' denied for anonymous user
17:14:27 INFO  [radicale:387][waitress] PROPFIND response status for '/calendar/workspace/2/' with depth '0' in 0.003 seconds: 401 Unauthorized
```

## Diagnostic

Currently,proxy used to radicale drop header authentification. This is a normal behaviour for a proxy, but radicale for some request case want an authentification header despite auth mecanism of radicale is set to None (so no check a all validity of user password here).

## Expected

- Do not drop header authentification in order to let radicale deal with a real auth header. We may replace it with a custom dumb auth header.

inkhey:
https://github.com/tracim/tracim/pull/1574 fix was verified to work (Read/Write case) on : 
- thunderbird (behaviour when reader and try to delete event is error message popup)
- caldav sync (behaviour when reader and try to delete event is false positive, content not truly deleted but client think it is).

PhilippeAccorsi:
Tested with OpenSync on Android and its work good.

PhilippeAccorsi:
Some test with user with role reader and try to delete event:

With thunderbird error appear if I tried to delete event "modification_failed"
With OpenSync no error appear but after synchronize event is come back
With davdroid with this new name https://www.davx5.com/  (also available in F-droid )no error appear but after synchronize event is come back

With this other client, we have crazy situation:
With CalDAV-sync no error appear and after synchronize event is not visible on phone but visible in other support.
With CalDAV Sync Adapter no error appear and after synchronize event is not visible on phone but visible in other support.

With this two app, its not possible to see again the event we have tried to delete.

PhilippeAccorsi:
Verified.

PR #1574, opened 2019-04-08
override default request header to drop to allow auth header

inkhey:
related to #1573


backend/tracim_backend/lib/proxy/proxy.py
```diff
@@ -39,24 +39,28 @@ def __init__(
         base_address: str,
         default_request_header_to_drop: typing.List[str] = DEFAULT_REQUEST_HEADER_TO_DROP,
         default_response_header_to_drop: typing.List[str] = DEFAULT_RESPONSE_HEADER_TO_DROP,
+        auth: typing.Optional[tuple] = None
```
lebouquetin: "optionnal tuple" does not help. What is the expected content?

lebouquetin: by the way, a list of headers to drop  must be named correctly in the plural form. Eg `default_request_headers_to_drop` instead of `default_request_header_to_drop`

inkhey: @lebouquetin Expected content is content following the spec of auth parameter of request of request lib. I didn't see much information than it want optional tuple, internal request lib check is about tuple or None.

inkhey: @lebouquetin plurals forms of "headers" parameters fix.

inkhey: maybe, should we decide, that we allow only str : `auth: typing.Optional[typing.Tuple[str, str]]` ?

lebouquetin: something like this: `auth: typing.Optional[typing.Tuple[str, str]]  # example: [login, password]`

backend/tracim_backend/lib/proxy/proxy.py
```diff
@@ -6,6 +6,8 @@
 import requests
 from pyramid.response import Response as PyramidResponse
 from requests import Response as RequestsResponse
+from requests.auth import HTTPBasicAuth, AuthBase
+
 from tracim_backend.lib.utils.request import TracimRequest
 
 HOP_BY_HOP_HEADER_HTTP = (
```
lebouquetin: What is this HOP_BY_HOP?!

backend/tracim_backend/views/agenda_api/radicale_proxy_controller.py
```diff
@@ -20,7 +20,11 @@
     RadicaleWorkspaceSubItemPathSchema
 from tracim_backend.views.core_api.schemas import UserIdPathSchema
 from tracim_backend.views.core_api.schemas import WorkspaceIdPathSchema
-
+from requests.auth import HTTPBasicAuth
+# dumb auth parameter, just to avoid radicale issue with empty
+# auth header
+RADICALE_HTTP_AUTH_USERNAME = 'tracim'
+RADICALE_HTTP_AUTH_PASSWORD = 'tracim'
```
lebouquetin: use `tracimpass` instead of `tracim` as a password: this make the code easier to understand

backend/tracim_backend/lib/proxy/proxy.py
lebouquetin: note: on line 23, this is not a `FIXME` but a `HACK` comment -we have no reason for the moment to remove this hack

inkhey: Those are header which should not be forwarded by proxy, see also :
 fr : https://developer.mozilla.org/fr/docs/Web/HTTP/Headers
en : https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers

> These headers are meaningful only for a single transport-level connection and must not be retransmitted by proxies or cached

inkhey: just added a comment about that in code, with link to RFC : https://tools.ietf.org/html/rfc2616#page-92

inkhey: done.

backend/tracim_backend/views/agenda_api/radicale_proxy_controller.py
inkhey: done.

backend/tracim_backend/lib/proxy/proxy.py
lebouquetin: Perfect. I'm pretty sure developers who will maintain this code are even not knowing that there are headers named hop-by-hob - so a link to the documentation is perfect.

Issue #1575, opened 2019-04-08
Renaming shared space or public name of user not rename associated agenda


PhilippeAccorsi:
## Bug description & symptom

If I rename workspaces, the associated agenda is not renamed.
if I rename public name of user, the associated agenda is not renamed.

## Diagnostic

The calendar is created but not updated.
We need to force update of caldav calendar props when user/workspace is updated

## Expected

- [ ] renaming workspace rename also associated agenda ( renaming using webdav should work also :warning:)
- [ ] renaming public name of user rename also associated agenda

## Prerequisites

Start tracim with caldav.

## How to reproduce

- Create agenda for one workpsace
- rename this workspace
or 
- change public name of user

PhilippeAccorsi:
Issue updated with public name information.

PhilippeAccorsi:
Verified.

PR #1578, opened 2019-04-09
Fix/1575  renaming caldav calendar name according to new workspace/user name

inkhey:
related to #1575 
need #1573 (caldav python client return error without it)


backend/tracim_backend/command/caldav.py
```diff
@@ -92,6 +92,10 @@ def take_app_action(
                 nb_error_agenda_access += 1
                 print('Cannot access to agenda server: connection error.')
                 logger.exception(self, exc)
+            except Exception as exc:
```
lebouquetin: @inkhey all the intelligence of this operation must be in an internal module, not in a command. I should be able to create all agendas out of a CLI executed command.

lebouquetin: it should be in AgendaApi class

backend/tracim_backend/lib/core/workspace.py
```diff
@@ -261,6 +261,32 @@ def execute_created_workspace_actions(self, workspace: Workspace) -> None:
...
+        from tracim_backend.lib.agenda.agenda import AgendaApi
+        if self._config.CALDAV_ENABLED:
+            agenda_api = AgendaApi(
+                current_user = self._user,
+                session = self._session,
+                config = self._config
+            )
+            if workspace.agenda_enabled:
+                try:
+                    agenda_already_exist = agenda_api.ensure_workspace_agenda_exist(workspace)
```
lebouquetin: `ensure_workspace_agenda_exist` => `ensure_workspace_agenda_exists`

backend/tracim_backend/lib/core/workspace.py
```diff
@@ -261,6 +261,32 @@ def execute_created_workspace_actions(self, workspace: Workspace) -> None:
                 except AgendaServerConnectionError as exc:
                     logger.error(self, 'Cannot connect to agenda server')
                     logger.exception(self, exc)
+                except Exception as exc:
+                    logger.error(self, 'Something goes wrong during agenda create/update')
+                    logger.exception(self, exc)
+
+    def execute_update_workspace_actions(self, workspace: Workspace) -> None:
+        from tracim_backend.lib.agenda.agenda import AgendaApi
```
lebouquetin: why do you have an import here? move it or add a comment about the reason for this import to be here

backend/tracim_backend/lib/core/workspace.py
```diff
@@ -261,6 +261,32 @@ def execute_created_workspace_actions(self, workspace: Workspace) -> None:
...
+        if self._config.CALDAV_ENABLED:
+            agenda_api = AgendaApi(
```
lebouquetin: do not instantiate an object if it may be useless. agenda_api should be instanciated in the `if workspace.agenda_enabled:`

backend/tracim_backend/lib/core/workspace.py
```diff
@@ -261,6 +261,32 @@ def execute_created_workspace_actions(self, workspace: Workspace) -> None:
                 except AgendaServerConnectionError as exc:
                     logger.error(self, 'Cannot connect to agenda server')
                     logger.exception(self, exc)
+                except Exception as exc:
+                    logger.error(self, 'Something goes wrong during agenda create/update')
+                    logger.exception(self, exc)
+
+    def execute_update_workspace_actions(self, workspace: Workspace) -> None:
```
lebouquetin: what is the goal of this method? I believe it is to execute some actions after a workspace has been updated.

Two remarks:

- this method must not be explicitily used out of the WorkspaceApi class - this is an internal behavior
- the name is wrong.

Currently, you write:

```
workspace_api.do_some_stuff()
workspace_api.execute_update_workspace_actions()
```

This is wrong, because the user of workspace_api will probably miss to call `execute_update_workspace_actions`

The right pattern is the following: 

In client code:

```
workspace_api.do_some_stuff()
```

In WorkspaceApi code:

```
class WorkspaceApi(...):
  def do_some_stuff(self, workspace):
    """ this is the public API, which is a wrapper around private api + pre/post calls"""
    self.on_workspace_update(workspace)
    workspace = self._do_some_stuff(workspace)
    self.on_workspace_updated(workspace)
    return workspace

  def _do_some_stuff(self, workspace):
    """ this is the real API, which is a wrapped by public api. It is never called directly """
    [...]
    return workspace
```

backend/tracim_backend/lib/core/workspace.py
```diff
@@ -261,6 +261,32 @@ def execute_created_workspace_actions(self, workspace: Workspace) -> None:
...
+            if workspace.agenda_enabled:
+                try:
+                    agenda_already_exist = agenda_api.ensure_workspace_agenda_exist(workspace)
+                    if agenda_already_exist:
+                        logger.warning(
+                            self,
+                            'workspace {} is just created but it own agenda already exist !!'.format(workspace.user_id)
+                        )
+                except AgendaServerConnectionError as exc:
+                    logger.error(self, 'Cannot connect to agenda server')
```
lebouquetin: add url of the agenda server for helping debug

backend/tracim_backend/lib/core/workspace.py
```diff
@@ -261,6 +261,32 @@ def execute_created_workspace_actions(self, workspace: Workspace) -> None:
...
+                    logger.exception(self, exc)
```
lebouquetin: wath is `exception()` method? this is probably `log_exception()`, no? could you rename this? (or its not our code?)
By the way, what is the goal of logging the exception in this case: the error is clearly identified

backend/tracim_backend/lib/webdav/resources.py
```diff
@@ -400,9 +400,19 @@ def moveRecursive(self, destpath):
...
+                    config=self.provider.app_config,
+                )
+                workspace_api.update_workspace(
+                    workspace=self.workspace,
+                    label=webdav_convert_file_name_to_bdd(basename(normpath(destpath))),
+                    description=self.workspace.description
+                )
                 self.session.add(self.workspace)
                 self.session.flush()
+                workspace_api.execute_update_workspace_actions(self.workspace)
```
lebouquetin: remove this: it must be implicit from outside of workspace_api

backend/tracim_backend/views/core_api/account_controller.py
```diff
@@ -96,6 +96,7 @@ def set_account_infos(self, context, request: TracimRequest, hapic_data=None):
             lang=hapic_data.body.lang,
             do_save=True
         )
+        uapi.execute_updated_user_actions(user)
```
lebouquetin: same remark: this must be implicit. Do not call yourself the on_user_updated() method (by the way, rename `execute_updated_user_actions` to `on_user_updated`

backend/tracim_backend/views/core_api/user_controller.py
```diff
@@ -238,6 +238,7 @@ def set_user_infos(self, context, request: TracimRequest, hapic_data=None):
             lang=hapic_data.body.lang,
             do_save=True
         )
+        uapi.execute_updated_user_actions(user)
```
lebouquetin: remove this

backend/tracim_backend/views/core_api/workspace_controller.py
```diff
@@ -167,6 +167,7 @@ def update_workspace(self, context, request: TracimRequest, hapic_data=None):  #
             agenda_enabled=hapic_data.body.agenda_enabled,
             save_now=True,
         )
+        wapi.execute_update_workspace_actions(request.current_workspace)
```
lebouquetin: remove this

backend/tracim_backend/lib/core/workspace.py
inkhey: 2 reasons.
1 - There is already a legacy "execute_created_user_actions" which work this way. I don't want to mix thoses mecanism and i do not think that the best place to refactor this mecanism. This problem as already been notify when we were talking about "plugin mecanism", it should probably been fix with it.

2 - This method should be launched after "save" method (or inside), we think about removing it but we do not did it yet. That is a huge refactor because save method is used something like 250 time, mostly in tests.

backend/tracim_backend/lib/core/workspace.py
inkhey: It's just same as logger.exception() method from https://docs.python.org/3.7/library/logging.html .
As we override logger by our own class we can rename it but that a bit weird to have similar naming for some method and different for others.

backend/tracim_backend/lib/webdav/resources.py
inkhey: see : https://github.com/tracim/tracim/pull/1578#discussion_r274287675

backend/tracim_backend/views/core_api/account_controller.py
inkhey: se: https://github.com/tracim/tracim/pull/1578#discussion_r274287675

backend/tracim_backend/views/core_api/user_controller.py
inkhey: see : https://github.com/tracim/tracim/pull/1578#discussion_r274287675

backend/tracim_backend/views/core_api/workspace_controller.py
inkhey: see : https://github.com/tracim/tracim/pull/1578#discussion_r274287675

backend/tracim_backend/command/caldav.py
inkhey: @lebouquetin how we deal that with proper cli response in terminal ?

backend/tracim_backend/lib/core/workspace.py
inkhey: nb: This is a drawback to apply change before commit (save do just flush), in some case it may be possible to have incoherent state because commit raise an issue but post_create/update actions has been done.

backend/tracim_backend/command/caldav.py
lebouquetin: if you means `print` instead of `logging.info`, I think you can add a parameter to the constructor or to the method call which allow to output to stdout instead of logging.

If you mean something else, let's explain.

backend/tracim_backend/lib/core/workspace.py
lebouquetin: > 1 - There is already a legacy "execute_created_user_actions" which work this way. I don't want to mix thoses mecanism and i do not think that the best place to refactor this mecanism. This problem as already been notify when we were talking about "plugin mecanism", it should probably been fix with it.

The problem is :

- you increase technical debt before the refactor (but you know this is technical debt)
- the code will be broken for all developements added before the refactoring (ie if I use workspace api in a new part of code, I have no reason "not to forgot the call to this new method)

> 2 - This method should be launched after "save" method (or inside), we think about removing it but we do not did it yet. That is a huge refactor because save method is used something like 250 time, mostly in tests.

All this explanations must appear somewhere in the code - this is not information to share in a merge request.

> For naming, i just do not change the naming from v1, that not a big deal to change them.

backend/tracim_backend/lib/core/workspace.py
lebouquetin: you're right, it's a wrong idea to rename it. I do not understand the interest of this method compared to using `logger.error(xxx)`

backend/tracim_backend/views/core_api/workspace_controller.py
lebouquetin: when reading the code, I must be warned that this way of using the API is broken by design.

backend/tracim_backend/command/caldav.py
inkhey: @lebouquetin Currently log and "normal printed" result are completely different things. Logging in stdout for command can be enabled with debug mode "-d", 
Printed result are like some kind of "frontend" whereas logging is for debugging/admin purpose.
I do not think mix them is a good idea.

inkhey: @buxx Have you got an idea about this ?

lebouquetin: @inkhey what is the problem when you write "how we deal that with proper cli response in terminal ?" ? You do not expose the problem so it's difficult to find a solution.

backend/tracim_backend/lib/core/workspace.py
inkhey: @lebouquetin  logger.exception do print exception on logging, logger.error is just a error level message. For this case, we can effectively not logging the exception but "connection error" give not lot of information. 
it can mean multiple things: see http://docs.python-requests.org/en/master/_modules/requests/exceptions/ (we do a except requests.exceptions.ConnectionError as exc) . Maybe should we be more precise.

backend/tracim_backend/command/caldav.py
inkhey: @lebouquetin I talk about this with @buxx , i will propably do this with another specific log for thing that mean to be printed for console or other thing .

backend/tracim_backend/lib/core/workspace.py
inkhey: @lebouquetin At this level we do not have many information about url. It's also unclear which url you want. The one use by command or the one use by proxy. The second one may be a bit tricky to retrieve here.

backend/tracim_backend/lib/core/workspace.py

backend/tracim_backend/lib/core/workspace.py

backend/tracim_backend/lib/core/workspace.py
inkhey: I added link to new issue to fix this later https://github.com/tracim/tracim/issues/1589

backend/tracim_backend/lib/webdav/resources.py
inkhey: https://github.com/tracim/tracim/issues/1589

backend/tracim_backend/views/core_api/user_controller.py
inkhey: https://github.com/tracim/tracim/issues/1589

backend/tracim_backend/views/core_api/workspace_controller.py
inkhey: added a warning in code.

backend/tracim_backend/views/core_api/account_controller.py
inkhey: https://github.com/tracim/tracim/issues/1589

backend/tracim_backend/lib/core/workspace.py
inkhey: added a explanation in code.

backend/tracim_backend/command/caldav.py
inkhey: will be fixed later, no so trivial : https://github.com/tracim/tracim/issues/1591

backend/tracim_backend/lib/core/workspace.py
lebouquetin: - I think you should at least complete the error log with address the service tries to connect to.
- Keep the logger.exception

backend/tracim_backend/lib/core/workspace.py
inkhey: will fix later #1592

backend/tracim_backend/lib/core/workspace.py
inkhey: will fix later #1592

PR #1599, opened 2019-04-12
Add static checks on pipelines

buxx:
Still no pipelines ...  idon't understand.


tracim:
Pipelines are working: It was a yaml syntax problem in travis config ...

Issue #1600, opened 2019-04-12
sidebar of agenda not displayed if on page for every agenda but user only has it's own agenda


Skylsmoi:
currently, if we are on the "all agendas" page and the user has no worspace with activated agenda, the sidebar isn't displayed

expected, it should be displayed in that case but still be hidden when displaying the agenda of a unique workspace

PhilippeAccorsi:
Verified.

PR #1601, opened 2019-04-12
[#1600] always display sidebar caldavzap on page for all agendas

Skylsmoi:
https://github.com/tracim/tracim/issues/1600

### DIFF
--- a/frontend_app_agenda/src/container/Agenda.jsx
+++ b/frontend_app_agenda/src/container/Agenda.jsx
@@ -178,8 +178,6 @@ class Agenda extends React.Component {
 
     if (!state.isVisible || !state.userWorkspaceListLoaded) return null
 
-    console.log('state.userWorkspaceList', state.userWorkspaceList)
-
     const config = {
       globalAccountSettings: {
         agendaList: state.userWorkspaceList.map(a => ({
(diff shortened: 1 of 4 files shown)


frontend_app_agenda/src/container/Agenda.jsx
```diff
@@ -193,7 +191,8 @@ class Agenda extends React.Component {
           idWorkspace: a.agenda_type === 'private' ? '' : a.workspace_id
         }))
       },
-      userLang: state.loggedUser.lang
+      userLang: state.loggedUser.lang,
+      shouldDisplayCaldavzapSidebar: !(state.userWorkspaceList.length === 1 && state.userWorkspaceList[0].agenda_type === 'workspace')
```
lebouquetin: My point is showing/hidding the calendar list is based on the uri.

- If you are in the `/ui/agendas/` then you show the sidebar,
- if you are in `/workspaces/<id>/agenda` you hide the agenda list

In a near future, you will have the case where the calendar is out of tracim calendars, in this case we will have to show the list

Skylsmoi: I've changed it so that it's now the frontend core that tells whether or caldavzap should force display the sidebar regarding the route.

Caldavzap algo is now:
if frontend core set `forceShowSidebar: true` then always show sidebar.
if frontend core set `forceShowSidebar: false` then show sidebar **only** if there is more than one calendar to display

see https://github.com/tracim/tracim/pull/1601/files#diff-eaa4f71493ea0798083dc4ef3d2e60e9R1995

lebouquetin: A remark by the way: `:idws` means id workspace, I think... in English in must be wsid (workspace id). You should write all your stuff in English.

Issue #1374, opened 2019-01-30
Set user lang through api according to user lang


inkhey:
Currently, frontend doesn't set at all lang parameter of user. This should be done in order to allow translation of email generated by backend.

inkhey:
@PhilippeAccorsi we need user lang in backend database to correctly translate email. currently frontend doesn't set lang in user update endpoint.

PhilippeAccorsi:
Ex:

If user as "french" in webbrowser config, all tracim is visible in french. But when user receive an notification email, this email is in english. The reason is: backend is not updated with actual value of language in web interface.

To solved that:
If user does not manually change the language, email notifications will stay in English. In order to get notifications in French, change language to English, then back to French.

Skylsmoi:
The same process is required after login successful in case it's a new user

Skylsmoi:
same process added

PhilippeAccorsi:
Verified.

PR #1619, opened 2019-04-19
[#1374] added same process for login

Skylsmoi:
https://github.com/tracim/tracim/issues/1374 see last comment for added fix


frontend/src/container/Login.jsx
```diff
@@ -151,6 +154,15 @@ class Login extends React.Component {
     if (fetchGetWorkspaceList.status === 200) props.dispatch(setWorkspaceList(fetchGetWorkspaceList.json))
   }
 
+  setDefaultUserLang = async loggedUser => {
+    const { props } = this
+    const fetchPutUserLang = await props.dispatch(putUserLang(loggedUser, props.user.lang))
+    switch (fetchPutUserLang.status) {
+      case 200: break
+      default: props.dispatch(newFlashMessage(props.t('Error while saving your default language')))
```
lebouquetin: 'Error while saving your default language' => 'Error while saving your language' (there is no "default language" for a user, there is a language)

lebouquetin: By the way, this new message does not seems to have associated translation in french.

Skylsmoi: fixed

PR #1623, opened 2019-04-23
add static tools config

buxx:
Note: This MR does not need green pipelines.


buxx:
> @buxx Is there any doc somewhere about using thoses tools ?

I open an issue about add documentation related to format code once we have formatted code.

buxx:
Issue here about doc: https://github.com/tracim/tracim/issues/1625

Issue #1226, opened 2018-12-05
Think about refactoring the way we deal with authorization exception when chained


inkhey:
**Is your feature request related to a problem? Please describe.**
as notified here : https://github.com/tracim/tracim/pull/1220#discussion_r238611737 ,
we raised only the first (for andAuthorizationChecker) or the last (for OrAuthorizationChecker) 
exception using auth mecanism. 

We should think about if we can deal better with those case where we have multiples errors for one check, but should return to frontend only one clear exception with error_code and without giving to much information about authorization mecanism.

Issue #1626, opened 2019-04-24
wrong file-path make in development.ini for reply by email when starting new docker images


PhilippeAccorsi:
## Bug description & symptom

Reply by email not working and we can see error in mail_fetcher.log `Socket fail with IMAP connection [Errno 21] Is a directory: '/var/tracim/data'`

## Diagnostic

In development.ini.sample this parameter `email.reply.lockfile_path` says that is path of file `email.reply.lockfile_path = %(here)s/email_fetcher.lock`.
When starting docker image of tracim, we change this path but we loose file name, so reply by email not working. line with error: ` sed -i "s|email.reply.lockfile_path = .*|email.reply.lockfile_path = /var/tracim/data|g" /etc/tracim/development.ini`

## Expected

- [ ] modify common.sh to add filename at end of path

## Prerequisites

activate reply by email for testing.

## How to reproduce

- Start image docker
- try to reply by email

## Impacted components

- *backend*

PhilippeAccorsi:
Verified.

PR #1628, opened 2019-04-24
[#1226] fix wrong parameter for reply by email [ci skip]

PhilippeAccorsi:
- [x] fix wrong parameter in common.sh for change in development.ini when you start new container whis new volume. 

If you have used volume before, development.ini file is not changed (you need to change this parameter manually).


PhilippeAccorsi:
related to #1626

Issue #1630, opened 2019-04-24
need some change for logo by default


PhilippeAccorsi:
## Bug description & symptom

Default logo is not the good one.

## Diagnostic

 It's the logo for progressive web app, and we need another logo for tracim.

## Expected

- [x] add new default logo

## Impacted components

- *backend*
- *frontend*

PhilippeAccorsi:
Issue edited: its not necessary to change something about manifest. Manifest use specific logo.
We just need to change default logo for tracim  `logo-tracim.png`.

PhilippeAccorsi:
Verified.

PR #1633, opened 2019-04-25
[#1630] change default logo of tracim

PhilippeAccorsi:
Here we change the default logo-tracim.png
This new file use new logo and transparency background.

No change for manifest is needed. the Manifest use specific files.


PhilippeAccorsi:
Related to #1630

Issue #1335, opened 2019-01-17
Show webdav in dashboard when activated


inkhey:
Webdav feature ui has been desactivated since : https://github.com/tracim/tracim/issues/886
We should reenable it but let is appear only if a config parameter (`webdav.ui.activated` ?) is true.
Frontend will be able to find information in endpoint `/api/v2/system/config`

PhilippeAccorsi:
@inkhey @Skylsmoi this issue is to introduce link for webdav in dashbord (like we have in tracim_v1).

PhilippeAccorsi:
For this feature, after discuss with @inkhey and @lebouquetin we add one parameter in development.ini with value `True` by default. This parameter is used to inform frontend about webdav url.

If parameter is `True` => frontend show webdav url in dashbord of shared space
If parameter is `False` => frontend not show webdav url  in dashbord of shared space

inkhey:
@Skylsmoi @lebouquetin I need to transmit at least both webdav url and webdav activation boolean to frontend, i tried something in #1643, we need to validate parameter name (and config parameter name) an the way we transmit it (i prefer transferring only one complete url with protocol and let frontend convert to gnome protocol if needed ( https become davs and http become dav)).

inkhey:
current parameter are: 
- webdav_enabled as boolean False
- webdav_url as string like : https://localhost:3030/webdav
both returned by /api/v2/system/config

lebouquetin:
it's ok for me. I agree with one url only and the frontend build `davs://xxx` from `https://xxx`

Skylsmoi:
@PhilippeAccorsi Testing information:
I have removed the global css rule on link (tag <a>). This rule caused me too much trouble to overrides.
As a result, links all over Tracim might have a design regression.
I have check all Tracim and have already fixed most of them.
BUT, if you ever ends up, finding one or two, notify me and I'll do the necessary.
Theses missing links should looks like a default blue text instead of their usual color.

PhilippeAccorsi:
- [ ] Missing translate about webdav
- [ ] Missing translate about caldav
- [ ] Icon about url for webdav is not a good one (its the same like we have for calendar)
    - need to find/discuss about the good icon for this link
- [ ] text about agenda has change, but not in `My account > agenda`

gbonaspetti:
I solve the translation problems, including the fourth item, but it's still missing to solve the icon issue.

gbonaspetti:
Some ideas to the icon issue:
[...]

PR #1643, opened 2019-04-26
Fix/1335 wedav enabled in config endpoint

inkhey:
related to #1335


frontend/src/container/Dashboard.jsx
```diff
@@ -518,13 +519,23 @@ class Dashboard extends React.Component {
...
+                  customClass='dashboard__section'
+                  introText={props.t("Use this link to integrate this shared space's agenda to your")}
+                  caldavText={props.t('CalDAV compatible software')}
                   agendaUrl={props.curWs.agendaUrl}
                 />
-              }
+              )}
+
+              {props.system.config.webdav_enabled && (
+                <WebdavInfo
```
lebouquetin: @Skylsmoi this is typically what I would expect to be generic. We should not have a AgendaInfo widget and a WebdavInfo widget but one DashboardExternalLinkInfo instantiated with specific parameters.

Issue #1624, opened 2019-04-23
app file is not in full height after decrease size of header


PhilippeAccorsi:
## Bug description & symptom

After reduce size of header, the file app not use the maximum height available.

## Diagnostic

"unknown"

## Expected

- [ ] The appa file use the maximum height is possible.

## Context and version information
Visible on firefox.

## How to reproduce

Open file app.

## Screenshots

![Sélection_003](https://user-images.githubusercontent.com/25848257/56588469-b169e700-65e3-11e9-833b-9a486b81b8c7.jpg)

## Impacted components

- *frontend*
- *css*

## Extra information

Space at end of thread app is too big also after decrease size of header, so I think its necessary to verify other app for bug like this.

Skylsmoi:
the height of app file is now correct

PhilippeAccorsi:
Verified.

PR #1647, opened 2019-04-26
[#1624] fix app file height

Skylsmoi:
https://github.com/tracim/tracim/issues/1624

### DIFF
--- a/frontend_app_file/src/css/index.styl
+++ b/frontend_app_file/src/css/index.styl
@@ -1,7 +1,7 @@
 @import "../../node_modules/tracim_frontend_lib/src/css/Variable.styl"
 
 .file
-  height calc(100% - 85px)
+  height calc(100% - 61px)
   &__header
     &__editionmode
       display none


frontend_app_file/src/css/index.styl
```diff
@@ -1,7 +1,7 @@
 @import "../../node_modules/tracim_frontend_lib/src/css/Variable.styl"
 
 .file
-  height calc(100% - 85px)
+  height calc(100% - 61px)
```
buxx: This cannot be a variable in stylus ?

Skylsmoi: well, it could but it would means a variable in the file Variable.styl of frontend_lib.
But it's the frontend's Header component design that will define the value of that variable to be then used by all apps.

That means the frontend's Header component have a design that ends up having a total height. Then this total height should be set in the frontend_lib variable then be used by components.

So in the end, we can't have everything that depends on one single variable. The Header component cannot depends on it since it's the one that defines it.

So I'm not sure what would be best

buxx: Please open an issue about possible refacto about applications size / container problematic. Then merge this MR.

Skylsmoi: see https://github.com/tracim/tracim/issues/1657

Issue #1641, opened 2019-04-26
add possibility to open content in new tab directly from recent activity on dashboard


PhilippeAccorsi:
## Feature description

Add possibility to CTRL+click on recent activity to open content in new tab.

## Context

At present you can open content in new tab with ctrl+click just in list of content. Not from dashboard recent activities.

## Impacted components

- *frontend*

Skylsmoi:
the link now handle the "open in new tab"
For folders:
- if user is at least content manager, he can click on it and it'll open the folder advanced app
- if user doesn't have enough rights, he can't click on it

PhilippeAccorsi:
Verified.

PR #1651, opened 2019-04-29
[#1641] added html link feature to recent activity in dashboard

Skylsmoi:
https://github.com/tracim/tracim/issues/1641

### DIFF
--- a/frontend/src/component/Dashboard/RecentActivity.jsx
+++ b/frontend/src/component/Dashboard/RecentActivity.jsx
@@ -1,7 +1,9 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import classnames from 'classnames'
+import { Link } from 'react-router-dom'
 import { Badge } from 'tracim_frontend_lib'
+import { PAGE } from '../../helper.js'
 
 require('./RecentActivity.styl')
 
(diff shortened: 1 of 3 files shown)


frontend/src/component/Dashboard/RecentActivity.jsx
```diff
@@ -25,24 +27,43 @@ export const RecentActivity = props =>
...
+          const recentActivityItemName = (
+            <div className='recentactivity__list__item__name'>
+              {content.label}
+              {content.type === 'file' && (
+                <Badge text={content.fileExtension} customClass='badgeBackgroundColor' />
+              )}
+            </div>
+          )
+
+          if (content.type === 'folder' && props.roleIdForLoggedUser < 4) return (
```
buxx: What is `4` ? It seems to be a role level. This must be a constant: I guess there is other occurrence in code, so I suggest opening an issue about refactor them into constants.

frontend/src/component/Dashboard/RecentActivity.jsx
```diff
@@ -25,24 +27,43 @@ export const RecentActivity = props =>
...
             <div
```
buxx: It is intentional to not have a Link in this case ?

buxx: I think i not understand correctly this part because we have no "folder" in "Recent activity", i'm wrong ?

frontend/src/container/Dashboard.jsx
```diff
@@ -481,10 +479,11 @@ class Dashboard extends React.Component {
               <div className='dashboard__workspaceInfo'>
                 <RecentActivity
                   customClass='dashboard__activity'
+                  workspaceId={props.curWs.id}
+                  roleIdForLoggedUser={idRoleUserWorkspace}
```
buxx: Does this line will print in html the current use role level ? If yes, why use html to store this information, current user informations are probably in javascript data no ?

frontend/src/component/Dashboard/RecentActivity.jsx
Skylsmoi: we do.
that's where the issue come from.
Folders should not be a clickable link for users that don't have enough rights to open the folder advanced app.

frontend/src/container/Dashboard.jsx
Skylsmoi: no it wont.
It's to allow the component RecentActivity to know whether it should create a link or not for the folder entries

frontend/src/component/Dashboard/RecentActivity.jsx
Skylsmoi: yes, the constant is in frontend, it should be moved to frontend_lib but it, indeed, requires some refactor
see https://github.com/tracim/tracim/issues/1656

buxx: Ok thank's

Issue #1652, opened 2019-04-29
Better copy: preservate revision_id order by creating subcontent and revision in correct order


inkhey:
## Bug description & symptom

Found during fixing #1499, before this fix, issue cannot occured as copy of content with comment is broken.
Currently copy is done by copying all revision and then copying all subcontent. This means that revision_id order beteween revisions of content and subcontent are not orderer the same way as before.
This broke "comment_ids" of revision in api which rely of revision_id order. Currently we can have same comment associated to multiple revision using current copy feature.

## Expected

- Do a correctly ordered copy.
- Verify comment_ids of revision work correctly.

PhilippeAccorsi:
Verified.

Issue #1499, opened 2019-03-23
Webdav: file copy return error but done properly on linux (Nautilus) when copy file from root to another root workspace


inkhey:
## Bug description & symptom

(we need to reproduce with log) 
Copy done with linux (Nautilus) in webdav has return error to ui but has been done properly. 
We should inspect this.

Some complementary information: 
- Has this issue happened on webdav copy with nautilus, webdav method used is probably COPY.
- Done with File.
- ~~Error can have occured on method used after copy not copy itself.~~

- Error happened from root to another root workspace.

## Screenshot

![Capture d’écran de 2019-03-23 10-29-45](https://user-images.githubusercontent.com/25848257/54864388-d3f3b080-4d56-11e9-8031-59811150e4f3.png)

buxx:
There is wedav log for "paste" action.
[webdav_error_1499.txt](https://github.com/tracim/tracim/files/2999188/webdav_error_1499.txt)

buxx:
There is log of working case (when copy from sub dir of workspace and copy in subfolder of other workspace)

[webdav_ok_1499.txt](https://github.com/tracim/tracim/files/2999192/webdav_ok_1499.txt)

inkhey:
```
❯ ack 500 webdav_error_1499.txt 
<139669296310144> --- COPY Response(500 Internal Server Error): ---
b"<!DOCTYPE HTML PUBLIC '-//W3C//DTD HTML 4.01//EN' 'http://www.w3.org/TR/html4/strict.dtd'>\n<html><head>\n  <meta http-equiv='Content-Type' content='text/html; charset=UTF-8'>\n  <title>500 Internal Server Error</title>\n</head><body>\n  <h1>500 Internal Server Error</h1>\n  <p>500 Internal Server Error: An internal server error occurred\n    Source exception: &#x27;&#x27;</p>\n<hr/>\n<a href='https://github.com/mar10/wsgidav/'>WsgiDAV/2.4.1</a> - 2019-03-23 09:27:54.575103\n</body></html>"
[pid: 508|app: 0|req: 2536/2536] 127.0.0.1 () {40 vars in 898 bytes} [Sat Mar 23 09:27:54 2019] COPY /webdav/Intranet/Bonnes%20pratiques%20d'utilisation%20des%20documents%20tracim.html => generated 484 bytes in 279 msecs (HTTP/1.1 500) 3 headers in 121 bytes (1 switches on core 0)
```

inkhey:
[...]

PR #1653, opened 2019-04-29
Fix/1499 fix copy and move of comments

inkhey:
related to #1499
will be continued by #1652 fixing copy revision order between subcontent and content revision.

### DIFF
--- a/backend/tracim_backend/lib/core/content.py
+++ b/backend/tracim_backend/lib/core/content.py
@@ -545,7 +545,7 @@ class ContentApi(object):
             # set label and file_extension
             content.file_name = filename
         else:
-            if content_type_slug == content_type_list.Comment.slug:
+            if self._allow_empty_label(content_type_slug):
                 # INFO - G.M - 2018-07-16 - Default label for comments is
                 # empty string.
                 content.label = ""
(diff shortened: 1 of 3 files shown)


backend/tracim_backend/lib/core/content.py
```diff
@@ -2011,3 +2098,11 @@ def generate_folder_label(self, workspace: Workspace, parent: Content = None) ->
             query = query.filter(Content.parent == parent)
 
         return _("New folder {0}").format(query.count() + 1)
+
+    def _allow_empty_label(self, content_type_slug):
```
lebouquetin: typing is missing

backend/tracim_backend/lib/core/content.py
```diff
@@ -1450,6 +1462,36 @@ def copy(
         new_file_extension: str = None,
         do_save: bool = True,
         do_notify: bool = True,
+    ) -> Content:
+        """
+        Copy nearly all content, revision included. Children not included, see
```
lebouquetin: The docstring is wrong. You copy all contents, children included now.

backend/tracim_backend/lib/webdav/resources.py
```diff
@@ -892,14 +892,13 @@ def copyMoveSingle(self, destpath, isMove):
         except ContentNotFound:
             destination_parent = None
         try:
-            new_content = self.content_api.copy(
+            self.content_api.copy(
```
lebouquetin: you changed the behavior without to change the name. It's dangerous. The new copy is always recursive? You need to make this information explicit somewhere (probably in the `copy()` method docstring)

backend/tracim_backend/lib/core/content.py
inkhey: done.

backend/tracim_backend/lib/core/content.py
inkhey: done.

backend/tracim_backend/lib/webdav/resources.py
inkhey: done. It is always recursive now.

backend/tracim_backend/lib/core/content.py
```diff
@@ -2099,7 +2099,7 @@ def generate_folder_label(self, workspace: Workspace, parent: Content = None) ->
 
         return _("New folder {0}").format(query.count() + 1)
 
-    def _allow_empty_label(self, content_type_slug):
+    def _allow_empty_label(self, content_type_slug: str):
```
lebouquetin: what is the return of the function? add full typing

inkhey: done.

PR #1655, opened 2019-04-30
Fix/1652  copy preserving id order

inkhey:
related to #1652.
:warning: need  #1653

### DIFF
--- a/backend/tracim_backend/lib/core/content.py
+++ b/backend/tracim_backend/lib/core/content.py
@@ -1551,18 +1551,32 @@ class ContentApi(object):
                 "and a valid filename".format(item.content_id, content_type_slug)
             )
 
-        content = item.copy(parent)
-        # INFO - GM - 15-03-2018 - add "copy" revision
+        content = self._copy(item, parent, label, workspace, file_extension, do_save, do_notify)
+
+        return content
+
+    def _copy(
+        self,
+        item: Content,
+        new_parent: Content = None,
+        new_label: str = None,
+        new_workspace: Workspace = None,
+        new_file_extension: str = None,
+        do_save: bool = True,
+        do_notify: bool = True,
+    ) -> Content:
+        # INFO - GM - 15-03-2018 - add "copy" revision for all content
+        content, children_new_content, children_original_content = item.copy(new_parent)
         with new_revision(
             session=self._session,
             tm=transaction.manager,
             content=content,
             force_create_new_revision=True,
         ) as rev:
-            rev.parent = parent
-            rev.workspace = workspace
-            rev.label = label
-            rev.file_extension = file_extension
+            rev.parent = new_parent
+            rev.workspace = new_workspace
+            rev.label = new_label
+            rev.file_extension = new_file_extension
             rev.revision_type = ActionDescription.COPY
             rev.properties["origin"] = {
                 "content": item.id,
@@ -1570,6 +1584,23 @@ class ContentApi(object):
             }
         if do_save:
             self.save(content, ActionDescription.COPY, do_notify=do_notify)
+
+        for original_content_id, new_child in children_new_content.items():
+            original_child = children_original_content[original_content_id]
+            with new_revision(
+                session=self._session,
+                tm=transaction.manager,
+                content=new_child,
+                force_create_new_revision=True,
+            ) as rev:
+                rev.workspace = new_workspace
+                rev.revision_type = ActionDescription.COPY
+                rev.properties["origin"] = {
+                    "content": original_child.id,
+                    "revision": original_child.last_revision.revision_id,
+                }
+            self.save(new_child, ActionDescription.COPY, do_notify=False)
+
         return content
 
     def _move_children_content_to_new_workspace(self, item: Content, new_workspace: Workspace):
--- a/backend/tracim_backend/lib/core/content.py
+++ b/backend/tracim_backend/lib/core/content.py
@@ -1551,40 +1551,89 @@ class ContentApi(object):
                 "and a valid filename".format(item.content_id, content_type_slug)
             )
 
-        content = self._copy(item, parent, label, workspace, file_extension, do_save, do_notify)
-
+        content, children_new_content, children_original_content = self._copy(item, parent)
+        content, _, _ = self._add_copy_revision(
+            item,
+            content,
+            children_original_content,
+            children_new_content,
+            parent,
+            label,
+            workspace,
+            file_extension,
+            do_save,
+            do_notify,
+        )
         return content
 
     def _copy(
+        self, content: Content, new_parent: Content = None
+    ) -> typing.Tuple[Content, typing.Dict[int, Content], typing.Dict[int, Content]]:
+        """
+        Create new content for content and his children, recreate all revision in order and
+        return all these new content
+        :param content: root content of copy
+        :param new_parent: new parent of root content
+        :return: new content created based on root content,
+        dict of new children content and original children content with original content id as key.
+        """
+        root_new_content = Content()
+        # INFO - G.M - 2019-04-30 - we store all children content created and old content id of them
+        # to be able to retrieve them for applying new revisions on them easily. key of dict is
+        # original content_id.
+        children_new_content = {}  # type: typing.Dict[int,Content]
+        # INFO - G.M - 2019-04-30 - we store alse old content of children to allow applying new
+        # revision related to old data. key of dict is original content id.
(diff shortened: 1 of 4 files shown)


lebouquetin:
@inkhey please fix the other MR? Then merge it in the current branch in order to have the real diff in the current MR

backend/tracim_backend/lib/core/content.py
```diff
@@ -1551,30 +1551,126 @@ def _copy_current(
...
+        new_parent: Content = None,
+        new_label: str = None,
+        new_workspace: Workspace = None,
+        new_file_extension: str = None,
+        do_save: bool = True,
+        do_notify: bool = True,
+    ) -> typing.Tuple[Content, typing.Dict[int, Content], typing.Dict[int, Content]]:
+        """
+        Add copy revision for all new content
+        :param original_content: original content of root file in copy
```
lebouquetin: why "file" ? its' content, doesn't it? same remark for all parameter documented in the  docstring

backend/tracim_backend/lib/core/content.py
```diff
@@ -1551,30 +1551,126 @@ def _copy_current(
...
+                else:
+                    related_parent = children_new_content[rev.parent_id]
+            # INFO - G.M - 2019-04-30 - copy of revision itself.
+            cpy_rev = ContentRevisionRO.copy(rev, related_parent)
+            related_content.revisions.append(cpy_rev)
+            self._session.add(related_content)
+            self._session.flush()
+        return root_new_content, children_new_content, children_original_content
+
+    def _add_copy_revision(
```
lebouquetin: `_add_copy_revision` => `_add_copy_revisions`

backend/tracim_backend/lib/core/content.py
```diff
@@ -1551,30 +1551,126 @@ def _copy_current(
...
+        self,
+        original_content: Content,
+        new_content: Content,
+        children_original_content,
```
lebouquetin: `children_original_content` => `original_content_children`
`children_new_content` => `new_content_children`

backend/tracim_backend/lib/core/content.py
```diff
@@ -1551,30 +1551,126 @@ def _copy_current(
...
+        children_new_content,
+        new_parent: Content = None,
+        new_label: str = None,
+        new_workspace: Workspace = None,
+        new_file_extension: str = None,
+        do_save: bool = True,
+        do_notify: bool = True,
+    ) -> typing.Tuple[Content, typing.Dict[int, Content], typing.Dict[int, Content]]:
```
lebouquetin: Do not use complex tuples. Define a container class and return it, somethink like

```
class AddCopyRevisionsResult(object):
    def __init__(xxx):
        self.new_content = new_content
        self.new_children_dict = new_children_dict  # dict key is original content id
        self.original children_dict = original_children_dict # dict key is original content id
```

Note: do not use terms like root content which has no meaning here, I think. *Root* is *Racine*, I think what you mean is "original" (or original root).

backend/tracim_backend/models/data.py
```diff
@@ -1173,6 +1173,18 @@ def children(self) -> ["Content"]:
 
         return valid_children
 
+    @property
+    def recursive_children(self) -> ["Content"]:
```
lebouquetin: use a method here, not a property, I think. `get_children_recursively()`

backend/tracim_backend/models/data.py
```diff
@@ -1173,6 +1173,18 @@ def children(self) -> ["Content"]:
...
+        """
+        Get all children recursively (including children of children...)
+        """
+        children = []
+        for child in self.children:
+            children.append(child)
+            for sub_child in child.recursive_children:
```
lebouquetin: Instead of:
```
            for sub_child in child.recursive_children:
                children.append(sub_child)
```

Write it like this:

```
            children.extend(child.get_recursive_children())
```

By the way, it would be more elegant to have a method `get_children(recursivly=True)`

backend/tracim_backend/models/data.py
```diff
@@ -1422,12 +1434,17 @@ def format_path(cls, url_template: str, content: "Content") -> str:
...
         return url_template.format(wid=wid, fid=fid, ctype=ctype, cid=cid)
 
-    def copy(self, parent):
-        cpy_content = Content()
-        for rev in self.revisions:
-            cpy_rev = ContentRevisionRO.copy(rev, parent)
-            cpy_content.revisions.append(cpy_rev)
-        return cpy_content
+    @property
+    def recursive_tree_revision(self):
```
lebouquetin: do not use a property, it's a real "function" here.

`recursive_tree_revision` => `get_tree_revisions`

backend/tracim_backend/models/data.py
```diff
@@ -1422,12 +1434,17 @@ def format_path(cls, url_template: str, content: "Content") -> str:
...
+        """Get all revision sorted by id of content and all his children recursively"""
+        revisions = []  # type: typing.List[ContentRevisionRO]
+        for revision in self.revisions:
+            revisions.append(revision)
+        for children in self.recursive_children:
+            for revision in children.revisions:
```
lebouquetin: replace:

````
    for children in self.recursive_children:
        for revision in children.revisions:
            revisions.append(revision)
````
by:

````
    for chil in self.recursive_children:
        revisions.extend(child.get_tree_revisions())
````

note : children is plura form of child. A loop must be `for child in self.children`

Comment (inkhey) [reply backend/tracim_backend/models/data.py]:
```
            for child in self.get_children(recursively=True):
                revisions.extend(child.revisions)
```
is more correct. we already get all children recursively, it doesn't make sens to get for each of them recursive tree revisions.

Comment (lebouquetin) [reply backend/tracim_backend/models/data.py]: ok

File: backend/tracim_backend/models/data.py (hunk 5 above)
Comment (inkhey) [reply backend/tracim_backend/models/data.py]: done

File: backend/tracim_backend/models/data.py (hunk 7 above)
Comment (inkhey) [reply backend/tracim_backend/models/data.py]: done

File: backend/tracim_backend/models/data.py (hunk 8 above)
Comment (inkhey) [reply backend/tracim_backend/models/data.py]: done.

File: backend/tracim_backend/lib/core/content.py (hunk 3 above)

File: backend/tracim_backend/lib/core/content.py (hunk 2 above)

File: backend/tracim_backend/models/data.py (hunk 6 above)
Comment (inkhey) [reply backend/tracim_backend/models/data.py]: done.

File: backend/tracim_backend/lib/core/content.py (hunk 1 above)
Comment (inkhey) [reply backend/tracim_backend/lib/core/content.py]: fixed.

File: backend/tracim_backend/models/data.py (hunk 7 above)
Comment (lebouquetin) [reply backend/tracim_backend/models/data.py]: @inkhey this is a plural form, so revision**s**. By the way, you are missing the typing.

File: backend/tracim_backend/models/data.py [hunk 9]
```diff
@@ -1422,12 +1433,15 @@ def format_path(cls, url_template: str, content: "Content") -> str:
         cid = content.content_id
         return url_template.format(wid=wid, fid=fid, ctype=ctype, cid=cid)
 
-    def copy(self, parent):
-        cpy_content = Content()
-        for rev in self.revisions:
-            cpy_rev = ContentRevisionRO.copy(rev, parent)
-            cpy_content.revisions.append(cpy_rev)
-        return cpy_content
+    def get_tree_revisions(self):
```
lebouquetin: missing typing.

inkhey: done.

backend/tracim_backend/models/data.py
```diff
@@ -1266,18 +1267,18 @@ def properties(self, properties_struct: dict) -> None:
...
         return format_timedelta(delta_from_datetime - self.created, locale=get_locale())
 
-    def datetime_as_delta(self, datetime_object, delta_from_datetime: datetime = None):
+    def datetime_as_delta(self, datetime_object, delta_from_datetime: datetime = None) -> timedelta:
         if not delta_from_datetime:
             delta_from_datetime = datetime.utcnow()
         return format_timedelta(delta_from_datetime - datetime_object, locale=get_locale())
 
-    def get_child_nb(self, content_type: str, content_status=""):
+    def get_child_nb(self, content_type: str, content_status="") -> int:
```
lebouquetin: missing typing sur content_status

inkhey: done.

Issue #1658, opened 2019-05-02
change default logo of tracim


PhilippeAccorsi:
## Bug description & symptom

Tracim use old logo 
![logo-tracim-old](https://user-images.githubusercontent.com/25848257/57064964-3cc03800-6cc8-11e9-83cf-b515d1f787a7.png)

## Expected

- [ ] use new one 
![logo-tracim](https://user-images.githubusercontent.com/25848257/57064974-43e74600-6cc8-11e9-88be-19362ae3a568.png)

PhilippeAccorsi:
Verified.

PR #1659, opened 2019-05-02
[#1658] change default logo of tracim

PhilippeAccorsi:
- [x] change default logo of tracim

PR #1660, opened 2019-05-02
[#1658] remove logo image from dist

PhilippeAccorsi:
#1658
- [x] remove logo image from `dist/`


PhilippeAccorsi:
related to #1659

Issue #1648, opened 2019-04-26
Remove automatic scroll to the top when opening content


PhilippeAccorsi:
## Bug description & symptom

When I open a content, in background I can see automatic scroll to the top.

## Diagnostic

"unknown"

## Expected

- [ ] When I open a content, no scroll is activated (look on demo.tracim.fr for normal case)

## Prerequisites

- Have more than 14 content in list of all content to see scroll bar on the right of the page

## How to reproduce

- go down with scroll bar
- open a content
=> you can see scroll to the top when your app is open
- if you need to open a content near the first one, you need to scroll down first.

## Screenshots

![automatic scroll to top](https://user-images.githubusercontent.com/25848257/56816415-cba20000-6843-11e9-9d20-c8ae5e16af4a.gif)

## Impacted components

"unknown"*

PhilippeAccorsi:
@lebouquetin I think its important to solved that for 2.2
In demo.tracim.fr (2.1) no automatic scroll when opneing content.

Skylsmoi:
the issue come from this commit https://github.com/tracim/tracim/commit/78b3704cd79a414fd2e929e13c18201f60bdb17c
which is fixing this issue https://github.com/tracim/tracim/issues/1563

Skylsmoi:
The problem is the routing of workspace content.
It needs 2 routes:
- list of content like `/ui/workspaces/1/contents`
- open a content like `/ui/workspaces/1/contents/html-document/4`

Having the 2 routes declared separately create the current issue.
Having them merged like in 78b3704 fixes the current issue but will recreate #1563

Skylsmoi:
apps should now properly reload the data from api when the content id change

Skylsmoi:
I rolled back the routing with the switch and put back the double matching route on content list and content open.
But I have fixed the usage of data received by _app_name__showApp custom event.
The handlers of "_showApp" events will now call a "_reloadContent" when the content id has changed

PhilippeAccorsi:
Verified.

PR #1662, opened 2019-05-02
[#1648] fixed automatic scroll to top on content open

Skylsmoi:
https://github.com/tracim/tracim/issues/1648


lebouquetin:
@Skylsmoi 

You have the following code in 3 jsx files:

```
      case 'html-document_showApp':
        console.log('%c<HtmlDocument> Custom event', 'color: #28a745', type, data)
        if (data.content.content_id !== state.content.content_id) {
          const event = new CustomEvent('appCustomEvent', {detail: {type: 'html-document_reloadContent',data: data.content}})
          document.dispatchEvent(event)
          return
        }
```

```
      case 'thread_showApp':
        console.log('%c<Thread> Custom event', 'color: #28a745', type, data)
        if (data.content.content_id !== state.content.content_id) {
          const event = new CustomEvent('appCustomEvent', {detail: {type: 'thread_reloadContent',data: data.content}})
          document.dispatchEvent(event)
          return
        }
```

```
      case 'file_showApp':
        console.log('%c<File> Custom event', 'color: #28a745', type, data)
        if (data.content.content_id !== state.content.content_id) {
          const event = new CustomEvent('appCustomEvent', {detail: {type: 'file_reloadContent',data: data.content}})
          document.dispatchEvent(event)
          return
        }

This is exactly the same code exect the text in the log and the type of custom event. This must be factorized, otherwise you will duplicate code for each new content type for doing exactly the same thing.

Skylsmoi:
@lebouquetin I know, this is part of the refactor that needs to be done on apps feature.
They share a lot a same code.

Issue #1661, opened 2019-05-02
Better support for env var settable config


inkhey:
## Bug description & symptom

Mechanism allowing to set environnement variable to override config file work great but some config parameters are not setted through into CFG object.
Those included some  important  "global" settings:
- sqlalchemy.url :warning: 
- beaker cookie dirs (session.data_dir, session.lock_dir)

But also some more specific settings:

- radicale special config
- other beaker session parameters
- other sqlalchemy parameters

## Diagnostic

We should at least allowed relying of CFG object parameter for sqlalchemy.url and beaker cookies dirs in order to support most case and be able to force beaker dir existence ( see https://github.com/tracim/tracim/issues/1524)

## Expected

- Fully Support only needed parameters, other parameter support should work only in config file.

inkhey:
i proposed to support now sqlalchemy.url and beaker cookies dir to allow fixing #1524 as soon as possible and think about a more complex support later.

inkhey:
done in  #1672: 
- use of env_var in travis conf file
- remove most generic config param in test_config.ini file, using dotenv mecanism instead in order to allow setting more easily default database and dir used for tests.
- rename TRACIM_BACKEND_TEST_CONFIG_FILE to BACKEND_TEST_CONFIG_FILE to avoid confusion between tracim var and non-tracim var.
- support for env var for 'session.data_dir' and 'session.lock_dir'.
- support for env var for 'sqlalchemy.url' for both tracim and alembic.
- move sliced_dict method to utils.
- fix issue on with 'type_' on migration script with mysql. (don't really know the reason of origin of the issue as tests work before, dependencies update which means more strict check ?)

inkhey:
@PhilippeAccorsi , @lebouquetin we merge the associated branch but we do not set the ticket in done, i move in to 2.3, "need test".

inkhey:
@PhilippeAccorsi to check if this work, you can just test if you can set beaker session path, sqlalchemy_url, radicale path, etc... 
Another way to check this, is just waiting https://github.com/tracim/tracim/pull/1687 which will use fully the feature for travis. If tests doesn't fail, this mean feature is working .

Skylsmoi:
As inkhey said, since #1687 is merged and tests have passed, I consider this issue verified

PR #1672, opened 2019-05-03
Fix/1661  better env var support

inkhey:
related to #1661 
description here: https://github.com/tracim/tracim/issues/1661#issuecomment-489103638_


backend/tracim_backend/tests/__init__.py
```diff
@@ -250,8 +259,14 @@ def override_settings(
         settings["sqlalchemy.url"] = "sqlite://"
         return settings
 
-    def init_database(self, settings: typing.Dict[str, typing.Any]) -> None:
-        self.engine = get_engine(settings)
+    def override_app_config(self) -> None:
```
lebouquetin: what is the usage of this method? need more docstring

backend/tracim_backend/tests/__init__.py
```diff
@@ -120,15 +121,16 @@ def setUp(self) -> None:
         # TODO - G.M - 2019-03-19 - Replace this code by something better, see
         # https://github.com/algoo/hapic/issues/144
         hapic._controllers = []
-        self.connect_database(create_tables=True)
         self.app_config = CFG(self.settings)  # type: CFG
+        self.override_app_config()
```
lebouquetin: I think this is not the way to do this properly, no?

backend/tracim_backend/models/setup_models.py
```diff
@@ -1,26 +1,42 @@
...
+if typing.TYPE_CHECKING:
+    from tracim_backend.config import CFG
 # run configure_mappers after defining all of the models to ensure
 # all relationships can be setup
 configure_mappers()
 
 
-def get_engine(settings, prefix="sqlalchemy."):
-    return engine_from_config(settings, prefix)
+def get_engine(app_config: "CFG", prefix="sqlalchemy."):
```
lebouquetin: typing

backend/tracim_backend/models/setup_models.py
```diff
@@ -1,26 +1,42 @@
...
+from tracim_backend.lib.utils.utils import sliced_dict
 from tracim_backend.models.auth import Group  # noqa: F401
 from tracim_backend.models.auth import Permission  # noqa: F401
 from tracim_backend.models.auth import User  # noqa: F401
 from tracim_backend.models.data import Content  # noqa: F401
 from tracim_backend.models.data import ContentRevisionRO  # noqa: F401
 from tracim_backend.models.meta import DeclarativeBase  # noqa: F401
 
+if typing.TYPE_CHECKING:
+    from tracim_backend.config import CFG
```
lebouquetin: why is the import here? if there is a reason, then a comment is required. If no reason, then move it to top of the file

backend/tracim_backend/migration/versions/32e629b17e2e_change_calendar_enabled_to_agenda_enabled_colum.py
```diff
@@ -5,20 +5,21 @@
...
 
-# revision identifiers, used by Alembic.
 revision = "32e629b17e2e"
 down_revision = "f889c2b59759"
 
 
 def upgrade():
     with op.batch_alter_table("workspaces") as bop:
-        bop.alter_column("calendar_enabled", new_column_name="agenda_enabled")
+        bop.alter_column("calendar_enabled", new_column_name="agenda_enabled", type_=Boolean)
```
lebouquetin: migrations are not supposed to change. What is the reason for this change?

backend/tracim_backend/migration/versions/32e629b17e2e_change_calendar_enabled_to_agenda_enabled_colum.py
```diff
@@ -5,20 +5,21 @@
...
     # ### end Alembic commands ###
 
 
 def downgrade():
     # ### commands auto generated by Alembic - please adjust! ###
     with op.batch_alter_table("workspaces") as bop:
-        bop.alter_column("agenda_enabled", new_column_name="calendar_enabled")
+        bop.alter_column("agenda_enabled", new_column_name="calendar_enabled", type_=Boolean)
```
lebouquetin: same remark/question

backend/tracim_backend/migration/env.py
```diff
@@ -14,6 +16,11 @@
 # this is the Alembic Config object, which provides
 # access to the values within the .ini file in use.
 config = context.config
+# INFO - G.M - 2019-05-02 - hack to get env url instead of current config file
+# url for migration script
+env_sqlalchemy_url = os.environ.get("TRACIM_SQLALCHEMY__URL")
+if env_sqlalchemy_url:
+    config.set_main_option("sqlalchemy.url", env_sqlalchemy_url)
```
lebouquetin: Instead of 

```
env_sqlalchemy_url = os.environ.get("TRACIM_SQLALCHEMY__URL")
if env_sqlalchemy_url:
    config.set_main_option("sqlalchemy.url", env_sqlalchemy_url)
```

we should write this something like :

```
config.set_main_option("sqlalchemy.url", env_sqlalchemy_url, config.get_main_option("sqlalchemy.url")
```

no ?

backend/tracim_backend/lib/utils/utils.py
```diff
@@ -299,3 +299,17 @@ def normpath(path):
     elif path == "":
         path = "/"
     return base_normpath(path)
+
+
+def sliced_dict(data: typing.Dict[str, any], beginning_key_string: str):
```
lebouquetin: typing

backend/tracim_backend/migration/versions/32e629b17e2e_change_calendar_enabled_to_agenda_enabled_colum.py
inkhey: i agree. This was needed to fix issue during test of migration for mysql. (this migration is the only one to fail). I do not really know why this parameter "type_" is now needed whereas it was not at all needed before. I can inspect deeper if you think that useful. Anyway this fix, solve the issue.

lebouquetin: what are we supposed to do on our production? migrate down and up again? Manually patch the database?

backend/tracim_backend/migration/env.py
inkhey: this syntax does not exist for config alembic file.

lebouquetin: ok.

backend/tracim_backend/migration/versions/32e629b17e2e_change_calendar_enabled_to_agenda_enabled_colum.py
lebouquetin: My feeling is downgrade/upgrade is ok, but I'd like to be sure.

backend/tracim_backend/tests/__init__.py
inkhey: you're right, i just refactored it.

backend/tracim_backend/tests/__init__.py
inkhey: done.

backend/tracim_backend/models/setup_models.py
inkhey: done.

backend/tracim_backend/migration/versions/32e629b17e2e_change_calendar_enabled_to_agenda_enabled_colum.py
inkhey: > what are we supposed to do on our production? migrate down and up again? Manually patch the database?
Nothing. old behaviour does exactly the same. It's just that for unknown reason it's now necessary to specify type_ for mysql. If you want to be sure, i can verify column types from old and new code to exactly match, but i do no see any reason that this method convert column type to another one when we don't specify it (+ if it does something like that, it should probably make many other test failed)

lebouquetin: ok, thanks for the explanation

backend/tracim_backend/models/setup_models.py
inkhey: explained: 'avoid circular import for type-checking usage of CFG' , done.

backend/tracim_backend/lib/utils/utils.py
inkhey: done

backend/tracim_backend/models/data.py
```diff
@@ -1163,6 +1163,7 @@ def children(self) -> ["Content"]:
             .query(func.max(ContentRevisionRO.revision_id))
             .filter(ContentRevisionRO.content_id.in_(children_content_ids))
             .group_by(ContentRevisionRO.content_id)
+            .order_by(ContentRevisionRO.content_id)
```
lebouquetin: it should be content_revision_id, doesn't it ?

backend/tracim_backend/models/data.py
```diff
@@ -1183,6 +1184,7 @@ def get_children(self, recursively: bool = False) -> ["Content"]:
             children.append(child)
             if recursively:
                 children.extend(child.get_children(recursively=recursively))
+        children = sorted(children, key=lambda child: child.content_id)
```
lebouquetin: same thing here: we should sort by content_id, and by content_revision_id if content_id is the same

backend/tracim_backend/models/data.py
inkhey: not here. we sort content not revision here.

backend/tracim_backend/models/data.py
inkhey: same answers : we sort content.

lebouquetin:
@inkhey this MR is still WIP? If not, please change the name so that we merge it.

inkhey:
@PhilippeAccorsi , @lebouquetin we merge the associated branch but we do not set the ticket in done, i move in to 2.3, "need test".

Issue #1083, opened 2018-10-25
Fix coverall urls


inkhey:
**Describe the bug**
coverall url like of file, this in this page are all broken:  https://coveralls.io/builds/19719155 .
Check if something is needed in order to fix this.

inkhey:
This issues is annoying as it's complicated to see coverage diff without this. :disappointed: 

This issue seems a bit complicated to handle because most backend test need actually to be run from backend dir (to access to config file). so we need to do test in a folder and then finding a way to 'repair' path for coveralls that image path are from root directory of 
i just try something (and try also named jobs to distinct coverage for unit test and coverage for all) but without success: 
 https://github.com/tracim/tracim/tree/fix/1083_fix_coveralls_config

inkhey:
already done with https://github.com/tracim/tracim/pull/1687

PR #1687, opened 2019-05-07
Fix/1524  make config paths mandatory

inkhey:
need #1661 , this fix also #1083 
related to #1524 
more info about content: https://github.com/tracim/tracim/issues/1524#issuecomment-490919128
Todo: 
- [X]  Fixed ! ~~coverall seems completly broken now (It was already partially broken #1083 before)~~

:warning:  Redo all tests and notify frontend for travis tests issue.
:warning: Discuss about https://github.com/tracim/tracim/issues/1524#issuecomment-490923517 and https://github.com/tracim/tracim/issues/1524#issuecomment-490930751


inkhey:
config/doc part is approved by @PhilippeAccorsi but code itself need review.

backend/tracim_backend/config.py
```diff
@@ -788,6 +819,48 @@ class CST(object):
         ASYNC = "ASYNC"
         SYNC = "SYNC"
 
+    def check_mandatory_param(self, param_name: str, value: typing.Any, when_str: str = "") -> None:
+        if not value:
```
buxx: docstring

backend/tracim_backend/config.py
```diff
@@ -788,6 +819,48 @@ class CST(object):
...
+            raise ConfigurationError(
+                'ERROR: "{}" configuration is mandatory {when_str}.'
+                "Set it before continuing.".format(param_name, when_str="")
+            )
+
+    def check_directory_path_param(
+        self, param_name: str, path: str, writable: bool = False, readable: bool = True
+    ) -> None:
```
buxx: docstring

backend/tracim_backend/command/database.py
```diff
@@ -119,7 +119,11 @@ def take_action(self, parsed_args: argparse.Namespace) -> None:
                 depot = DepotManager.get()
                 depot_files = depot.list()
                 for file_ in depot_files:
-                    depot.delete(file_)
+                    try:
+                        depot.delete(file_)
+                    # TODO - G.M - 2019-05-09 - better handling of exception here
+                    except Exception:
+                        print("Something goes wrong during deletion of {}".format(file_))
```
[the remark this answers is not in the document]
inkhey: done

backend/tracim_backend/config.py
inkhey: done

backend/tracim_backend/config.py
inkhey: done

inkhey:
> @inkhey you delete the database before to drop files. My feeling is, you should do it in reverse order, no? (first drop files - which may fail, then if everything is ok drop the database. Otherwise, You have no reverence to files anymore for deleting content).

Do the other way is also not really nice, it means you have content without associated file which mean content "broken" in tracim.

lebouquetin:
I agree that it's a completely different topic (so you can create a specific issue)
The way it is currently done, we keep files but we have no way to know what they are / what they contain (while having a dtabase without files content is also broken but understandable)

The right way to do things would probably to generate a file listing file which have not been deleted. (to discuss when fixing the specific issue you'll add)

inkhey:
> I agree that it's a completely different topic (so you can create a specific issue)

https://github.com/tracim/tracim/issues/1863

Issue #1712, opened 2019-05-14
Some buttons is now blue, ex in app file (regression)


PhilippeAccorsi:
## Bug description & symptom

Some buttons are now in blue color. In tracim 2.1 the color is black. Ex: download button in app file.

## Diagnostic

"unknown"

## Expected

- [ ] find all regression about color button
- [ ] Color of all button is black now

## How to reproduce

- open app file on tracim rc1

## Screenshots

Old color (visible on https://demo.tracim.fr)
![old color](https://user-images.githubusercontent.com/25848257/57696434-0b912180-7651-11e9-981e-b4ad0c9de9db.jpg)

Current color (with tracim rc1)
![current color](https://user-images.githubusercontent.com/25848257/57696477-22377880-7651-11e9-9bfd-3d31cb839ba8.jpg)

## Impacted components

- *frontend*
- *htmldoc / file / thread app*
- maybe more ....

gbonaspetti:
The default color for <a> tags it's blue according Bootstrap. I change the default for black in the Generic file in the frontend.

PhilippeAccorsi:
Verified.

PR #1714, opened 2019-05-14
[#1712] change button colors to black

gbonaspetti:
https://github.com/tracim/tracim/issues/1712


frontend/src/css/Generic.styl
```diff
@@ -1,6 +1,7 @@
 @import "../../node_modules/tracim_frontend_lib/src/css/Variable.styl"
 
 a
+  color black
```
Skylsmoi: this commit https://github.com/tracim/tracim/commit/1bcc0c4d9f1997e01dccca7f8042d7d61e74d6e3#diff-603fe74d68ec0a8ae870de005c6e5cca actually did the exact opposite a few days ago.

The problem was to use the generic class .primaryColorFont on a link. It caused a specificity issue that made it impossible to have a link with the primaryColorFont.

So I removed the generic rule for links and we now have to set them manually.

You, then, have to remove this generic rule you just added and add some more specific rules to set the color of links that are still blue

frontend_app_file/src/component/PreviewComponent.styl
```diff
@@ -83,6 +83,7 @@
     &__icon
       font-size 20px
       cursor pointer
+      color black
```
Skylsmoi: you should use the variable `fontColor` for the default black font color

Issue #1514, opened 2019-03-24
Z-Index too high for copyright in login-page.


inkhey:
## Bug description & symptom
Z-index is too high for copyright in login page.

> Page de login : "copyright © 2013 - 2018 tracim.fr" en zindex trop élevé (il passe par dessus le bouton valider). il faudrait que ça passe sous le formulaire.

## Expected
reduce it.

gbonaspetti:
The copyright didn't have a high z-index, so I changed the z-index from the card to be a little higher.

Skylsmoi:
Also change the year of the copyright to 2019
Do the same for the sidebar

PhilippeAccorsi:
Verified.

PR #1741, opened 2019-05-17
[#1514] change card z-index to be bigger then copyright

gbonaspetti:
https://github.com/tracim/tracim/issues/1514


PhilippeAccorsi:
Related also to #1742

Issue #1767, opened 2019-05-21
Removes cy.wait


basilelegal:
Sometimes cypress tests don't pass on travis but pass on our local machine.

A quick fix for these tests is to use cy.wait, it's a bad solution and we should removes them when we have time.

basilelegal:
The bug should be fixed by #1775 we can remove `cy.wait` when its confirmed

basilelegal:
random failing test
![Screenshot at 2019-06-05 06-05-59](https://user-images.githubusercontent.com/7879232/58949119-b0d37d00-8759-11e9-978a-5f19468689f4.png)

raphj:
Is this issue fixed? `grep -r cy.wait` return several occurrences in `functionnal_tests`.
@Skylsmoi

Skylsmoi:
well, if there are occurrences, it seems that it isn't fixed ^^;

raphj:
An instance of this issue I just closed: https://github.com/tracim/tracim/issues/2522

PR #1720, opened 2019-05-15
Fix/1688  fix app folder closing and switch from admin user and workspace

Skylsmoi:
https://github.com/tracim/tracim/issues/1688

PR #1775, opened 2019-05-21
Fix an error in redirection after a 401 causing cypress to crash

basilelegal:
#1767


Skylsmoi:
checks are in failure because they have been canceled to give priority to https://github.com/tracim/tracim/pull/1720
They will need to be restarted after

PhilippeAccorsi:
This is an error of merge but it's not a problem, because this fix is present first in branch 1720 (merged first)

Issue #1781, opened 2019-05-21
Sidebar breaks in Administration when we have a big description for a Shared space


gbonaspetti:
## Bug description & symptom

*If we put a big description for a shared space and we go to Administration > Shared space, the sidebar it's broken (see the gif).*

## How to reproduce

1. *Put a big description to a shared space*
2. *Go to Administration > Shared space*

## Screenshots

![Peek 2019-05-21 17-04](https://user-images.githubusercontent.com/35422692/58108073-6eebf800-7beb-11e9-85af-6f8a6bae7592.gif)

gbonaspetti:
I've limited the description and title width in *Administration > Shared* spaces and added ellipses if the height exceeds one line.

PhilippeAccorsi:
This is not working, text is visible just in one line.

gbonaspetti:
Apparently, it is necessary to have a max-width in order for the ellipsis to work within a table. So, I added max-width 0

Information taken from here: https://stackoverflow.com/questions/9789723/css-text-overflow-in-a-table-cell

LucBarry:
Verified

LucBarry:
Verified

Issue #1757, opened 2019-05-20
Sidebar responsive problem in xl screen


gbonaspetti:
## Bug description & symptom

*Sidebar is chopped at xl screens in the Users page.*

## Expected

*Sidebar works like in other screens.*

## How to reproduce

1. *Put the screen configuration whit a width higher then 1200px*
1. *Go to Administration > Users*

## Screenshots

![SideBarUsers](https://user-images.githubusercontent.com/35422692/58015151-c9595b80-7afa-11e9-8969-6159460f61b2.png)

gbonaspetti:
Linked to https://github.com/tracim/tracim/issues/1572.
As we can see at page 105 of Flexbox in CSS (https://books.google.fr/books?id=xwsmDwAAQBAJ&dq), the image shows that if we use flex-basis 0% we give more importance to the flex parameters grow and shrink that the content, and its works for us.
![Screenshot from 2019-05-22 14-57-01](https://user-images.githubusercontent.com/35422692/58188465-ec316e80-7cb8-11e9-8516-713169fd2789.png)
 So I change the properties deleting the lines specifics for change this at componentWillUnmount and componentDidMount.

celiamgtt:
Verified

PR #1787, opened 2019-05-22
[#1781] limiting description and title width in Administration

gbonaspetti:
https://github.com/tracim/tracim/issues/1781

### DIFF
--- a/frontend_app_admin_workspace_user/src/css/index.styl
+++ b/frontend_app_admin_workspace_user/src/css/index.styl
@@ -19,7 +19,6 @@
     &:hover, &:nth-child(even):hover
       background-color grey-hover
     &__td-link
-      max-width 120px
       white-space nowrap
       overflow hidden
       text-overflow ellipsis
@@ -27,7 +26,6 @@
       &:not(:hover)
         color inherit
     &__td-description
-      max-width 440px
       white-space nowrap
       overflow hidden
       text-overflow ellipsis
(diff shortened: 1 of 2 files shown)


frontend_app_admin_workspace_user/src/css/index.styl
```diff
@@ -19,9 +19,18 @@
...
     &__td-link
+      max-width 120px
+      white-space nowrap
+      overflow hidden
+      text-overflow ellipsis
       cursor pointer
       &:not(:hover)
         color inherit
+    &__td-description
+      max-width 440px
```
Skylsmoi: This value doesn't work on small devices.
You should add some breakpoint to reduce it.

frontend_app_admin_workspace_user/src/css/index.styl
```diff
@@ -19,9 +19,18 @@
     &:hover, &:nth-child(even):hover
       background-color grey-hover
     &__td-link
+      max-width 120px
```
Skylsmoi: This value doesn't work on small devices.
You should add some breakpoint to reduce it.

lebouquetin:
Note for both of you: we do not expect admin screen to be responsive. It's acceptable to have admin screens only with a computer-targetted rendering

gbonaspetti:
For this solution to work, we need to solve the issue https://github.com/tracim/tracim/issues/1757 first.

Issue #1788, opened 2019-05-22
switch from one sharedspace to another not close oppened app and its impossible to open same app in this new shared space


PhilippeAccorsi:
## Bug description & symptom

When I switch from one shared space to another one, app opened is not close. And after that if I try to open new content with same app, app is close but not open with new content.

## Diagnostic

*"unknown"*

## Expected

- [x] switching shared space close openned app
- [x] after swith I can open content with same app

## How to reproduce

- open one shared space
- open one content in this shared space (ex: html-document)
- switch to another shared space
=> app is not close but you are not in good shared space
- try to open content with same app (one html-document)
=> app visible is close but is not oppenned with new content

## Screenshots

![bug_switch_sharedspace](https://user-images.githubusercontent.com/25848257/58174806-052d2600-7c9f-11e9-8803-112e150e9813.gif)

## Impacted components

- *frontend*
- *htmldoc / file / thread app*

Skylsmoi:
fixed apps feature not opening when switching workspace + fixed apps fullscreen that could not refresh itself when needed (especially the app admin)

PhilippeAccorsi:
Verified.

PR #1791, opened 2019-05-22
[#1757] App admin workspace user flex auto issue

gbonaspetti:
https://github.com/tracim/tracim/issues/1757


Skylsmoi:
I need to fix https://github.com/tracim/tracim/issues/1788 before being able to properly test the fix
The review will have to wait a bit

Issue #1804, opened 2019-05-24
browser save file in cache also when you update file.


PhilippeAccorsi:
## Bug description & symptom

I download a file and after that I update this file with new revision. When I download this file again, the browser use file in cache and not the new file uploaded in tracim.

## Diagnostic

* "unknown".*

## Expected

- [ ] when file is updated, browser download new revision of file and not use file in cache.
Idea: add something at end of url to force new download of file
- [ ] working also for full pdf and pdf of one page

## How to reproduce

- upload one file on trcim
- download this file
- update this file
- download again this file
=> the second file downloaded is the same of first download

## Impacted components

- *frontend*

PhilippeAccorsi:
Problem also with preview of file.
Fix in this commit https://github.com/tracim/tracim/commit/e699edadfff59024c52552456a0cf642aad03d74

PhilippeAccorsi:
Verified.

PR #1805, opened 2019-05-24
[#1804] add revision id to download urls of app file to avoid cache o…

Skylsmoi:
…n new revision

https://github.com/tracim/tracim/issues/1804


PhilippeAccorsi:
We know why one cypress test failed. Fix is in progress.
We can merge this PR to make tracim 2.2. rc6.

Issue #1815, opened 2019-05-28
avatar and name about last change in content timeline is not the name of present user when status is changed


PhilippeAccorsi:
## Bug description & symptom

When we change status the name of author of this change is not the present user but the author of this content.

## Expected

- [ ] show the name of user who changed the status of content
 
## How to reproduce

- open one content created by another user
- change its status
- look at the name of user who changed the status

## Impacted components

- *backend*

inkhey:
This bug concern : 
- move
- copy
- change status

After rapid check, that's a backend bug. Setting of owner of revision is not done in those case, so owner remain same as latest.

lebouquetin:
@PhilippeAccorsi you write that backend and frontend are impacted. Could you check directly in the API if the returned information is good. If so, then the issue is frontend, if info is wrong, then the only impacted component is backend

PhilippeAccorsi:
I have updated issue (just backend issue)

@inkhey all situations:

- create
- delete
- restore delete
- archive
- restore archive
- status
- label
- edit / description / raw content
- update file
- move
- copy
- comment

gbonaspetti:
Verified.

PR #1828, opened 2019-05-29
Fix/1815 change owner for all actions on content

inkhey:
related to #1815

## Checkpoints
- [x] Code is clear enough
- [x] If there are FIXME in the code, then there are associated issues and issue is referenced in the FIXME -> No fixme
- [x] If there are TODO, NOTE or HACK in code, date and developer initials are present -> No fixme
- [x] Automated tests have been written (covering feature or non regression), or *at least* an issue has been created for test implementation -> associated test to this fixture now check also owner change for case copy/move/set_status


inkhey:
@Skylsmoi integration test failed here for unknown reason. That may mean that frontend expect uncorrect behaviour this issue fix.

backend/tracim_backend/lib/core/content.py
```diff
@@ -1405,7 +1406,7 @@ def _move_current(
         if new_parent and new_parent != item.parent:
             content_type = content_type_list.get_one_by_slug(item.type)
             self._check_valid_content_type_in_dir(content_type, new_parent, new_workspace)
-
+        item.owner = self._user
```
lebouquetin: I think `self._user` may be `None`. Please check, and if so, then the code should be `item.owner = self._user if self._user`

backend/tracim_backend/lib/core/content.py
```diff
@@ -1641,6 +1642,7 @@ def _add_copy_revisions(
             content=new_content,
             force_create_new_revision=True,
         ) as rev:
+            rev.owner = self._user
```
lebouquetin: `rev.owner = self._user if self._user`

backend/tracim_backend/lib/core/content.py
```diff
@@ -1362,6 +1362,7 @@ def restore_content_default_allowed_content(self, content: Content) -> None:
 
     def set_status(self, content: Content, new_status: str):
         if new_status in content_status_list.get_all_slugs_values():
+            content.owner = self._user
```
lebouquetin: `content.owner = self._user if self._user`

inkhey:
> I think you did not fix the bug in case the content or label are modified

Issue didn't happened in thoses case.

backend/tracim_backend/lib/core/content.py

backend/tracim_backend/lib/core/content.py

backend/tracim_backend/lib/core/content.py
inkhey: this syntax is not correct , replace by : 
```
if self._user:
  item.owner = self._user
```

buxx:
Travis pipelines are green (but github say "in progress", i think it is sticked). So i merge.

Issue #1807, opened 2019-05-24
improve travis testing with specific parameter in commit message


PhilippeAccorsi:
## Feature description

Add some parameter in travis to test just backend or frontend when is it necessary.

## Context

At present travis test backend and frontend for each commit. If we add `[skip ci]` travis is not started. We need to make travis more quick if we push just frontent commit (for test before MR) or just backend.

## Prerequisites

Some good documentation/idea:
~~~
Build only when the commit message matches against the given regular expression:
    commit_message !~ /(no-deploy|wip)/
~~~
https://docs.travis-ci.com/user/conditions-v1
With that we can activate just backend or just frontend stage if other stage is not necessary. Exemple: if we add [frontend] in commit message, travis just start stage for frontent. We can save lot of time with this.

https://docs.travis-ci.com/user/build-stages/
https://docs.travis-ci.com/user/conditional-builds-stages-jobs/
https://docs.travis-ci.com/user/conditions-testing

## Impacted components

- *backend*
- *frontend*

PhilippeAccorsi:
@Skylsmoi @basilelegal @inkhey @buxx @lebouquetin 
Do you think its a good idea to look in this direction to see if its possible to save some time when we push some commit ?

buxx:
Yes i think so.

PhilippeAccorsi:
To fix this we have added this possibility:

- skip cypress test by adding [skip ci cypress] in commit message
- skip pytest test by adding [skip ci pytest] in commit message

PhilippeAccorsi:
Verified.

PR #1837, opened 2019-06-04
Fix/1807  add possibility skip cypress or pytest

PhilippeAccorsi:
- [x] add possibility to skip cypress test by adding `[skip ci cypress]` in commit message
- [x] add possibility to skip pytest by adding `[skip ci pytest]` in commit message


PhilippeAccorsi:
Related to #1807

Issue #1734, opened 2019-05-17
webdav not working with MacOS (v10.14.4)


PhilippeAccorsi:
## Bug description & symptom

When we try to move some file with Webdav and finder, one file ._*** is created for each file (we think is temporary file) and one file is created with normal name. But no data is written. So this two file are empty and file ._*** is not deleted.

After this try, its not possible to reconnect with webdav. Thats also not working with linux.

## Diagnostic

*"unknown"*

## Expected

- [ ] it should be possible to upload files with finder in macos
- [ ] upload file with finder not create temporary file **or** this file is not detected like a good file by tracim to make sure no notification is send. *note de damien: we should probably consider ._xxx as temporary files like we do for windows and do not store them (i think this is what we do currently)*

## Context and version information

- Desktop environment 
    - *OS: MacOs 10.14.4*
    - *Soft: finder*

## Prerequisites

*"unknown"*

## How to reproduce

- Open finder
- try to move file with webdav

## Impacted components

- *backend*
- *webdav*

inkhey:
Bug cannot be reproduce with waitress 1.1.0, but happened with uWSGI 2.0.14-debian

lebouquetin:
@inkhey please test with latest uwsgi version (2.0.18)

inkhey:
Latest version of uwsgi(2.0.18) doesn't work.

After some research, i found that problem is link to the fact we send this header:
`Transfer-Encoding`: chunked and no `Content-Length` .
https://en.wikipedia.org/wiki/Chunked_transfer_encoding

Uwsgi is not able to support this kind of chunked transfer encoding.

There is a discussion about a feature about a flag `--wsgi-manage-chunked-input` which **may** solve the issue : https://github.com/unbit/uwsgi/issues/1428 but this feature doesn't exist yet in last version of uwsgi.

Waitress server which work doesn't really use chunk for security reason, as found here:
https://github.com/Pylons/waitress/issues/179 and https://github.com/Pylons/waitress/pull/203 . In fact it convert chunked file into normal one with buffering, and set Content-Length header.

One idea to verify behavior in chunked context (which doesn't happened in waitress case), may be testing with cheroot, which is the server used by wsgidav sample.

Another way to fix this issue is by using same mecanism as waitress in webserver proxy.
for `mod_proxy_http` of apache , https://httpd.apache.org/docs/trunk/mod/mod_proxy_http.html 
we can use:
`SetEnv proxy-sendcl`
Tested and this is working.

[...]

PR #1842, opened 2019-06-04
use proxy_sendcl in apache sample used for docker

inkhey:
related to #1734 

## Checkpoints
- [X] Code is clear enough.
- [x] Manual check of docker created.

We need to verify docker builded with this work properly and fix the issue.


inkhey:
verification to do : 
- build docker image
- check if docker image run properly
- try upload of file with docker in docker image with mac os x

PhilippeAccorsi:
Tested with docker image (on debian) and webdav with macbook

Need future improvement to not save or not show temporary file created by macos with webdav.

PR #1854, opened 2019-06-05
Fix/1109  sanitize script in comments

basilelegal:
Now removes scripts tags from comments, if there's only a script in the comment it'll refuse it as an empty comment.

Need to be merged after #1851 

## Checkpoints

- [x] Code is clear enough


buxx:
Travis push pipeline green but travis pr pipeline red because develop have an fmt error. So we can merge.

Issue #1844, opened 2019-06-04
add version of tracim directly in frontend code (visible on top of copyright)


PhilippeAccorsi:
## Feature description

Its possible to know wich version of tracim is used.

## expected

- [ ] add version of tracim directly on top of copyright. Center this information.

## Context

At present its not easy to know if client use last version of tracim or not.

## Impacted components

- *frontend*

## Extra information

For the moment its just a small fix in frontend.
One API exist but the API not return all information correctly for the moment, so we need to fix this quickly for tracim 2.3   
In the futur, this value need to be changed dynamicaly.

PhilippeAccorsi:
We need to define what is the good solution first to fix this quick without loss time when we make final solution.

lebouquetin:
The goal is to have a constant used in frontend which is easy to update without risk of breaking the code.

celiamgtt:
Working, but the current version displayed is 2.2 instead of 2.3

celiamgtt:
We'll change the version later in the process.

PR #1857, opened 2019-06-05
front [skip ci pytest] - adds the frontend version in the sidebar

basilelegal:
#1844

## Checkpoints

- [x] Code is clear enough

### DIFF
--- a/frontend/src/helper.js
+++ b/frontend/src/helper.js
@@ -2,7 +2,7 @@ import i18n, { getBrowserLang } from './i18n.js'
 
 const configEnv = require('../configEnv.json')
 
-export const TRACIM_APP_VERSION = 'V2.2'
+export const TRACIM_APP_VERSION = 'v2.2'
 
 // this function is declared in i18n to avoid cyclic imports and reexported here for consistency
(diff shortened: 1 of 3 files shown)


frontend/src/helper.js
```diff
@@ -2,6 +2,8 @@ import i18n, { getBrowserLang } from './i18n.js'
 
 const configEnv = require('../configEnv.json')
 
+export const TRACIM_APP_VERSION = 'V2.2'
```
Skylsmoi: PLEASE USE MINOR 'v' KTHX §

frontend/src/helper.js
```diff
@@ -2,6 +2,8 @@ import i18n, { getBrowserLang } from './i18n.js'
 
 const configEnv = require('../configEnv.json')
 
+export const TRACIM_APP_VERSION = 'v2.2'
```
lebouquetin: I would prefer (if possible) to have a version.js file including online this part. This way, it's clear for the release manager that version must be updated in version.js file

basilelegal: ok

Issue #1333, opened 2019-01-17
Prod docker: activate webdav, async and reply by email by default


PhilippeAccorsi:
**At present**

By default for using webdav, async and reply by email, we need to add env variable to start docker.

**Expected**

Production use a special config:
- [x] In entrypoint: activate by default webdav, mail_fetcher and mail_notifier
- [x] In development.ini: async is used by default

PhilippeAccorsi:
We activate webdav, caldav and async email by default.
Its possible to activate reply by email and email notification (maybe its not really necessary if we used env variable for all config)

Mail fetcher is not activated by default because its necessary to add imap config first.

PhilippeAccorsi:
Verified.

PR #1859, opened 2019-06-06
Fix/1333  activate feature of tracim by default

PhilippeAccorsi:
- [x] activate caldav by default (its possible to deactivate caldav)
- [x] activate webdav by default (its possible to deactivate webdav)
- [x] add possibility to activate notification by email and reply by email (maybe not really necessary if we use env variable for config)
- [x] tracim use async by default (mail_notifier is activated)

Related to #1333


tools_docker/Debian_Uwsgi/common.sh
```diff
@@ -19,6 +19,10 @@ if [ ! -f /etc/tracim/development.ini ]; then
     sed -i "s|webdav.listen = .*|webdav.listen = 127.0.0.1:3030|g" /etc/tracim/development.ini
     sed -i "s|;webdav.root_path = /|webdav.root_path = /webdav|g" /etc/tracim/development.ini
     sed -i "s|webdav.base_url = .*|webdav.base_url = http://localhost:8080|g" /etc/tracim/development.ini
+    sed -i "s|email.processing_mode = sync|email.processing_mode = async|g" /etc/tracim/development.ini
+    sed -i "s|; email.async.redis.host = .*|email.async.redis.host = localhost|g" /etc/tracim/development.ini
+    sed -i "s|; email.async.redis.port = .*|email.async.redis.port = 6379|g" /etc/tracim/development.ini
+    sed -i "s|; email.async.redis.db = .*|email.async.redis.db = 0|g" /etc/tracim/development.ini
```
lebouquetin: what is "0" for `email.async.redis.db` ? its the name of the database?

PhilippeAccorsi: I don't know :(
Its the default value in development.ini.sample (and in code) (I don't have take time to know what is this parameter).

@inkhey can you explain here this parameter?

lebouquetin: Actually, this is the value that sounds strange (if its the db name I would have expected something like tracim_cache or similar... but maybe it's redis which only use numbers for naming databases/indexes?)

PhilippeAccorsi: This paramter seem to be used if you used multiple redis database. But by default the config (in documentation) is `null`. I don't know if `null` and `0` is the same for redis.

PhilippeAccorsi: Redis used number and not name for database.

lebouquetin: ok

Issue #1866, opened 2019-06-06
End point move content returns 500 when trying to move a content into itself


Skylsmoi:
## Bug description & symptom

sending to the api this request:
`/api/v2/workspaces/3/contents/14/move`
with this body:
`{"new_workspace_id":3,"new_parent_id":14}`

returns error 500:
`{"details": {"error_detail": {}}, "code": null, "message": "maximum recursion depth exceeded while getting the str of an object"}`

## Expected

- [ ] I would assume a 400 with a specific error code
- [ ] add an API test case (this is a border which will not be tested often if not auto-tested)

## How to reproduce

curl:
```
curl 'http://192.168.1.153:6543/api/v2/workspaces/3/contents/14/move' -X PUT -H 'Pragma: no-cache' -H 'Origin: http://192.168.1.153:8090' -H 'Accept-Encoding: gzip, deflate' -H 'Accept-Language: fr-FR,fr;q=0.9,en-US;q=0.8,en;q=0.7' -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/68.0.3440.84 Safari/537.36' -H 'Content-Type: application/json' -H 'Accept: application/json' -H 'Cache-Control: no-cache' -H 'Referer: http://192.168.1.153:8090/ui/workspaces/3/contents?folder_open=15,13,14,19' -H 'Cookie: defaultLanguage=fr; lastConnection=1; session_key=d9d7fb1b170383c67fea5f933898fe34d419f1f16a62f22f5bef43b0b1501f1f06c1797b' -H 'Connection: keep-alive' --data-binary '{"new_workspace_id":3,"new_parent_id":14}' --compressed
```

Note that it was taken from a development server

## Impacted components

Backend API

## Extra information

Frontend will do the necessary to avoid to be able to send such request that makes no sens

Skylsmoi:
Note that the same occur if you try to move a folder in one of its subfolder

lebouquetin:
This is another border case to test (especially when moving to an indirect child).
@Skylsmoi you implement the check on frontend, right? (to avoid this case even without to send a request)

Skylsmoi:
Mmh I didn't plan to since I only have the immediate parent id for a given content.
But I can

tracim:
You must check this. You have all content available for checking this (you can just make a recursive call for this to check «on mouseover»)

gbonaspetti:
Verifed.

PR #1898, opened 2019-06-11
Fixes error when moving a folder in itself or in one of his children

basilelegal:
#1866 

Raise an error if you move a folder in itself or in one of his children. 

## Checkpoints

- [x] Code is clear enough.
- [x] An issue have been created for any new FIXME.
- [x] Comments are formated correctly.
- [x] Code is tested or an issue have been created to write tests.


backend/tracim_backend/lib/core/content.py
```diff
@@ -1403,6 +1405,10 @@ def move(
         must_stay_in_same_workspace: bool = True,
         new_workspace: Workspace = None,
     ) -> None:
+        if item.content_id == new_parent.content_id:
+            raise ConflictingMoveInItself("You cannot copy a folder into itself")
```
lebouquetin: Technically, `move` is an operation on any type of content, not specified to folders. For example, even if the user interface does not allow it (at the moment), you can move a comment to another content. So the messages must be more generic:

- "You cannot copy a folder into itself" => "You can't move a content into itself" (by the way, it's not a copy, it's a move)
- "The destination folder is in the source folder" => "You can't move a content into one of its children")

backend/tracim_backend/lib/core/content.py
```diff
@@ -1403,6 +1405,10 @@ def move(
...
+        if new_parent in item.children:
```
lebouquetin: children includes all children recursively?

backend/tracim_backend/tests/functional/test_contents.py
```diff
@@ -29,6 +29,26 @@ class TestFolder(FunctionalTest):
 
     fixtures = [BaseFixture]
 
+    def _setup_basics(self) -> None:
```
lebouquetin: this implementation is something like `setUp()` / `tearDown()` pattern, no?
The way to implement it is to define a `setup_class()` class method, https://docs.pytest.org/en/latest/xunit_setup.html

lebouquetin: You seem not to share the use of `_setup_basics()` in all your test cases... so probably you must implement a new FolderTest class (eg `TestFolderMove`)

backend/tracim_backend/tests/functional/test_contents.py
```diff
@@ -29,6 +29,26 @@ class TestFolder(FunctionalTest):
...
+        self.dbsession = get_tm_session(self.session_factory, transaction.manager)
+        self.admin = self.dbsession.query(User).filter(User.email == "admin@admin.admin").one()
+        self.workspace_api = WorkspaceApi(
+            current_user=self.admin, session=self.dbsession, config=self.app_config
+        )
+        self.content_api = ContentApi(
+            current_user=self.admin, session=self.dbsession, config=self.app_config
+        )
+        self.workspace = self.workspace_api.create_workspace(label="test", save_now=True)
+        self.folder = self.content_api.create(
```
lebouquetin: this data is part of the test case, not the context, so it should move into the test body itself. otherwise, the reading of the test case `test_api__move_folder_into_itself__err_400` is not clear: We do not understand what `self.folder` is.

setup() and tearDown() are about context, all data directly included in the test must be created in the test.

Issue #1900, opened 2019-06-12
Fix memory streaming upload during in webdav


inkhey:
## Bug description & symptom

#696 fix streaming upload for tracim web case, but webdav context is not currently doing proper streaming.

## Diagnostic

example with small file (44mo) and big file (512mo):
![image18](https://user-images.githubusercontent.com/7851652/59337518-0cc17700-8d01-11e9-8270-0074d2066446.png)

Problem is linked to multiple troubles:
- use of .read() on self._file_stream
- mecanism of FakeFileStream save itself data in memory until close where data is properly saved.

## Expected

- We should save content by chunck without storing all content in memory. Still unclear how we technically do this.

inkhey:
we decide to temporary fix memory issue here using temp file, see #1907, for true streaming upload mecanism , see issue #1911

inkhey:
Verified

PR #1907, opened 2019-06-13
Fix/1900 streaming upload for webdav

inkhey:
related to #1900.
This does:
- webdav upload using a temporary file instead of in memory
- webdav download of file object does not read whole file in memory before sending.

This does not fix:
- use true streaming mecanism (use temporary fix instead) see #1911 
- fix memory issue when downloading big non-file content (html-document, threads) see #1913 

## Checkpoints

- [x] Code is clear enoug

### DIFF
--- a/backend/tracim_backend/lib/webdav/resources.py
+++ b/backend/tracim_backend/lib/webdav/resources.py
@@ -960,6 +960,8 @@ class OtherFileResource(FileResource):
 
     @webdav_check_right(is_reader)
     def getContent(self):
+        # TODO - G.M - 2019-06-13 - find solution to handle properly big file here without having
+        # big file in memory. see https://github.com/tracim/tracim/issues/1913
         filestream = compat.BytesIO()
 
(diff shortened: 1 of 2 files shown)


lebouquetin:
@inkhey what do you mean by : "use true streaming mecanism (use temporary fix instead)" ?

lebouquetin:
my feeling is we can merge even if the memory issue is not fixed for not-a-file contents. The status with this fix is "tracim will be better than it already is", no?

inkhey:
> @inkhey what do you mean by : "use true streaming mecanism (use temporary fix instead)" ?

Normally we should not store data into tempfile but directly in depot file, but because of the wsgidav code, i do not see easy way to do this. That's why i proposed this fix to avoid ram issue which it many time worst than using temp file.

> my feeling is we can merge even if the memory issue is not fixed for not-a-file contents. The status with this fix is "tracim will be better than it already is", no?

Yes.

Issue #1812, opened 2019-05-27
Config SQLALCHEMY__URL should be mandatory


buxx:
## Bug description & symptom

If `sqlalchemy.url` is not set, Tracim start. But crash with non explicit error.

## Diagnostic

This parameter should be considered as required at config initialization.

## Expected

Explicit error should be raised if config not provided OR (discuss it) use a default sqlite.

## Context and version information

All Tracim versions

## Prerequisites

no one

## How to reproduce

Remove `sqlalchemy.url`  config and start Tracim.

## Impacted components

*Optional - list components which will be impacted.*

*For example:*

- *backend*

inkhey:
@buxx i do think, we should return an error.
This was problematic before as sqlalchemy_url wasn't a true "tracim parameter". That's now ok, so we can add mandatory mecanism like other parameter.
We should discuss if we want to have similar mecanism for alembic, which mean some specific code in alembic part.

Skylsmoi:
I'm seeing this error `tracim_backend.exceptions.ConfigurationError: ERROR: "SQLALCHEMY__URL" configuration is mandatory .Set it before continuing.`

I find it explicit enough.
Although there is a punctuation typo around the dot

PR #1917, opened 2019-06-14
Fix/1812 make sqlalchemy  url mandatory

inkhey:
related to #1812

## Checkpoints

- [ ] Code is clear enough

### DIFF
--- a/backend/tracim_backend/migration/env.py
+++ b/backend/tracim_backend/migration/env.py
@@ -8,6 +8,7 @@ from alembic import context
 from sqlalchemy import engine_from_config
 from sqlalchemy import pool
 
+from tracim_backend.exceptions import ConfigurationError
 from tracim_backend.models.meta import metadata
 from tracim_backend.models.setup_models import *  # noqa: F403,F401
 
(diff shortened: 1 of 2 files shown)


backend/tracim_backend/migration/env.py
```diff
@@ -21,6 +22,8 @@
 env_sqlalchemy_url = os.environ.get("TRACIM_SQLALCHEMY__URL")
 if env_sqlalchemy_url:
     config.set_main_option("sqlalchemy.url", env_sqlalchemy_url)
+if not config.get_main_option("sqlalchemy.url"):
+    raise ConfigurationError("SQLALCHEMY__URL is mandatory")
```
lebouquetin: It seems not clear to me : you check `sqlalchemy.url` but you raise `SQLALCHEMY__URL` error. Do your code check the url from both config and envvar configuration?

inkhey: Yes, it does it, sqlalchemy.url in config of alembic is at this point the correct value to check (there is no CFG here).
If you check the lines before we override config if there is env_var setted.

inkhey: as SQLALCHEMY__URL is the parameter name of tracim that the best thing to return as mandatory.

inkhey:
see #1920 for better fix for alembic context.

Issue #1955, opened 2019-06-19
Search returns the same result twice


gbonaspetti:
## Bug description & symptom

Sometimes the search return the same result twice for the searched keywords.

## Expected

Unique results for the same searched keywords.

## Screenshots

![Screenshot from 2019-06-19 11-55-59](https://user-images.githubusercontent.com/35422692/59756181-83beb880-9289-11e9-8f6e-c3c001222a9f.png)

inkhey:
Verified by reproducing case where backend return doublon :  happened when we get same content in multiple pages (bad remove duplicate code in backend) as we have multiple comment returning it.

PR #1957, opened 2019-06-19
[#1955] filter to unique results at search

gbonaspetti:
#1955 

## Checkpoints

- [x] Code is clear enoug

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -33,6 +33,7 @@
     "i18next": "^10.5.0",
     "js-cookie": "^2.2.0",
     "lodash": "^4.17.11",
+    "lodash.uniqby": "^4.7.0",
     "prop-types": "^15.6.0",
     "query-string": "^6.1.0",
     "radium": "^0.24.1",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -33,7 +33,6 @@
     "i18next": "^10.5.0",
     "js-cookie": "^2.2.0",
     "lodash": "^4.17.11",
-    "lodash.uniqby": "^4.7.0",
     "prop-types": "^15.6.0",
     "query-string": "^6.1.0",
     "radium": "^0.24.1",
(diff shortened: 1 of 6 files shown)


frontend/src/reducer/searchResult.js
```diff
@@ -17,18 +18,22 @@ const defaultResult = {
 }
 
 export default function searchResult (state = defaultResult, action) {
+  function uniqueResults (allResults) {
```
Skylsmoi: move this function at the top level of the file

frontend/package.json
```diff
@@ -33,6 +33,7 @@
     "i18next": "^10.5.0",
     "js-cookie": "^2.2.0",
     "lodash": "^4.17.11",
+    "lodash.uniqby": "^4.7.0",
```
Skylsmoi: I forgot that lodash was already a dependency
so remove this line (using the npm command line)

frontend/src/reducer/searchResult.js
```diff
@@ -17,18 +18,22 @@ const defaultResult = {
...
     case `${APPEND}/${SEARCH_RESULTS_LIST}`:
-      return {...state, resultsList: [...state.resultsList, ...action.appendSearchResultsList]}
+      return {...state, resultsList: uniqueResults([...state.resultsList, ...action.appendSearchResultsList])}
 
     case `${SET}/${SEARCHED_KEYWORDS}`:
       return {...state, searchedKeywords: action.searchedKeywords}
 
     case `${SET}/${CURRENT_NUMBER_SEARCH_RESULTS}`:
-      return {...state, currentNumberSearchResults: action.currentNumberSearchResults}
+      return {...state, currentNumberSearchResults: state.resultsList.length}
```
Skylsmoi: since this reducer doesn't use the parameter action anymore, it should be removed from the action in action-creators.sync.js too.
But if it is now only calculated from resultsList.length, it's better to use a constant equals to props.resultsList.length in the component where you need it

Issue #1981, opened 2019-06-20
In xl screens, the Advanced Dashboard button is under the square with the user name


gbonaspetti:
## Bug description & symptom

If the screen width is between 1200px and 1246px, the Advanced Dashboard button is 
under the square with the user name (only in French).

## Expected

That the button is fully visible.

## Screenshots

![Screenshot from 2019-06-20 16-20-07](https://user-images.githubusercontent.com/35422692/59856544-6c5df900-9377-11e9-9441-32b071c08cae.png)

celiamgtt:
Button cropped sometimes in french version
![Capture d'écran de 2019-06-21 10-05-00](https://user-images.githubusercontent.com/50453784/59908281-5b5fc700-940d-11e9-9f6f-b86510db75ed.png)

celiamgtt:
Verified

PR #1987, opened 2019-06-21
[#1981] changing Advanced Dashboard button position

gbonaspetti:
#1981 

## Checkpoints

- [x] Code is clear enough

### DIFF
--- a/frontend/src/css/Generic.styl
+++ b/frontend/src/css/Generic.styl
@@ -11,7 +11,7 @@ a
   vertical-align middle
 
 .outlineTextBtn
-  padding 10px 14px
+  padding 10px 15px
   background-color transparent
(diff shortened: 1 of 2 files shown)


frontend/src/css/Generic.styl
```diff
@@ -11,7 +11,7 @@ a
   vertical-align middle
 
 .outlineTextBtn
-  padding 10px 15px
+  padding 10px 14px
```
Skylsmoi: please don't edit this generic class
I prefer not to because we are on the verge of new version, it's not the right time to change such a generic css rule

Skylsmoi: prefer reducing the font size for now

Issue #1973, opened 2019-06-20
Backend simple search should not return doublon.


inkhey:
## Bug description & symptom

Doublon issue with simple search is solved here in frontend: https://github.com/tracim/tracim/issues/1955
But that's also mean backend always does not return proper result.

## Expected

- Do not return anymore doublon result with simple search
- non regression test in our case (does happened with comment returning to same content): add a content with multiples comment with search string into them and verifiy if not doublon are found.

PhilippeAccorsi:
@inkhey what kind of simple test can I make to test this feature?

inkhey:
> @inkhey what kind of simple test can I make to test this feature?

Not so easy. You should verify that /api/v2/search/content doesn't return at all doublon.

The Easiest way to check this is done by non-regression test here, it is to create a tracim with 2 comments , having a keyword in both comments, search by size of "1" and verify if page 2 is empty. (there is just one content, you should not return multiple items).

PhilippeAccorsi:
Ok. So Test of non-regression exist, so Its ok for me.
Thanks.

Verified.

PR #2004, opened 2019-06-24
[2.4.0]Fix/1973 fix doublon in simple search backend

inkhey:
related to #1973

## Checkpoints

- [x] Code is clear enough


backend/tracim_backend/lib/core/content.py
```diff
@@ -2158,8 +2158,8 @@ def search(
                 # INFO - G.M - 2019-06-13 - avoid duplication of same content in result list
                 continue
             if current_offset >= offset:
-                parsed_content_ids.append(content.content_id)
                 results.append(content)
+            parsed_content_ids.append(content.content_id)
             current_offset += 1
 
         return results, current_offset
```
lebouquetin: I see now that we return a tuple. We must avoid this kind of behavior which is hard to understand: the typing help understanding the type of data, but understand what the data is requires to read the code.

inkhey: code changed in order to don't need them.

backend/tracim_backend/lib/core/content.py
```diff
@@ -49,6 +49,7 @@
 from tracim_backend.exceptions import WorkspacesDoNotMatch
 from tracim_backend.lib.core.notifications import NotifierFactory
 from tracim_backend.lib.core.userworkspace import RoleApi
+from tracim_backend.lib.search.models import SimpleContentSearchResponse
```
lebouquetin: I suggest to rename `SimpleContentSearchResponse` to `SimpleContentSearchResult`?

lebouquetin: By the way, the search feature must be moved to `lib.search.xxx` module. There is no reason for `lib.core.content` having a dependency on `lib.search.models`.

What you should do is:

- have a *simple search* module which has a dependency on `lib.core.content` and which implements the search method.
- have a dependency on `lib.search.models` in *simple search* module
- have a `lib.core.content` module which does not known anything about search.

inkhey: Is the new implementation ok ?

lebouquetin:
@inkhey if only copy/paste, then split clearly core / simple / elasticsearch parts of the search before we merge

Issue #2005, opened 2019-06-24
starting docker without env variable have not good workflow


PhilippeAccorsi:
## Bug description & symptom

If you just migration tracim 2.2 to 2.3, you lost email notification and reply by email functionnality; if you don't use env variable.

## Diagnostic

We can see this in common.sh script:
~~~
# Activate email
if [ "$EMAIL_NOTIFICATION" = "1" ]; then
    sed -i "s|email.notification.activated = .*|email.notification.activated = True|g" /etc/tracim/development.ini
else
    sed -i "s|email.notification.activated = .*|email.notification.activated = False|g" /etc/tracim/development.ini
fi

# Activate reply by email
if [ "$REPLY_BY_EMAIL" = "1" ]; then
    sed -i "s|email.reply.activated = .*|email.reply.activated = True|g" /etc/tracim/development.ini
else
    sed -i "s|email.reply.activated = .*|email.reply.activated = False|g" /etc/tracim/development.ini
fi
~~~

Its not a good idea to use else without controlling value of the parameter.
We need to explicit: if parameter is "1" else if parameter is "0".

## Expected

- [ ] add "if condition is 0" in else function.
- [ ] verify if no other change is needed for another variable

## How to reproduce

Start tracim 2.3 on same path of tracim 2.2 without adding env variable.

## Impacted components

- *backend*

PhilippeAccorsi:
It's now possible to start tracim with env variable so its not necessary to use this function.
If we want to start tracim with notification we can add all necessary variable directly to docker images.

And if you just want to deactivate notification for exemple, you can start docker with variable about notification activated with value "False". In this case value write in development.ini is not erase.

PhilippeAccorsi:
For using reply by email we need to activate specific service so "REPLY_BY_EMAIL=1" is necessary but all other parameter can be set directly with env variable.

PhilippeAccorsi:
Verified.

PR #2056, opened 2019-07-03
[2.3.1] [#2005]  starting docker without envvars not deactivate notification

PhilippeAccorsi:
- [x] repair workflow to not change something in develoment.ini if you don't use env variable
related to #2005 

## Checkpoints

- [x] Code is clear enough

### DIFF
--- a/tools_docker/Debian_Uwsgi/common.sh
+++ b/tools_docker/Debian_Uwsgi/common.sh
@@ -152,12 +152,16 @@ fi
 if [ "$EMAIL_NOTIFICATION" = "1" ]; then
     sed -i "s|email.notification.activated = .*|email.notification.activated = True|g" /etc/tracim/development.ini
 else
+    if [ "$EMAIL_NOTIFICATION" = "0" ]; then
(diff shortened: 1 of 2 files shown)


tools_docker/Debian_Uwsgi/entrypoint.sh
```diff
@@ -118,13 +118,6 @@ fi
 cd /tracim/backend/
 tracimcli search index-create -c /etc/tracim/development.ini
 
-if [ "$UPDATE_INDEX_ELASTICSEARCH" = "1" ]; then
```
inkhey: Why is this code is dropped ?

PhilippeAccorsi:
This branch is created from this one https://github.com/tracim/tracim/pull/2055

tools_docker/Debian_Uwsgi/entrypoint.sh
PhilippeAccorsi: We drop this feature for the moment because starting tracim with populate take too much time.
We added documentation to make this update manually directly in the container.

inkhey: ok.

Issue #619, opened 2018-07-03
Better fixtures and test


inkhey:
TODO:
- [X] wrapp lib usage in tests to avoid heavy refactoring when changing lib signature ( https://github.com/tracim/tracim/pull/2069)
- [x] convert current fixtures to pytest fixture style. ( https://github.com/tracim/tracim/pull/2089)
- ~~Avoid one fixture for all tests~~ . check #2120 
- ~~Create intelligent fixture, not used in all tests but common to some tests in order to avoid too much code duplication. This need discussion.~~ check #2120 
- ~~Avoid duplication of similar test, check for https://docs.pytest.org/en/latest/parametrize.html usage. this should probably be done after converting to pytest fixture style because parametrize doesn't work with pytest+parametrize by default , see https://stackoverflow.com/questions/18182251/does-pytest-parametrized-test-work-with-unittest-class-based-tests~~ . no  clear case where such modification is needed now.

PhilippeAccorsi:
@inkhey can you explain here what is done, what is not yet developed?
And create separated issue for each functionnality.

Its better to have for exemple five specific issue, because we can work on each when we have time and merge & close issue when work is done. Now we have just one "big" issue and we dont know what is really done, what is not easy to solved.

Thanks.

inkhey:
@PhilippeAccorsi i do think we can close this issue as #2120 new issue is created to fix problem that are not fixed here, other fix of tests (like using parametrisation) are too much unclear to be considered as true bug now. We do need case where current behavior is a problem to fix it.

PhilippeAccorsi:
I close this issue., because no needed test (all test is in CI).

PR #2069, opened 2019-07-05
Fix/619  wrapp internal lib usage in tests

inkhey:
related to #619 
First part: wrap internal lib usage in tests.

## Checkpoints

- [x] Code is clear enough


backend/tracim_backend/tests/__init__.py
```diff
@@ -97,7 +103,66 @@ def get_mailhog_mails(self) -> typing.List[typing.Any]:
         ).json()
 
 
-class FunctionalTest(unittest.TestCase):
+class TestWithHelpers(object):
+
+    session = None
```
buxx: Where is filled this attribute ?

inkhey: This attribute is field by other object. In fact session is by default None (to avoid error somewhere in code) but this class TestWithHelpers is used by other class which set session.

buxx: :ok_hand:

Issue #2092, opened 2019-07-12
Middle click on logo should open new tab


buxx:
## Feature description

When middle click on the logo, web browser should be able to open a new tab. 

## Context

Currently, middle click on the logo as same effect as left click.

## Impacted components

- frontend

PhilippeAccorsi:
Verified.

PR #2108, opened 2019-07-18
[Nice to have in 2.4] Fix/2092 logo route

metzgegu:
#2092 

## Checkpoints

- [x] Code is clear enough

### DIFF
--- a/frontend/src/component/Header/Logo.jsx
+++ b/frontend/src/component/Header/Logo.jsx
@@ -1,6 +1,7 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import { Link } from 'react-router-dom'
+import { PAGE } from '../../helper.js'
 
 const Logo = props => {
   return (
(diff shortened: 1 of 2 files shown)


frontend/src/component/Header/Logo.jsx
```diff
@@ -1,16 +1,17 @@
...
-    </a>
+    </Link>
   )
 }
 export default Logo
 
 Logo.propTypes = {
   logoSrc: PropTypes.string.isRequired,
-  onClickImg: PropTypes.func.isRequired
+  to: PropTypes.string
```
Skylsmoi: for each non required props, you should have a defaultProps value.
Also check that the default props works, meanings clicking on the logo without having a `to` specified to `<Logo />` doesn't generate a js error

metzgegu:
defaultprops values added

metzgegu:
Can't add Cypress test because Cypress don't support middle click

Skylsmoi:
we will do a unit test

Issue #2011, opened 2019-06-25
Adding an image to Tinymce causes an input to appear at the bottom of the screen


gbonaspetti:
## Bug description & symptom

Adding an image to Tinymce causes an input ("Choose file" + file name) to appear at the bottom of the screen and this allows a horizontal scroll that shouldn't exist.

## Expected

Adding an image does not impact the general interface, this input must be hidden.

## Context and version information

input id : hidden_tinymce_fileinput

## How to reproduce

1. Go to a content of any type
2. Click at Rich text in the timeline
3. Add an image
4. Scroll to the bottom of the page

(The same happens at Rich text in Text Documents)

## Screenshots

![Peek 2019-06-25 09-49](https://user-images.githubusercontent.com/35422692/60080504-5910ac00-9730-11e9-997e-45e427f28f71.gif)

PhilippeAccorsi:
Verified.

PR #2113, opened 2019-07-22
Fix/2011 Hide the file input when upload an image into tinymce

metzgegu:
#2011 

## Checkpoints

- [x] Code is clear enough

### DIFF
--- a/frontend/dist/assets/tracim/tinymceInit.js
+++ b/frontend/dist/assets/tracim/tinymceInit.js
@@ -72,8 +72,8 @@
             fileTag = document.createElement('input')
             fileTag.id = 'hidden_tinymce_fileinput'
             fileTag.type = 'file'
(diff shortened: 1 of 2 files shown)


frontend/dist/assets/tracim/tinymceInit.js
```diff
@@ -73,6 +73,7 @@
             fileTag.id = 'hidden_tinymce_fileinput'
             fileTag.type = 'file'
             $('body').append(fileTag)
+            $(`#${fileTag.id}`).hide()
```
Skylsmoi: It would be better to not have the input getting visible in first place instead of adding some code to hide it afterwards.

If it's really the only way of hiding it, explain why

metzgegu: Yes, it's changed

Issue #2096, opened 2019-07-12
[Regression] blank page visible if I open Shared spaces and notifications in user profile after comming from Admin > users


PhilippeAccorsi:
## Bug description & symptom

If I click on Shared spaces and notifications in user profil after opening this profile from administration users, A blank page is visible with this error in console:
~~~
TypeError: "n is undefined"
    r https://xxxx/assets/tracim.app.js:8
    fi https://xxxx/assets/tracim.app.js:8
    fi https://xxxx/assets/tracim.app.js:8
    Qo https://xxxx/assets/tracim.vendors~app.js:81
    Di https://xxxx/assets/tracim.vendors~app.js:81
    Ka https://xxxx/assets/tracim.vendors~app.js:81
    qa https://xxxx/assets/tracim.vendors~app.js:81
    Tu https://xxxx/assets/tracim.vendors~app.js:81
    Eu https://xxxx/assets/tracim.vendors~app.js:81
    Ru https://xxxx/assets/tracim.vendors~app.js:81
    En https://xxxx/assets/tracim.vendors~app.js:81
tracim.vendors~app.js:81:74997
~~~

## Diagnostic

*"unknown".*

## Expected

- [x] Its possible to open `Shared spaces and notifications` correctly
- [ ] ~write cypress test about this regression~

## Context and version information

Tracim 2.3.1

## Prerequisites

To test on develop branch, you need this fix first https://github.com/tracim/tracim/issues/2076

## How to reproduce

- On serve tracim 2.3.1
- activate email notification
- Create 2 users
- create one workspace
- add this two users in this workspce
- go in Administration > users
- Click on name of one users
=> now you can see profile of this users
- click on `Shared spaces and notifications`
=> tracim show blank page 

## Impacted components

* "unknown"*

metzgegu:
We can't write cypress test because now in our cypress test config, we don't have e-mail notification enabled

PhilippeAccorsi:
I think we need to start two kind of cypress test. First, each test where email notification is not necessary, second, each test where email notification is necessary.
In this case, we just need to activate email notification (without smtp configuration).

I think we need to add new test in travis with specific config of backend (config with email notification).

PhilippeAccorsi:
@inkhey what do you used to test email in backend test? Do you use specific config for that.

inkhey:
@PhilippeAccorsi we do support email notification test with mailhog feature, there is many working example in config. It mean specific configuration for test but also specific fixture for mailhog.
[...]

Issue #2116, opened 2019-07-24
Cypress should be able to run with email notification enabled


Skylsmoi:
## Feature description

Currently, Cypress only runs with email notification disabled.
Some Tracim's feature cannot be tested.

## Impacted components

Functional tests

## Extra information

See https://github.com/tracim/tracim/issues/2096#issuecomment-512826417 for some info

PR #2122, opened 2019-07-25
[Nice to have in 2.4] Fix/2096 Notifications settings in user profile

metzgegu:
#2096

## Checkpoints

- [x] Code is clear enough


Skylsmoi:
Some cypress test can be added for this fix:
- open an admin account page and check that each sub menu open without error (my profile, notification, pasword and agenda)
- the same test can be added to user account if not already exists (the test should be really similar)

metzgegu:
First we need to add Cypress test with notification enabled #2116

Skylsmoi:
That's right.
We'll do later.

Issue #2125, opened 2019-07-29
Public share of a file


lebouquetin:
## Feature description

A tracim user can share a file so that other users (not necessarily Tracim users) can download it (latest version)

In file application, a tab allow to share / unshare file:

- form includes an email field (which can contain one or several email adresses separated by coma or line return).
- an optional password field

Example of email address content:

```
Damien Dupond <ddupond@voila.fr>
Michel Delpech <mdelpech@voilaa.fr>, n.mouskouri@goodmusic.io
```

When tracim user validates "the share", the backend should create a `content_share` entity related to the content:
    - relationship is based on the `content_id`
    - one share by email address
    - send a notification to each recipient

Note: the share should have the following properties:

- `created` datetime property (default value is now())
- `disabled` datetime property (default value is null)
- `enabled` boolean property (default value is True)

The shares must appear in the timeline like: `file shared with XXX` (where XXX is the email or the associated name). If the share was for several users, then the message should be `file shared with 10 users` (and the overlay should be the list of users)

When the user decides to drop a share (for 1 user at a time), the backend should set the given share to disabled and set the disabled datetime. Then, in timeline, the disable of the share should also appear like `stop sharing of the file with XXX`.

Note : the share table may be a sub-content of given content where created/disabled would be based on the statut of the share object.

For each share, a token is created (so if you share with 3 users, there are 3 share objects, each with a dedicated token), allowing to disable a share for one user at a time and allowing to protect access according to token+password.

The notification email sent to the user must be translatable (and translated), and the content would be:

```
Michel Dupont shared a file with you: [cv.pdf]({user_friendly_link})

<p>Note: you can also use the direct link: {direct_link})</p>
<p>${_("Enjoy your day :)")}</p>
<p>${_("Suricat', your digital assistant")}</p>

<pre>
--
${_("Suricat', the bot")}
${config.WEBSITE__TITLE}
<a href="${config.WEBSITE__BASE_URL}">${config.WEBSITE__BASE_URL}</a>
${_("powered by tracim software")} - <a href="https://www.tracim.fr">https://www.tracim.fr</a>
</pre>
```

Note : public share APIs are:

- private and related to content APIs for configuration stuff
- public and in another set of routes for getting content

lebouquetin:
Screenshot of the share list: https://algoo.tracim.fr/ui/workspaces/93/contents/file/10354

inkhey:
[...]

PR #2147, opened 2019-08-05
Feature/2125  public share of a file

inkhey:
related to #2125 

## Checkpoints

- [x] Code is clear enough


backend/development.ini.sample
```diff
@@ -286,11 +286,15 @@ email.notification.reply_to.email = reply+{content_id}@trac.im
 email.notification.references.email = thread+{content_id}@trac.im
 email.notification.content_update.subject = [{website_title}] [{workspace_label}] {content_label} ({content_status_label})
 email.notification.created_account.subject = [{website_title}] Someone created an account for you
+email.notification.share_content_to_emitter.subject = [{website_title}] you shared {content_filename} to {nb_receivers} users
```
buxx: That's for an email who inform you about your share with other persons ?

inkhey: yes.

backend/tracim_backend/applications/share/__init__.py
```diff
@@ -0,0 +1,4 @@
+from tracim_backend.applications.share.application import get_app  # noqa: F401
+from tracim_backend.applications.share.config import check_config  # noqa: F401
+from tracim_backend.applications.share.config import load_config  # noqa: F401
+from tracim_backend.applications.share.controller import import_controller  # noqa: F401
```
buxx: See with @lebouquetin about use `__ini__`files. Import proposition to avoid them can be:

``` python
import tracim_backend.applications.upload_permissions.config as upload_permissions_config

upload_permissions_config.load_config
```

backend/tracim_backend/applications/share/application.py
```diff
@@ -0,0 +1,14 @@
+from tracim_backend.app_models.applications import Application
+from tracim_backend.config import CFG
+
+
+def get_app(app_config: CFG):
```
buxx: type

backend/tracim_backend/applications/share/config.py
```diff
@@ -0,0 +1,43 @@
...
+from tracim_backend import CFG
+from tracim_backend.exceptions import ConfigurationError
+from tracim_backend.lib.utils.translation import translator_marker as _
+
+
+def load_config(app_config: CFG) -> CFG:
+    # share content email
+    app_config.EMAIL__NOTIFICATION__SHARE_CONTENT_TO_RECEIVER__TEMPLATE__HTML = app_config.get_raw_config(
+        "email.notification.share_content_to_receiver.template.html"
+    )
```
buxx: A question about app_config: When do `app_config.get_raw_config(`: What happens if config not provided ?

If return None and it is a required config: not very stable.

inkhey: checking is done after loading. That's why there is both load_config and check_config method.
This allow to deal with some complex case where some config are depending on other and it make sense to be required only in some case.

buxx:
Correct me if i'm wrong: When user share content with a person, two mails are sent:

1. One to the targeted person
2. One to use who share the content

The second email seems not useful. it was required by #2147 ?

inkhey:
> Correct me if i'm wrong: When user share content with a person, two mails are sent:
> 
>     1. One to the targeted person
> 
>     2. One to use who share the content
> 
> 
> The second email seems not useful. it was required by #2147 ?

Yes, see https://github.com/tracim/tracim/issues/2125#issuecomment-516029735

backend/tracim_backend/applications/share/controller.py
```diff
@@ -0,0 +1,215 @@
...
+            raise ContentTypeNotAllowed()
+
+        try:
+            file = DepotManager.get().get(content_share.content.depot_file)
+        except IOError as exc:
+            raise TracimFileNotFound(
+                "file related to revision {} of content {} not found in depot.".format(
+                    content_share.content.revision_id, content_share.content.content_id
+                )
+            ) from exc
```
buxx: Why do you manage case where file is not present on disk ? If we really want to mange this case, it is a global case and code should be moved otherwise - i think.

backend/tracim_backend/applications/share/controller.py
```diff
@@ -0,0 +1,215 @@
...
+        filename = hapic_data.path.filename
+        if not filename or filename == "raw":
```
buxx: What is this case where filename is equal to "raw" ? (constant seem more adapted here)

backend/tracim_backend/applications/share/controller.py
inkhey: We should not allow to run code which will run template which just does not exist.
We do the same check for other template too in config.

backend/tracim_backend/applications/share/controller.py
inkhey: 1. filename come from path. We can expect any value. In case we have raw or nothing, we do return a file with same filename as original file. In other case, we override this.
2. We do exactly the same for other get file endpoint

backend/tracim_backend/applications/share/__init__.py
inkhey: changed.

backend/tracim_backend/applications/share/application.py
inkhey: done.

backend/tracim_backend/applications/share/lib.py
```diff
@@ -0,0 +1,172 @@
...
+                type=ContentShareType.EMAIL,
+                created=created,
+                share_group_id=share_group_id,
+                enabled=True,
+            )
+            self.save(content_share)
+            content_shares.append(content_share)
+            self._session.flush()
+
+        if do_notify and self._config.EMAIL__NOTIFICATION__ACTIVATED:
```
buxx: Maybe an error when `do_notify` is True and `self._config.EMAIL__NOTIFICATION__ACTIVATED` is False ?

backend/tracim_backend/applications/share/lib.py
```diff
@@ -0,0 +1,172 @@
...
+                        email=self._user.email
+                    ),
+                )
+            except SMTPException as exc:
+                raise NotificationSendingFailed(
+                    "Notification for shared file can't be send " "(SMTP error)."
+                ) from exc
+        return content_shares
+
+    def _get_email_manager(self, config: CFG, session: Session):
```
buxx: type

backend/tracim_backend/applications/share/lib.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def get_content_shares_in_context(
+        self, content_shares: typing.List[ContentShare]
+    ) -> typing.List[ContentShareInContext]:
+        content_shares_in_context = []
+        for content_share in content_shares:
+            content_shares_in_context.append(self.get_content_share_in_context(content_share))
+        return content_shares_in_context
+
+    def get_content_share_by_token(self, share_token: str):
```
buxx: type

backend/tracim_backend/applications/share/lib.py
```diff
@@ -0,0 +1,172 @@
...
+        return (
+            self._session.query(ContentShare).filter(ContentShare.share_token == share_token).one()
+        )
+
+    def check_password(self, content_share: ContentShare, password: typing.Optional[str]) -> None:
```
buxx: Add docstring with which exception is raised

backend/tracim_backend/applications/share/models.py
```diff
@@ -0,0 +1,111 @@
...
+    content_id = Column(Integer, ForeignKey("content.id"), nullable=False)
+    author_id = Column(Integer, ForeignKey("users.user_id"), nullable=False)
+    email = Column(Unicode(MAX_EMAIL_LENGTH), nullable=False)
+    share_token = Column(Unicode(MAX_SHARE_TOKEN_LENGTH), nullable=False)
+    share_group_id = Column(Unicode(MAX_GROUP_SHARE_ID_LENGTH), nullable=False)
+    type = Column(Enum(ContentShareType), nullable=False)
+    _password = Column("password", Unicode(MAX_HASHED_PASSWORD_LENGTH), nullable=True)
+    enabled = Column(Boolean, unique=False, nullable=False, default=True)
+    created = Column(DateTime, unique=False, nullable=False, default=datetime.utcnow)
+    disabled = Column(DateTime, unique=False, nullable=True, default=None)
```
buxx: This field could be renamed `disabled_datetime`

backend/tracim_backend/applications/share/lib.py
```diff
@@ -0,0 +1,172 @@
...
+    def share_content(
+        self,
+        content: Content,
+        emails: typing.List[str],
+        password: typing.Optional[str] = None,
+        do_notify=True,
+    ) -> typing.List[ContentShare]:
+        content_shares = []
+        created = datetime.now()
+        share_group_id = str(uuid.uuid4())
```
buxx: To be sure to avoid strand string representation of uuid4, use `uuid.uuid4().hex`

backend/tracim_backend/applications/share/models.py
```diff
@@ -0,0 +1,111 @@
...
+    content = relationship("Content", remote_side=[Content.id], backref="shares")
+    author = relationship("User", remote_side=[User.user_id])
+
+    @classmethod
+    def _hash(cls, cleartext_password_or_token: str) -> str:
```
buxx: Please add a docstring about what this method do

backend/tracim_backend/applications/share/models.py
```diff
@@ -0,0 +1,111 @@
...
+            self._password = None
+        else:
+            self._password = self._hash(cleartext_password)
+
+    def _get_password(self) -> str:
+        """Return the hashed version of the password."""
+        return self._password
+
+    @classmethod
+    def _validate_hash(cls, hashed: str, cleartext_password_or_token: str) -> bool:
```
buxx: This method deal only with password hashs ? If yes could be renamed `_validate_password_hash`

inkhey: this method is not password specific, it's similar code than for tracim core lib.
We should maybe put this one and _hash in another place like utils but this mean modifying core code (which is not the goal of this PR).

backend/tracim_backend/applications/share/lib.py
inkhey: done.

backend/tracim_backend/applications/share/models.py
inkhey: done.

backend/tracim_backend/applications/share/lib.py
inkhey: done.

backend/tracim_backend/applications/share/lib.py
inkhey: done.

backend/tracim_backend/applications/share/lib.py
inkhey: done.

backend/tracim_backend/config.py
```diff
@@ -147,6 +147,9 @@ def load_config(self) -> None:
         self._load_caldav_config()
         self._load_search_config()
         self._load_collaborative_document_edition_config()
+        import tracim_backend.applications.share.config as share_app_config
```
buxx: Cannot be imported on top of file ? If no, think to add a NOTE about cyclic import problem.

backend/tracim_backend/config.py
```diff
@@ -569,6 +573,9 @@ def check_config_validity(self) -> None:
         self._check_caldav_config_validity()
         self._check_search_config_validity()
         self._check_collaborative_document_edition_config_validity()
+        import tracim_backend.applications.share.config as share_app_config
```
buxx: idem

backend/tracim_backend/config.py
```diff
@@ -839,7 +846,9 @@ def _set_default_app(self, enabled_app_list: typing.List[str]) -> None:
             creation_label="Create an office document",
             available_statuses=content_status_list.get_all(),
         )
+        import tracim_backend.applications.share.application as share_app
```
buxx: idem

backend/tracim_backend/config.py
inkhey: currently, it may work (need confirmation), but as those are import for specific app. import should be later done dynamically according to enabled app.
I prefer to have very clear point of connection between app and config to make thing easy when we will be able to make thing pluggable.

backend/tracim_backend/tests/utils.py
```diff
@@ -55,6 +56,20 @@ def get(
         )
 
 
+class ShareApiFactory(object):
+    def __init__(self, session: Session, app_config: CFG, admin_user: User):
+        self.session = session
+        self.app_config = app_config
+        self.admin_user = admin_user
+
+    def get(self, current_user: typing.Optional[User] = None) -> ShareLib:
```
buxx: `create` is more appropriate for creation. `get` is more oriented "return something already existing".

backend/development.ini.sample
buxx: ok

buxx:
> > Correct me if i'm wrong: When user share content with a person, two mails are sent:
> > ```
> > 1. One to the targeted person
> > 
> > 2. One to use who share the content
> > ```
> > 
> > 
> > The second email seems not useful. it was required by #2147 ?
> 
> @buxx Yes,required, see [#2125 (comment)](https://github.com/tracim/tracim/issues/2125#issuecomment-516029735)

ok

backend/tracim_backend/tests/utils.py
inkhey: we can change this but get is method used for other factories. Changing all of them is a big refactor (but maybe not too difficult), changing only one make thing inconsistent.

backend/tracim_backend/applications/share/controller.py
buxx: I don't understand why you're speaking about "template". This code deal with content's file, not template. Or i miss something ...

About my original comment: What i'm saying is: A controller views should not have to manage the case "depot file cannot be retrieved because there is corrupted data". Only `@hapic.handle_exception(ParentErrorOfDepotProblems)` is enough.

backend/tracim_backend/tests/utils.py
inkhey: get and create usage is clear for normal object but here idea behind this is more "i want a valid lib, with this configuration, i don't mind about which is done behind".

backend/tracim_backend/applications/share/controller.py
inkhey: I clearly don't understand the point.

backend/tracim_backend/applications/share/models.py
buxx: Okay, can you open an technical dept issue please ?

backend/tracim_backend/config.py
buxx: That's a good reason. Can you add it as a note ?

backend/tracim_backend/applications/share/controller.py
inkhey: Oh, sorry. it remind me another piece of code.
There is currently not such thing about depot in ContentLib, not sure how we should handle this.

buxx: I'm considering this controller view should not manage this case because the case where binary file is not on disk should not happen because:

* When File `Content` type  do not own file: we should not be able to share it
* When File `Content` type has been deleted: shares are they deleted ? They must.

But, as you say the case "binary file can be absent on disk when deal with DepotManager" is a general case on Tracim. So i suggest opening a technical debt issue to think about all of that.

backend/tracim_backend/applications/share/controller.py
buxx: The point of the technical debt issue is: for example,as new developer, when you read this code you d'ont understand what it do. So, refactor in func with docstring, or use explicit named constant, etc.

backend/tracim_backend/tests/utils.py
buxx: see this later

inkhey: https://github.com/tracim/tracim/issues/2169

backend/tracim_backend/applications/share/controller.py
inkhey: ticket about moving this code here : https://github.com/tracim/tracim/issues/2168

backend/tracim_backend/applications/share/models.py
inkhey: ticket here: https://github.com/tracim/tracim/issues/2170

backend/tracim_backend/config.py

backend/tracim_backend/config.py

backend/tracim_backend/config.py

backend/tracim_backend/applications/share/lib.py
inkhey: fixed, do_notify get  self._config.EMAIL__NOTIFICATION__ACTIVATED value.

backend/tracim_backend/applications/share/controller.py
inkhey: Explained in comment.

backend/tracim_backend/applications/share/models.py
inkhey: ticket here : https://github.com/tracim/tracim/issues/2171

Issue #2031, opened 2019-06-28
add possibility to limit number of sharedspace available for each user


PhilippeAccorsi:
## Feature description

- We need to define how many sharedspace each user can create.
- at the moment we define in the config file the number of sharedspaces peruser
    - 0 means no mlimit

note: do not count deleted sharedspaces

Informations to add in the APIs:

- add max_sharedspace_per_user in config endpoint
- add owned_sharedspaces in user sharedspaces endpoint + query filters that allow to choose to get owned sharedspaces or sharedspaces with a role or all sharedspaces (by default)
- add owner in sharedspace serialisation

## Context

Need this feature for hosting.

## Impacted components

- *backend*
- *frontend*

inkhey:
This probably means that we should know who is the creator of one workspace, this information is not stored yet.

We should find a solution to make this work also in context of migration of database. 
What should be the behavior for workspace that already exist ?

inkhey:
this look depending on #2027

inkhey:
begin a fix here : https://github.com/tracim/tracim/tree/fix/2031__limit_number_of_workspaces_allowed_for_user

This code, does a default limit of 100 workspace per user and return error when creating new workspace over this limit. The limit can be setted only by database.

We do need to explicit : 
- how should we change this value (specific endpoint available only for admin, something else ?)
- Which value to use when creating/inviting a new account ?
- if we handle "unlimited case", this can be tricky as we cannot use none which can already mean "no change". current solution is consider that 100+ value is ok for most case so no need for unlimited if maximum value is easy to set.

PhilippeAccorsi:
Description edited this morning with new information.

inkhey:
not sure about api : 
current filter on workspace user are : 
"show_workspace_with_role" and  "show_owned_workspace" both as query bool (0 or 1 value) parameter.

inkhey:
@gbonaspetti , @basilelegal , with the new spec here. it mean 
`/api/v2/users/me/workspaces` and `/api/v2/users/{user_id}/workspaces` return not same default value as before. You should set `show_owned_workspace` query param to 0 to retrieve legacy behavior. 
Don't know how important this change may affect frontend.

inkhey:
after discussion avec @Skylsmoi, seems that's not a big refactor to make thing works correctly in frontend.

so all previous call to /api/v2/users/me/workspaces and /api/v2/users/{user_id}/workspaces should be changed to add query param show_owned_workspace to 0.

[...]

PR #2257, opened 2019-08-22
Fix/2031  limit number of sharespaces

inkhey:
related to #2031 

## Checkpoints

- [x] Code is clear enough


backend/tracim_backend/lib/agenda/agenda.py
```diff
@@ -193,7 +193,7 @@ def get_user_agendas(
             workspace_api = WorkspaceApi(
                 current_user=self._user, session=self._session, config=self._config
             )
-            workspaces = workspace_api.get_all_for_user(user)
+            workspaces = workspace_api.get_all_for_user(user, with_role=True, owned=False)
```
buxx: Why exclude owned workspaces here ?

inkhey: This code excludes only workspace where user is owner and has no role.
We do not expect user without any role to access to agenda feature. This change is just to obtain same behavior as before.

buxx: Okay I got it. `with_role` and `owned` permit to include them. If both are set to False, list is always empty.
Can you rename into `include_with_role` and `include_owned` and add a docstring to `get_all_for_user` please ?

backend/tracim_backend/lib/core/workspace.py
```diff
@@ -83,6 +84,11 @@ def create_workspace(
         agenda_enabled: bool = True,
         save_now: bool = False,
     ) -> Workspace:
+        from tracim_backend.lib.core.user import UserApi
```
buxx: When import is not at top of file for a specific reason, it is preferable to indicate why in a note.

backend/tracim_backend/lib/core/workspace.py
```diff
@@ -178,20 +184,35 @@ def get_one_for_current_user(self, id):
     def get_all(self):
         return self._base_query().all()
 
-    def get_all_for_user(self, user: User, ignored_ids=None):
-        workspaces = []
+    def _get_workspace_owned_by_user(self, user_id: int) -> typing.List[Workspace]:
```
buxx: Add a `s` to `workspace` ?

backend/tracim_backend/lib/core/workspace.py
```diff
@@ -178,20 +184,35 @@ def get_one_for_current_user(self, id):
...
+        return self._base_query_without_roles().filter(Workspace.owner_id == user_id).all()
+
+    def _get_workspace_ids(self, workspaces: typing.List[Workspace]) -> typing.List[int]:
+        workspace_ids = []
+        for workspace in workspaces:
+            workspace_ids.append(workspace.workspace_id)
+        return workspace_ids
```
buxx: I'm not sure specific method is useful for that. @lebouquetin What is your opinion about use short code like following instead a real function ?

```python
[w.id for w in workspaces]
```

backend/tracim_backend/lib/webdav/resources.py
```diff
@@ -176,7 +179,14 @@ def createCollection(self, name: str):
...
         workspace_name = webdav_convert_file_name_to_bdd(name)
-        new_workspace = self.workspace_api.create_workspace(workspace_name)
+        try:
+            new_workspace = self.workspace_api.create_workspace(workspace_name)
+        except (
+            UserNotAllowedToCreateMoreWorkspace,
+            EmptyLabelNotAllowed,
+            WorkspaceLabelAlreadyUsed,
+        ):
+            raise DAVError(HTTP_FORBIDDEN)
```
buxx: `DAVError` class own `contextinfo` (`str`) parameter in constructor: we can use it to add some information about error (like `DAVError(HTTP_FORBIDDEN, contextinfo=str(exc)`).

inkhey: Added this to all webdav case Forbidden error where it's possible to have explicit error.

inkhey: This is a nice feature for debugging, but those informations are not provided by webdav client like nautilus (don't know if some client does give thoses informations).

backend/tracim_backend/lib/core/workspace.py

backend/tracim_backend/lib/core/workspace.py
inkhey: explained.

backend/tracim_backend/lib/agenda/agenda.py
inkhey: done.

backend/tracim_backend/lib/core/workspace.py
lebouquetin: @buxx @inkhey this way of writing code may be complicated to understand... when it is combined with conditionnal options and sub-inlining operations.

The example you write is ok: the code is clear and easy to read and understand

inkhey: fixed.

Issue #2299, opened 2019-08-27
error 403 appear when reader open file


PhilippeAccorsi:
## Bug description & symptom

When reader open file, an error 403 appear :
~~~
Warning
Error while loading share links list
~~~

error code 4003 : `InsufficientUserRoleInWorkspace`

## Diagnostic

*"unknown".*

## Expected

- [ ] reader user can use tracim without error about shared file
- [ ] reader not need to know everything about shared feature
- [ ] contributor can only read share links

## How to reproduce

- add reader user in sharedspace
- open one file
=> error appear directly

## Screenshots

This is what reader see at present when file is shared:
![Sélection_025](https://user-images.githubusercontent.com/25848257/63759257-78a58b80-c8bd-11e9-89c8-931ae4739e9d.jpg)

## Impacted components

*"unknown"*

## Extra information

PhilippeAccorsi:
Verified.

PR #2315, opened 2019-08-28
[#2299] Fix share file feature visiblity according to member role's

metzgegu:
#2299 

## Checkpoints

- [x] Code is clear enough


frontend_app_file/src/container/File.jsx
```diff
@@ -905,7 +907,7 @@ class File extends React.Component {
                   shouldScrollToBottom={state.mode !== MODE.REVISION}
                 />
               },
-              {
+              ...(state.loggedUser.userRoleIdInWorkspace > 1 ? [{
```
Skylsmoi: to be standard valid, you should write it that way:
``` jsx
...(state.loggedUser.userRoleIdInWorkspace > 1
  ? [{
    ...
  }]
  : []
)
```

Issue #2028, opened 2019-06-28
add possibility to specify limit for file what you want to upload in tracim


PhilippeAccorsi:
## Feature description

Limit size of file its possible to upload in tracim

- limit is for file size (not for html page for example)
- check size on frontend size and on backend size
- use content-length for check
- add documentation for apache configuration
- add max file size in config endpoint.

## Context

Need this feature for hosting.

## Impacted components

- *backend*
- *frontend*

UPDATE 02:09.2019 :warning:  Only backend part of the feature is done in this issue for apache doc/frontend part see new issue #2345

inkhey:
For frontend: 
It depend on what frontend can do, but frontend seems the best place to disallow sending big file without having to wait for error.

For Backend: 
I'm not sure that's really interesting to make limit here in api, it's already possible to limit size of body request send and this return, as far i know, specific error code. Does it make sense to handle case where webserver doesn't failed but return error in backend in those situation ?

inkhey:
https://developer.mozilla.org/fr/docs/Web/HTTP/Status/413

inkhey:
If frontend is able to make limitation, we should maybe return size limitation information with a backend endpoint, in this case it make sense to technically refuse too big file in backend although it was already received.

inkhey:
try limitation in backend according to content_length value, work for both webdav and api context.
see https://github.com/tracim/tracim/tree/fix/2028__limit_size_of_file_upload

PhilippeAccorsi:
Description edited this morning.

inkhey:
Wip branch here : https://github.com/tracim/tracim/tree/feature/2028__limit_size_of_file_upload

inkhey:
@PhilippeAccorsi 
- Backend part is done in : https://github.com/tracim/tracim/tree/feature/2028__limit_size_of_file_upload
- Frontend part can be done with value given in /config endpoint (in backend part)
- Documentation for apache is unclear ?

inkhey:
frontend should :
- do not allow upload file with more than size limit given by content_length_file_size_limit in content endpoint.
- support specific error code related to size limit ? (should not occured at all because of first check in /config )

PhilippeAccorsi:
> @PhilippeAccorsi
> 
>     * Backend part is done in : https://github.com/tracim/tracim/tree/feature/2028__limit_size_of_file_upload
> 
>     * Frontend part can be done with value given in /config endpoint (in backend part)
> 
[...]

PR #2331, opened 2019-08-29
Feature/2028  limit size of file upload

inkhey:
related to #2028 
require #2031.

Only backend part of #2028 

## Checkpoints

- [x] Code is clear enough


inkhey:
in draft PR because waiting for #2031 merge.

buxx:
Why do not call check file size inside ContentApi instead from http controllers ?

inkhey:
> Why do not call check file size inside ContentApi instead from http controllers ?

Because only way to do this properly (refuse too big file) is by checking content_length header. Another solution may be to make content_length as new param of update/create content method, not sure it's a good idea.

Issue #2351, opened 2019-09-03
Explicit error in case of no file given when uploading file.


inkhey:
## Bug description & symptom

In endpoint where we use input files, we do not correctly handle the "required" case.
Best support with hapic should be done later, see https://github.com/tracim/tracim/issues/2350

## Expected

Support theses case with 400 error and specific error message.

PhilippeAccorsi:
Verified.

PR #2353, opened 2019-09-03
handle no file given case validation error

inkhey:
related to #2351

## Checkpoints

- [x] Code is clear enough


backend/tracim_backend/tests/functional/test_contents.py
```diff
@@ -2321,6 +2321,23 @@ def test_api__create_file__err_400__filename_already_used(
         assert "code" in res.json.keys()
         assert res.json_body["code"] == ErrorCode.CONTENT_FILENAME_ALREADY_USED_IN_FOLDER
 
+    def test_api__create_file__err_400__no_file_given(
+        self, workspace_api_factory, content_api_factory, session, web_testapp
+    ) -> None:
+        """
+        create one file of a content but filename is already used here
```
inkhey:
```suggestion
        create one file of a content but no input file given
```

backend/tracim_backend/models/context_models.py
```diff
@@ -140,7 +140,7 @@ def __init__(self, loggedin_user_password: str, email: str) -> None:
 
 
 class SimpleFile(object):
-    def __init__(self, files: cgi.FieldStorage) -> None:
+    def __init__(self, files: cgi.FieldStorage = None) -> None:
```
buxx: Why this diff ?

inkhey: files param is not mandatory here, but is mandatory later in code, None case allow use to know there. if we do not set None by default here, code will raise 500 here instead of raising 400 error 2005 later.

buxx: ok

Issue #2356, opened 2019-09-03
The semicolon does not work as a email separator


gbonaspetti:
## Bug description & symptom

For sharing creation, the semicolon does not work as a email separator.

## Expected

Add new separator ;

PhilippeAccorsi:
Semicolon already exist in code but not explicit in frontend. Fix with this commit https://github.com/tracim/tracim/commit/dcde97ab77919536fe0289f5ac57d379095ea469

PhilippeAccorsi:
Verified.

PR #2357, opened 2019-09-03
[#2356] add semicolon as separator

gbonaspetti:
#2356 

## Checkpoints

- [x] Code is clear enough


lebouquetin:
@Skylsmoi what is the interest to write the function without to use variable?

You mean write the code like this?

```
  return string.split(' ').join(',').split('\n').join(',').split(';').join(',').split(',').filter(notEmptyString => notEmptyString !== '')
```

The right way to implement this should be:

```
function stringToList(separators=",\n;") {
  // split string according to separators given as parameter
}
```

This way to write the code is "open to extension, closed to modification"

Skylsmoi:
That's what I was expecting.
The point is to get used to function chaining

frontend_lib/src/helper.js
```diff
@@ -155,11 +155,17 @@ export const displayFileSize = (bytes, decimals) => {
   return parseFloat((bytes / Math.pow(k, i)).toFixed(dm)) + ' ' + sizes[i]
 }
 
-// INFO - GB - 2019-09-03 - In this function each split is one separator that we change for a comma
-export const parserStringToList = string => {
-  let list = string.split(' ').join(',').split('\n').join(',').split(';').join(',').split(',')
-  list = list.filter(notEmptyString => notEmptyString !== '')
-  return list
+export const parserStringToList = (string, separatorList) => {
```
Skylsmoi: as @lebouquetin proposed, you should give a default value to parameter separatorList.
That way, and since it'll always be the same value for separatorList, it'll make the function call simpler by not having to give the that parameter.

frontend_lib/src/helper.js
```diff
@@ -155,11 +155,17 @@ export const displayFileSize = (bytes, decimals) => {
...
+  let mainSeparator = ''
+  separatorList.forEach((separator, index) => {
+    if(index === 0){
```
Skylsmoi: I don't think you have to use this, you can force the mainSeparator to, let's say, a comma and be fine with it

frontend_lib/src/helper.js
```diff
@@ -155,11 +155,17 @@ export const displayFileSize = (bytes, decimals) => {
...
+      mainSeparator = separator
+    }
+    else {
+      string = string.split(separator).join(mainSeparator)
```
Skylsmoi: don't modify the parameter of a function, create a new variable

Issue #2366, opened 2019-09-04
add upload message to upload notification email


inkhey:
## Bug description & symptom

Currently, upload message doesn't return message of the uploader in email received by workspaces users.

## Expected

- [ ] add uploader message in mail template after files list.
- [ ] translation working
- [ ] if  line start with word "Message:", in case are without message need to be explicit

PhilippeAccorsi:
Edited issue to add what is expected.

PhilippeAccorsi:
Verified.

PR #2378, opened 2019-09-06
Fix/2366  message of uploader in email notification

inkhey:
related to #2366.

## Checkpoints

- [x] Code is clear enough


backend/tracim_backend/applications/upload_permissions/lib.py
```diff
@@ -293,6 +295,7 @@ def upload_files(
                 workspace_in_context=workspace_lib.get_workspace_with_context(
                     upload_permission.workspace
                 ),
+                uploader_message=message,
```
buxx: There is a problem about `message` variable:

In http views, this is a non required parameter:

``` python
class UploadDataFormSchema(marshmallow.Schema):
    message = StrippedString(required=False, allow_none=False)
    # ...
```

But as start of `upload_files` method, this is a required parameter:

``` python
    def upload_files(
        self,
        upload_permission: UploadPermission,
        uploader_username: str,
        message: str,
        # ...
```

So, as start of `upload_files` method, `message` variable is not tested as an Optional type. 

Maybe you have just ti change typing. If not, create a technical debt issue is to long to modify before 2.4.0.

inkhey: you 're right message typing is uncorrect.
It's just fixed.

buxx: perfect

Issue #2395, opened 2019-09-11
Space between the buttons is not the same between apps


gbonaspetti:
## Bug description & symptom

The space between the status button and the delete button (trash can) is not the same between the file app and the other 2 apps (html doc and discussion) 

## Expected

All have the same interface.

PR #2396, opened 2019-09-11
[#2395] fix status button to have the same margin for all apps

gbonaspetti:
#2395 

## Checkpoints

- [ ] Code is clear enough

### DIFF
--- a/frontend_lib/src/component/Input/SelectStatus/SelectStatus.styl
+++ b/frontend_lib/src/component/Input/SelectStatus/SelectStatus.styl
@@ -1,6 +1,7 @@
 @import '../../../css/Variable.styl'
 
 .selectStatus
+  margin-right 15px
   &__dropdownbtn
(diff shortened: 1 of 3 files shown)


frontend_lib/src/component/Input/SelectStatus/SelectStatus.styl
```diff
@@ -1,6 +1,7 @@
 @import '../../../css/Variable.styl'
 
 .selectStatus
+  margin-right 15px
```
Skylsmoi: the positioning of a generic component shouldn't be in the component itself.

The reason is that if we use this component somewhere where we must not have a margin-right: 15px, we will have to either add a props to the component to make this margin conditional or give the wrapper a negative margin to cancel it.

So, here, it's better to make each usage of the component a margin-right.

in conclusion: revert your change and add the missing margin-right to the apps that needs it

Issue #2474, opened 2019-10-07
add possibility to configure default profile of new users


PhilippeAccorsi:
## Feature description

Currently, when creating a new user through invitation process, the profile of the user is "normal user". For public instance, it should be "trusted user" in order invited people to be allowed to create shared spaces.

The most flexible way to allow this is to add a default profile parameter in configuration file.

## Impacted components

- *backend*

## Extra information

- The value to set in developement.ini.template should be "standard user".
- Note : the config parameter should be an explicit string, not an integer value (which is not auto-explicit)

PhilippeAccorsi:
Tested with development.ini and with env variable.
Verified.

PR #2478, opened 2019-10-08
Fix/2474  allow configure default profiles of new users

inkhey:
related to #2474 

this does:
- make default profile for user creation and user invitation as config parameter
- refactor profile code to have a more clear code

## Checkpoints

- [x] Code is clear enough


backend/tracim_backend/models/auth.py
```diff
@@ -87,14 +88,45 @@ class AuthType(enum.Enum):
...
+        if len(profiles) != 1:
+            raise ProfileDoesNotExist()
+        return profiles[0]
+
+    @classmethod
+    def get_profile_from_slug(self, slug: str):
+        profiles = [item for item in list(Profile) if item.slug == slug]
+        if len(profiles) != 1:
+            raise ProfileDoesNotExist()
+        return profiles[0]
```
buxx: For these 3 class methods:

* thirst parameter name must be `cls` and not `self` (PyCharm should underline it in yellow)
* You can replace `list(Profile)` by `list(cls)` for better maintainability

inkhey: done.

Issue #2475, opened 2019-10-07
add possibility to auto-complete all users of instance


PhilippeAccorsi:
## Feature description

At present when you try to add member in a shared space, in autocomplete you only seeusers you already know.
 
We need a parameter in config file to activate access toall users of the instance. If this parameter is set to "True", autocomplete may return any user found in the instance.
If parameter set to "False", autocomplete should work like it does currently (only show users you already know).

## Impacted components

- *backend*

PhilippeAccorsi:
This commit content:
- add feature to disable filter on known_members endpoint
- add limit of user returned to 5 in known_members endpoint.

PhilippeAccorsi:
verified.

PhilippeAccorsi:
If tracim user have profile "user" and role "manager", this tracim user can't see autocomplete in add member list if parameter `known_members.filter` is set to "true" (default value). 
If parameter is set to false, this tracim user can see autocomplete in add member list.

PhilippeAccorsi:
Verified.

PR #2479, opened 2019-10-08
Fix/2475 mode view all instance members on known members endpoint

inkhey:
related to  #2475 

this contains:
- add feature to disable filter on known_members endpoint
- add limit of user returned to 5 in known_members endpoint.

## Checkpoints

- [x] Code is clear enough


backend/tracim_backend/lib/core/user.py
```diff
@@ -187,7 +194,7 @@ def get_known_user(
         )
         # INFO - G.M - 2018-07-27 - if user is set and is simple user, we
         # should show only user in same workspace as user
-        if self._user and self._user.profile.id <= Group.TIM_USER:
+        if filter_results and self._user and self._user.profile.id <= Group.TIM_USER:
```
buxx: We should raise an exception if `filter_results and not self._user`

buxx: Second point about `and self._user.profile.id <= Group.TIM_USER`: i'm not sure if `get_known_user` should continue to use this limitation because now, this is  ̀filter_results` which indicate to filter or not. If level of user must be check, it should be the responsibility of caller.

inkhey: > Second point about `and self._user.profile.id <= Group.TIM_USER`: i'm not sure if `get_known_user` should continue to use this limitation because now, this is ̀filter_results` which indicate to filter or not. If level of user must be check, it should be the responsibility of caller.

Doing this meaning add logic to controller, not sure that a good thing, we do have 4 case:
- filter_config = False and profile == 'user' -> no filter
-  filter_config = False and profile > 'user' -> no filter
-  filter_config = True and profile == 'user' -> apply filter
-  filter_config = True and profile > 'user' -> apply filter.

Issue #2524, opened 2019-11-25
Add mecanism to add custom javascript/app code in tracim


inkhey:
## Feature description

There is currently no easy way to add custom JavaScript code in Tracim. Idea of this feature is to add configuration of folder where user can put JavaScript code which will be loaded in Tracim like
normal application.

PhilippeAccorsi:
Verified.

PR #2525, opened 2019-11-25
[for 2.5.1] add mecanism to inject easily javascript code in existing tracim

inkhey:
related to #2524 

## Checkpoints

- [x] Code is clear enough

### DIFF
--- a/backend/tracim_backend/views/frontend.py
+++ b/backend/tracim_backend/views/frontend.py
@@ -18,7 +18,9 @@ APP_FRONTEND_PATH = "app/{minislug}.app.js"
 
 
 class FrontendController(Controller):
-    def __init__(self, dist_folder_path: str, custom_toolbox_folder_path: typing.Optional[str]) -> None:
+    def __init__(
+        self, dist_folder_path: str, custom_toolbox_folder_path: typing.Optional[str]
+    ) -> None:
         self.dist_folder_path = dist_folder_path
         self.custom_toolbox_folder_path = custom_toolbox_folder_path
         self.custom_toolbox_files = []  # type: typing.List["os.DirEntry"]
--- a/backend/tracim_backend/views/frontend.py
+++ b/backend/tracim_backend/views/frontend.py
@@ -82,12 +82,12 @@ class FrontendController(Controller):
         configurator.add_route("index", INDEX_PAGE_NAME, request_method="GET")
         configurator.add_view(self.index, route_name="index")
 
+        if self.custom_toolbox_folder_path:
+            configurator.add_static_view(
+                name="assets/custom_toolbox", path=self.custom_toolbox_folder_path
+            )
+
         for dirname in os.listdir(self.dist_folder_path):
             configurator.add_static_view(
                 name=dirname, path=os.path.join(self.dist_folder_path, dirname)
-            )
-
-        if self.custom_toolbox_folder_path:
-            configurator.add_static_view(
-                name="custom_toolbox", path=self.custom_toolbox_folder_path
             )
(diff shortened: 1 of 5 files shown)


inkhey:
> You named the feature "custom_app" in the backend code.
> If the frontend, on my branch, I have named it "instanceSpecificCode".
> Theses are too different, we need to harmonize it.
> I don't like "custom_app" because it is too generic, it doesn't differentiate this "app" from tracim's app.
> I used the name "code" because it doesn't have to be an app.
> 
> I think we should call this feature "instanceSpecificScript". I'm open to suggestions.
> 
> Note: I will do possible required frontend changes
> 
> Note2: Someone still needs to do a code quality review.

If we do instance_specific_script, what should be the path for file /instance_specific_script/ is ok for you ? Current code allow to make available all content in (renamed here) `FRONTEND__INSTANCE_SPECIFIC_SCRIPT_FOLDER_PATH` parameter in  /instance_specific_script/myfile but only .js are added to index.html as script.
-

backend/tracim_backend/views/frontend.py
```diff
@@ -17,15 +18,29 @@
 
 
 class FrontendController(Controller):
-    def __init__(self, dist_folder_path: str):
+    def __init__(self, dist_folder_path: str, custom_toolbox_folder_path: typing.Optional[str]):
```
buxx:
```suggestion
    def __init__(self, dist_folder_path: str, custom_toolbox_folder_path: typing.Optional[str]) -> None:
```

frontend/dist/index.mak
```diff
@@ -92,6 +92,10 @@
     <script type='text/javascript' src='/app/admin_workspace_user.app.js'></script>
     <script type='text/javascript' src='/app/workspace_advanced.app.js'></script>
 
+    % for custom_toolbox_file in custom_toolbox_files:
+    <script type='text/javascript' src='/custom_toolbox/${custom_toolbox_file.name}'></script>
```
Skylsmoi: add `/asset` before `/custom_toolbox`

backend/tracim_backend/views/frontend.py
```diff
@@ -17,15 +18,29 @@
...
         self.dist_folder_path = dist_folder_path
+        self.custom_toolbox_folder_path = custom_toolbox_folder_path
+        self.custom_toolbox_files = []  # typing.List["os.DirEntry"]
```
buxx:
```suggestion
        self.custom_toolbox_files = []  # type: typing.List["os.DirEntry"]
```

```suggestion
        self.custom_toolbox_files = []  # type: typing.List[os.DirEntry]
```

backend/tracim_backend/views/frontend.py
```diff
@@ -17,15 +18,29 @@
...
+            )
 
-    def _get_index_file_path(self):
+    def _get_index_file_path(self) -> str:
         index_file_path = os.path.join(self.dist_folder_path, INDEX_PAGE_NAME)
         if not os.path.exists(index_file_path):
             raise FileNotFoundError()
         return index_file_path
 
+    def _get_custom_toolboxes_files(self, custom_toolbox_dir: str) -> typing.List["os.DirEntry"]:
```
buxx:
```suggestion
    def _get_custom_toolboxes_files(self, custom_toolbox_dir: str) -> typing.List[os.DirEntry]:
```

inkhey: Not possible here. os.DirEntry is not importable in python 3.5. So if we do this, we have to import os.DirEntry and code will failed in 3.5.
see : https://stackoverflow.com/questions/38307995/create-os-direntry

backend/tracim_backend/views/frontend.py
inkhey: same answers as latest : this fix doesn't work on python3.5.

frontend/dist/index.mak
inkhey: done. but dir is /assets/ not /asset/

Issue #2565, opened 2019-12-05
Explicit origin user in invite message when possible.


inkhey:
## Bug description & symptom

When we receive invitation in tracim, we get "Someone invited you to join tracim" message
Something like: "John invited you to join tracim" will be better;

This mean change for:
- email subject (currently `[{website_title}] Someone created an account for you`)
- email content (currently `{origin_user_name} invited you to join <i>{website_title}</i>.` )
- email FROM header (obtain "username (through tracim)" instead of "tracim notification")

As user invitation can be done in some very specific case without origin (command line, ldap, etc...), we should think about case where there is no "origin_user" and adapt subjet, content and from to these case.

PhilippeAccorsi:
Verified.

PR #2566, opened 2019-12-05
explicit origin user for email notification

inkhey:
related to #2565 
-> change template to make email about new account created more explicit about origin.

## Checkpoints

- [ ] Code is clear enough

### DIFF
--- a/backend/tracim_backend/lib/mail_notifier/notifier.py
+++ b/backend/tracim_backend/lib/mail_notifier/notifier.py
@@ -332,7 +332,9 @@ class EmailManager(object):
 
             send_email_through(self.config, email_sender.send_mail, message)
 
-    def notify_created_account(self, user: User, password: str) -> None:
+    def notify_created_account(
+        self, user: User, password: str, origin_user: typing.Optional[User] = None
+    ) -> None:
         """
         Send created account email to given user.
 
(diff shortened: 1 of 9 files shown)


backend/tracim_backend/lib/mail_notifier/notifier.py
```diff
@@ -351,12 +353,13 @@ def notify_created_account(self, user: User, password: str) -> None:
         subject = translated_subject.replace(EST.WEBSITE_TITLE, str(self.config.WEBSITE__TITLE))
         message = MIMEMultipart("alternative")
         message["Subject"] = subject
-        message["From"] = self._get_sender()
+        message["From"] = self._get_sender(origin_user)
```
lebouquetin: In origin_user, you may have the real user (with its specific email), no?

lebouquetin: The code is not safe, I think: if you call this method with a wrong user instance, you will try to send email with a from you are not authorized to.

lebouquetin: After discussion, I'm wrong: `self._get_sender(origin_user)` builds a "from" like `Damien via Tracim<notification+4@tracim.fr>`. Probably renaming this method would be a way not to misunderstand what the code does

Issue #2540, opened 2019-12-02
in app gallery the breadcrumb is not working for folder name


PhilippeAccorsi:
## Bug description & symptom

When I click on folder name visible in breadcrumb (in app gallery), tracim open list of all content bug folder is not visible (all folder are closed)

## Diagnostic

*Optional - explain what has been found and what is not yet clear. If unknown, write "unknown".*

## Expected

- [x] If I click on folder name in breadcrumb, this folder is visible directly

## Context and version information

- Desktop environment (if applicable)
    - *OS: Buster*
    - *Browser firefox*

## How to reproduce

- open app gallery from one folder
- click on folder name in breadcrumb

## Impacted components

*If unsure, write "unknown"*

PhilippeAccorsi:
Verified.

PR #2577, opened 2019-12-10
[for 2.6.0] Fix/2540  fix breadcrumbs folder link in app gallery

metzgegu:
#2540 

## Checkpoints

- [x] Code is clear enough


frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -126,23 +124,35 @@ class Gallery extends React.Component {
     document.removeEventListener(CUSTOM_EVENT.APP_CUSTOM_EVENT_LISTENER, this.customEventReducer)
   }
 
-  buildBreadcrumbs = () => {
+  async loadContentDetails () {
+    const { state } = this
+
+    const workspaceLabel = await this.loadWorkspaceData()
```
Skylsmoi: wrong naming. loadWorkspaceData should return an object with a property `label`. Or loadWorkspaceData should be called loadWorkspaceLabel()

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -154,20 +164,34 @@ class Gallery extends React.Component {
     // app crash telling it cannot render a Link outside a router
     // see https://github.com/tracim/tracim/issues/1637
     // GLOBAL_dispatchEvent({type: 'setBreadcrumbs', data: {breadcrumbs: breadcrumbsList}})
-    this.setState({ breadcrumbsList: breadcrumbsList })
+    return breadcrumbsList
   }
 
   loadFolderDetail = async (workspaceId, folderId) => {
```
Skylsmoi: you have a function loadFolderDetail and getFolderDetail, these 2 names are too ambiguous. We need better names.
From my understanding, this function will load the parents id up to the root workspace. It's name should show that.

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -154,20 +164,34 @@ class Gallery extends React.Component {
...
+          )
+          if (fetchContentDetail.apiResponse.status === 200) {
+            folderParentsId.push(fetchContentDetail.body.parent_id)
+          } else {
+            this.sendGlobalFlashMessage(props.t('Error while loading folder detail'))
+            break
+          }
+        }
+
+        return { fileName, folderParentsId }
```
Skylsmoi: if your function return an object, it must always return something predetermined. In some cases, your function will return undefined. This is highly error prone

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -154,20 +164,34 @@ class Gallery extends React.Component {
...
+        let folderParentsId = fetchContentDetail.body.parent_id ? [fetchContentDetail.body.parent_id] : []
+
+        while (fetchContentDetail.body.parent_id !== null) {
+          fetchContentDetail = await handleFetchResult(
+            await getFolderDetail(state.config.apiUrl, workspaceId, fetchContentDetail.body.parent_id)
+          )
+          if (fetchContentDetail.apiResponse.status === 200) {
+            folderParentsId.push(fetchContentDetail.body.parent_id)
+          } else {
+            this.sendGlobalFlashMessage(props.t('Error while loading folder detail'))
```
Skylsmoi: this error case can spam.
Your while variable can end up in an infinite loop.
You need to handle your iteration variable manually depending on `fetchContentDetail`. Don't depend on the result of another function. Even more so that the other function directly depends on what the api does

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -126,23 +124,35 @@ class Gallery extends React.Component {
...
-  buildBreadcrumbs = () => {
+  async loadContentDetails () {
+    const { state } = this
+
+    const workspaceLabel = await this.loadWorkspaceData()
+    let folderDetail
+    if (state.folderId) {
+      folderDetail = await this.loadFolderDetail(state.config.appConfig.workspaceId, state.folderId)
+    }
+    const breadcrumbsList = await this.buildBreadcrumbs(workspaceLabel, folderDetail, false)
```
Skylsmoi: I'm not a fan of the fact that loadContentDetails handle the breadcrumbs internally.
the breadcrumbs depends on that function but should be handled outside of it

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -164,34 +166,43 @@ class Gallery extends React.Component {
...
     )
 
     switch (fetchContentDetail.apiResponse.status) {
       case 200:
-        const fileName = fetchContentDetail.body.filename
-        let folderParentsId = fetchContentDetail.body.parent_id ? [fetchContentDetail.body.parent_id] : []
+        folderDetail.fileName = fetchContentDetail.body.filename
+        folderDetail.folderParentsId = fetchContentDetail.body.parent_id ? [fetchContentDetail.body.parent_id] : []
+
+        let fetchLoopCondition = fetchContentDetail.body.parent_id !== null
```
Skylsmoi: if you create a variable for a boolean test, the variable name should reflect what the test means, here, you can call it `hasReachRootWorkspace`. That way, it tells what `parent_id === null` means

frontend_app_gallery/src/container/Gallery.jsx
Skylsmoi: @metzgegu I think you did not fix this. If the api doesn't returns http status 200, your iteration variable wont be updated and might spam api calls

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -164,34 +166,43 @@ class Gallery extends React.Component {
...
+            }
           } else {
             this.sendGlobalFlashMessage(props.t('Error while loading folder detail'))
             break
           }
         }
 
-        return { fileName, folderParentsId }
+        return folderDetail
       default: this.sendGlobalFlashMessage(props.t('Error while loading folder detail'))
```
Skylsmoi: default case doesn't return anything, in this case, function will return undefined. Even you handle the case where you call that function, it is usually a bad practice

frontend_app_gallery/src/container/Gallery.jsx
Skylsmoi: @metzgegu still not fixed. Your function can still return undefined in the `default` case

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -154,22 +166,48 @@ class Gallery extends React.Component {
...
+    let folderDetail
+
+    let fetchContentDetail = await handleFetchResult(
       await getFolderDetail(state.config.apiUrl, workspaceId, folderId)
     )
 
     switch (fetchContentDetail.apiResponse.status) {
       case 200:
-        this.setState({ fileName: fetchContentDetail.body.filename })
+        folderDetail = {}
```
Skylsmoi: don't initialize your variable to undefined and then reinitialize it to an empty object. Just initialize it to what you need. It'll also fix the fact that your function can return undefined

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -154,22 +166,48 @@ class Gallery extends React.Component {
...
+          )
+          if (fetchContentDetail.apiResponse.status === 200) {
+            if (fetchContentDetail.body.parent_id === null || prevParentId === fetchContentDetail.body.parent_id) {
+              hasReachRootWorkspace = false
+            } else {
+              folderDetail.folderParentsId.push(fetchContentDetail.body.parent_id)
+            }
+          } else {
+            this.sendGlobalFlashMessage(props.t('Error while loading folder detail'))
+            hasReachRootWorkspace = false
```
Skylsmoi: if `false` in case of an api error, you still end up in an infinite loop

Skylsmoi: wrong comment. I resolve this conversation

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -154,22 +166,46 @@ export class Gallery extends React.Component {
...
+          }
+        }
         break
-      default: this.sendGlobalFlashMessage(props.t('Error while loading folder detail'))
+      default:
+        this.sendGlobalFlashMessage(props.t('Error while loading folder detail'))
     }
+    return folderDetail
```
Skylsmoi: if you know `folderDetail` will always have the 2 properties `fileName` and `folderParentsId`, initialize them when you declare your variable

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -234,7 +277,7 @@ export class Gallery extends React.Component {
     }))).filter(i => i !== false)
   }
 
-  loadWorkspaceData = async () => {
+  loadWorkspaceLabel = async () => {
```
Skylsmoi: this function can return undefined in case of not api http status 200

Issue #2541, opened 2019-12-02
buttons are all the time visible in fullscreen and Slideshow


PhilippeAccorsi:
## Bug description & symptom

Button to rotate image, zoom, etc are visible all the time also when we are in fullscreen or in slideshow.

## Diagnostic

*Optional - explain what has been found and what is not yet clear. If unknown, write "unknown".*

## Expected

- [x] button's are visible if I place my mouse near corner for exemple. Normal process is to show this button at the begining but not after (just if I move mouse in button zone).

## How to reproduce

- Open image in full screen and start slideshow

## Impacted components

- *frontend*

## Extra information

Skylsmoi:
I agree and I would add that the buttons should only be visible when moving the cursor and should fadeout when mouse is idle
Like in video player in fullscreen (youtube, netflix, vlc ...)

PhilippeAccorsi:
Regression when you are in fullscreen, button is visible but we can't click on it.

Question: is it possible to activate white view on button when mouse is over (like its visible for cross or zoom)

metzgegu:
Yes it is possible to activate white color on button when mouse is hover,
however `React-image-lightbox` don't provide any props to add tooltip on prev or next button

Also can you give more information about how to reproduce the regression bug in fullscreen mode ?

PhilippeAccorsi:
I think the bug appear before we revert this commit https://github.com/tracim/tracim/commit/82c15583b9753e26b9e14363739a5dba87863a34

metzgegu:
TODO : 

- [ ] Hide arrows buttons when auto play enabled
- [ ] Add hover style to react-lightbox custom buttons
- [ ] Add title to next and prev buttons in react-lightbox

PhilippeAccorsi:
After discuss with @Skylsmoi and @metzgegu :
- Hide arrows buttons also when mouse not moving (like we have for gallery__action__button__lightbox ).

PhilippeAccorsi:
> however `React-image-lightbox` don't provide any props to add tooltip on prev or next button
> 
It is the same problem for zoom button and cross to close fullscreen?

metzgegu:
> > however `React-image-lightbox` don't provide any props to add tooltip on prev or next button
> 
> It is the same problem for zoom button and cross to close fullscreen?

Yes

PhilippeAccorsi:
Tested on branch and verified (wait merge to close)

PhilippeAccorsi:
Verified on develop.

[...]

PR #2578, opened 2019-12-10
[for 2.6.0] Fix/2541  hide lightbox buttons in app gallery when mouse not active

metzgegu:
#2541 

## Checkpoints

- [x] Code is clear enough

### DIFF
--- a/frontend_app_gallery/src/container/Gallery.jsx
+++ b/frontend_app_gallery/src/container/Gallery.jsx
@@ -406,77 +406,77 @@ class Gallery extends React.Component {
     if (state.imagesPreviews[state.fileSelected]) this.lightboxRotation.changeAngle(state.imagesPreviews[state.fileSelected].rotationAngle)
 
     return (
-      <div class='gallery__mouse__listener' onMouseMove={this.onMouseMove}>
-        <PageWrapper customClass='gallery'>
-          <PageTitle
-            title={state.folderId ? state.fileName : state.content.workspaceLabel}
-            icon={'picture-o'}
-            breadcrumbsList={state.breadcrumbsList}
-          />
+      <PageWrapper customClass='gallery'>
+        <PageTitle
+          title={state.folderId ? state.fileName : state.content.workspaceLabel}
+          icon={'picture-o'}
+          breadcrumbsList={state.breadcrumbsList}
+        />
 
-          <PageContent>
-            <div className='gallery__action__button'>
-              <button
-                className='btn outlineTextBtn nohover primaryColorBorder'
-                onClick={() => this.onClickSlickPlay(!state.autoPlay)}
-              >
-                <span className='gallery__action__button__text'>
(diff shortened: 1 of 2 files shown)


frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -386,168 +386,184 @@ class Gallery extends React.Component {
...
-            </button>
-
-            <button
-              className='btn outlineTextBtn nohover primaryColorBorder gallery__action__button__rotation__left'
-              onClick={() => this.rotateImg(state.fileSelected, DIRECTION.LEFT)}
-            >
-              <span className='gallery__action__button__text'>{props.t('Rotate 90° left')}</span>
-              <i className={'fa fa-fw fa-undo'} />
-            </button>
+      <div class='gallery__mouse__listener' onMouseMove={this.onMouseMove}>
```
Skylsmoi: I wont agree to this without a very good reason. This is something you must ask me beforehand before I can't be sure to check all these small but important things on a code review.

The purpose of a code review is to check code quality, not architecture. 

There is a PageWrapper component to ensure that every fullscreen app have the same layout
Adding a wrapper might compromise the homogeneity of tracim

I also need a feedback on the React possible built-in features for css modification on mouse move events

metzgegu: React provide the the function `onMouseMove` to record the mouse moving in a dom element, so I added a div which wrap all the gallery but yes it isn't the best solution.

I need to add a `div` which is only displayed when the lightbox is open, this `div` need to be fixed and have the size of the screen in order to catch `onMouseMove`.

Css modifications use simple javascript because it is a hack for imageLightBox.

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -406,77 +406,77 @@ class Gallery extends React.Component {
...
-            {state.displayLightbox && (
+          <Fullscreen
+            enabled={this.state.fullscreen}
+            onChange={fullscreen => this.setState({ fullscreen })}
+          >
+            <div ref={modalRoot => (this.modalRoot = modalRoot)} />
+          </Fullscreen>
+
+          {state.displayLightbox && (
+            <div className='gallery__mouse__listener' onMouseMove={this.onMouseMove}>
```
Skylsmoi: the handler should not be named as the event

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -386,6 +386,20 @@ class Gallery extends React.Component {
     return state.imagesPreviews[state.fileSelected].rawFileUrl
   }
 
+  onMouseMove = () => {
+    clearInterval(this.mouseMoveTimeout)
+    if (this.state.displayLightbox) {
+      document.getElementsByClassName('ril__toolbar')[0].style.transform = 'translateY(0px)'
```
Skylsmoi: add comment to explain why you do DOM manipulation

Issue #2588, opened 2019-12-12
Add context manager and use Session subclass to allow bypass revision protection


inkhey:
## Bug description & symptom

To make #2401 work correctly, we need to add a mecanism allowing us to by pass revision protection check.

## Expected

Allowing to bypass revision protection with a special context manager.

PhilippeAccorsi:
Verified.

PR #2589, opened 2019-12-12
allow bypassing revision deletion check using context manager and tra…

inkhey:
…cimSession subclass

related to #2588 

## Checkpoints

- [x] Code is clear enough

### DIFF
--- a/backend/tracim_backend/models/tracim_session.py
+++ b/backend/tracim_backend/models/tracim_session.py
@@ -9,8 +9,8 @@ class TracimSession(Session):
     Subclass of Sqlalchemy session to add tracim specific stuff
     """
 
-    def __init__(self, *arg, **kwargs) -> None:
-        super().__init__(self, *arg, **kwargs)
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(*args, **kwargs)
         self._allow_revision_deletion = False  # type: bool
 
     def set_allowed_revision_deletion(self, value: bool) -> None:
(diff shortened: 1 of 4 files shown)


backend/tracim_backend/models/tracim_session.py
```diff
@@ -0,0 +1,24 @@
...
+        self.allow_revision_deletion = False
+
+
+@contextmanager
+def unsafe_tracim_session(session: TracimSession):
+    """
+    allow to bypass protection on tracim revisions
+    """
+    original_allow_revision_deletion = session.allow_revision_deletion
+    session.allow_revision_deletion = True
```
lebouquetin: I'm not found of this pattern which uses a statefull class instead of building a dedicated object. I'd like to talk about this tomorrow moring in order to be sure there is no better way to do it

backend/tracim_backend/models/tracim_session.py
```diff
@@ -0,0 +1,24 @@
...
+    Subclass of Sqlalchemy session to add tracim specific stuff
+    """
+
+    def __init__(self, **kwargs):
+        Session.__init__(self, **kwargs)
+        self.allow_revision_deletion = False
+
+
+@contextmanager
+def unsafe_tracim_session(session: TracimSession):
```
buxx: I would prefer a naming with "content_revision" (this context manager only allow "revision_deletion").

inkhey: renamed to "unprotected_content_revision"

Issue #1891, opened 2019-06-10
Responsive problems with tracim in fullscreen mode on iPhone 6s 


celiamgtt:
## Bug description & symptom

There is some responsive problems with tracim in fullscreen mode on IPhone 6s :
- the sidebar button is missing (we can close and open the sidebar by clicking on the expected location of the button, but the button isn't displayed)
- when a document has a horizontal scrolling activate, the vertical scrolling inside the document is impossible

## Diagnostic

"unknown"

## Expected

- [x] Sidebar button should be available/displayed
- [x] Vertical scrolling available inside a document

## Context and version information

    - Device: iPhone 6s
    - iOS: 11.2.6
    - Browser: Chrome & Safari

## Screenshots

![Peek 10-06-2019 12-41](https://user-images.githubusercontent.com/50453784/59190488-2043d500-8b7d-11e9-96d5-1ca6a9636fdf.gif)

![Peek 10-06-2019 12-40](https://user-images.githubusercontent.com/50453784/59190504-2d60c400-8b7d-11e9-8234-e10512d59e56.gif)

Skylsmoi:
This issues seems to be related to the parent div which has overflow: hidden and the sidebar arrow has position: fixed, the sidebar arrow become hidden only on IOS.
This workaround have been found [here](https://stackoverflow.com/questions/29482203/iphone-fixed-position-div-invisible

Skylsmoi:
to test this issue:
Verify that the btn to show/hide the sidebar exists and works on different devices/browser.
Then verify it exists and works on ios devices (use browserstack)

PhilippeAccorsi:
This is fix and tested => Vertical scrolling available inside a document
But other fix not working correctly.
When shared space is open (in sidebar), sidebar button is not displayed. We need to close sharedspace and after that, sometime sidebar button is displayed or not displayed.

PhilippeAccorsi:
Remove `.zip` to look video of bug

[bug_iphone_tracim_sidebar_low_quality.mp4.zip](https://github.com/tracim/tracim/files/4025758/bug_iphone_tracim_sidebar_low_quality.mp4.zip)

celiamgtt:
![gif_sidebar_mobile](https://user-images.githubusercontent.com/50453784/71814612-4121c980-307d-11ea-8663-9b74ccf05986.gif)

PhilippeAccorsi:
This fix is rollback because this fix add too many regression on tracim (app file, collabora, etc....).
[...]

PR #2598, opened 2019-12-31
[for 2.6.0] Fix hidden sidebar arrow on ios

metzgegu:
#1891 

This issues seems to be related to the parent div which has `overflow: hidden` and the sidebar arrow has `position: fixed`, the sidebar arrow become hidden only on IOS.
This workaround have been found [here](https://stackoverflow.com/questions/29482203/iphone-fixed-position-div-invisible)

## Checkpoints

- [x] Code is clear enough

### DIFF
--- a/frontend/src/css/Sidebar.styl
+++ b/frontend/src/css/Sidebar.styl
@@ -15,6 +15,9 @@ sidebarColorDarken = darken(sidebarColor, 15%)
   justify-content space-between
   height 100%
 
+.sidebarpagecontainer
+  -webkit-transform translateZ(0)
+
 .sidebar
   position relative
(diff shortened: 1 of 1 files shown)


frontend/src/css/Sidebar.styl
```diff
@@ -15,6 +15,9 @@ sidebarColorDarken = darken(sidebarColor, 15%)
   justify-content space-between
   height 100%
 
+.sidebarpagecontainer
```
Skylsmoi: please add a comment to explain this with the link to the stack issue

Skylsmoi:
we cannot test this because it would require to run test on an ios environment

Issue #2593, opened 2019-12-19
Page title should change according to current route


metzgegu:
## Feature description

In order to easily navigate through tracim contents, the title should change depending of the current route

## Context

Page title need to change when the route is changed

## Impacted components

Frontend

metzgegu:
Currently the `title` is hardcoded in index.html (eg `<title>Tracim</title>`) but I think it should be a parameter in development.ini in order to set the default title with the instance name and change the title according to which app is currently rendered

metzgegu:
### We need to define which pages will change the default title.
Obvious pages are contents app pages : 
- App file
- App html-document
- App thread
- App collabora online
Other interesting pages : 
- Workspace content list, dashboard, gallery, agenda
- Admin interfaces
- Account management

Skylsmoi:
spec [here](https://algoo.tracim.fr/ui/workspaces/49/contents/file/11518)

PhilippeAccorsi:
problem with 
- ui/guest-upload | Public upload => wording here is `Share folder` but page name is `Received files`
- ui/guest-upload/ => title is not translated dynamically
- when file is opened with collabora, title is not refreshed and if I refreshed page with F5 title just show website.title (value visible in development.ini).

PhilippeAccorsi:
Verified on branch.

- ~~ui/guest-upload | Public upload => wording here is Share folder but page name is Received files~~
- ~~ui/guest-upload/ => title is not translated dynamically~~
- ~~when file is opened with collabora, title is not refreshed and if I refreshed page with F5 title just show website.title (value visible in development.ini).~~

PhilippeAccorsi:
Verified on develop.

PR #2602, opened 2020-01-02
[for 2.6.0] Dynamic title tag according to current page

metzgegu:
#2593 

## Checkpoints

- [ ] Code is clear enough
- [x] tested and verified.


frontend/src/container/Tracim.jsx
```diff
@@ -127,7 +131,10 @@ class Tracim extends React.Component {
     const { props } = this
 
     const fetchGetConfig = await props.dispatch(getConfig())
-    if (fetchGetConfig.status === 200) props.dispatch(setConfig(fetchGetConfig.json))
+    if (fetchGetConfig.status === 200) {
+      props.dispatch(setConfig(fetchGetConfig.json))
+      document.title = fetchGetConfig.json.instance_name
```
Skylsmoi: you are bypassing the custom event process here. Even if it is in the same file as where you handle the custom event, keep using it's process.
It's only a matter of consistency

frontend/src/container/Dashboard.jsx
```diff
@@ -130,6 +130,7 @@ class Dashboard extends React.Component {
         if (props.appList.some(a => a.slug === 'agenda') && fetchWorkspaceDetail.json.agenda_enabled) {
           this.loadCalendarDetail()
         }
+        GLOBAL_dispatchEvent({ type: CUSTOM_EVENT.SET_TITLE, data: { title: fetchWorkspaceDetail.json.label } })
```
Skylsmoi: shouldn't you also handle the event ALL_APP_CHANGE_LANGUAGE in this file ?

frontend/src/container/WorkspaceContent.jsx
```diff
@@ -113,6 +113,7 @@ class WorkspaceContent extends React.Component {
         props.history.push(PAGE.WORKSPACE.CONTENT_LIST(state.workspaceIdInUrl) + '?' + qs.stringify(newUrlSearch, { encode: false }))
         this.setState({ appOpenedType: false })
 
+        this.setPageTitle(props.currentWorkspace.label)
```
Skylsmoi: I think this is the correct approach for when you have more than one place to call the custom event in a single file.
Please refactor every files calling the custom event more than one with this approach.

But rename the function name according to the new custom event name (see comment bellow)

frontend/src/container/Tracim.jsx
```diff
@@ -89,6 +89,10 @@ class Tracim extends React.Component {
         console.log('%c<Tracim> Custom event', 'color: #28a745', type, data)
         this.props.dispatch(appendBreadcrumbs(data.breadcrumbs))
         break
+      case CUSTOM_EVENT.SET_TITLE:
+        console.log('%c<Tracim> Custom event', 'color: #28a745', type, data)
+        document.title = data.title
```
Skylsmoi: I think we should always show the instance name at the end of the title, and/or always have the word tracim. Please do some research about what should be included in the title.

frontend_lib/src/customEvent.js
```diff
@@ -17,6 +17,7 @@ export const CUSTOM_EVENT = {
   REFRESH_WORKSPACE_LIST_THEN_REDIRECT: 'refreshWorkspaceList_then_redirect',
   RELOAD_CONTENT: type => `${type}_reloadContent`,
   SET_BREADCRUMBS: 'setBreadcrumbs',
+  SET_TITLE: 'setTitle',
```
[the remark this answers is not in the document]
metzgegu: Is `SET_HEAD_TITLE` maybe even better ? because `<title>` is in the `<head>` tag

frontend/src/container/Dashboard.jsx
metzgegu: Not in this case because I only display the label of the current workspace, but we still need to discuss about which title should be for all the page, (e.g in dashboard `Dashbord {{workspaceLabel}}` and `ContentList {{workspaceLabel}}`)

frontend/src/container/Tracim.jsx
metzgegu: After some research, I think we should too, a lot of web application use `{{pageName}} - {{appName}}` (google drive, OneDrive, gitLab), but what will be the appName ? `Tracim` or the specific instance name ?

frontend_lib/src/customEvent.js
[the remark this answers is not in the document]
Skylsmoi: I agree, it's even better

frontend/src/container/Tracim.jsx
Skylsmoi: I don't have the answer, do ask around and investigate to be able to give Damien all the info he needs to decide (pros and cons of both possibilities)

frontend/src/container/Dashboard.jsx
Skylsmoi: ok for ALL_APP_CHANGE_LANGUAGE.

For the list of titles, prepare a document with the exhaustive list of pages that requires a title change with the title you think would fit for each

frontend_lib/src/customEvent.js
```diff
@@ -17,7 +17,7 @@ export const CUSTOM_EVENT = {
   REFRESH_WORKSPACE_LIST_THEN_REDIRECT: 'refreshWorkspaceList_then_redirect',
   RELOAD_CONTENT: type => `${type}_reloadContent`,
   SET_BREADCRUMBS: 'setBreadcrumbs',
-  SET_TITLE: 'setTitle',
+  SET_HEAD_TITLE: 'setTitle',
```
Skylsmoi: setTitle should also be changed to setHeadTitle

frontend/src/container/Account.jsx
```diff
@@ -77,7 +77,10 @@ class Account extends React.Component {
 
   customEventReducer = ({ detail: { type, data } }) => {
     switch (type) {
-      case CUSTOM_EVENT.ALL_APP_CHANGE_LANGUAGE: this.buildBreadcrumbs(); break
+      case CUSTOM_EVENT.ALL_APP_CHANGE_LANGUAGE:
+        this.buildBreadcrumbs()
+        this.setHeadTitle(this.props.t('My Account'))
```
Skylsmoi: this will break once the PR about merge generic code of apps is merged.
We'll see what to do

PhilippeAccorsi:
@metzgegu this is ok for me. All working good.

Issue #617, opened 2018-07-03
Reorganise backend code to make him more modular, with "applications"


inkhey:
This issues is about reorganising code to make code use "application" who have:
- controllers
- schema
- models
- metadata (name, color, ...)
- associated content_type
- config
- etc

Todo:

- [ ] put specific application code in specific dir in applications dir.
- [ ] make load/unload of app dynamic (https://github.com/tracim/tracim/issues/2555)
- [ ] clarify enable/disable of app in config (see https://github.com/tracim/tracim/issues/2215)
- [ ] Upgrade docker file

### Original issue:

> # Reorganise schema/models/controllers for app with plugin architecture in mind
> 
> > (from le bouquetin) You will probably have a folder named /plugins/html-document where you will find models, controllers, etc.
> 
> folder app pluginification of :
> - [ ] controllers
> - [ ] marshmallow_schema
> - [ ] models related
> - [ ] others ?

inkhey:
As we do have example for share and upload permission app, we can follow the same folder organization.

inkhey:
@PhilippeAccorsi :
- This PR convert CALDAV_ENABLED and COLLABORATIVE_DOCUMENT_EDITION__ACTIVATED deprecated as deprecated parameters.
- To activate/disable any app including agenda and collaborative document edition, you need to set app.enabled.
- app.enabled order change ordering in ui currently, we should use order of app available in development.ini.sample for docker configuration.

PhilippeAccorsi:
> @PhilippeAccorsi :
> 
>     * This PR convert CALDAV_ENABLED and COLLABORATIVE_DOCUMENT_EDITION__ACTIVATED deprecated as deprecated parameters.
> 
>     * To activate/disable any app including agenda and collaborative document edition, you need to set app.enabled.
> 
>     * app slug order in app.enabled change ordering in ui currently, we should use order of app available in development.ini.sample for docker configuration.

Ok. So for docker I need to make some change, and change documentation about how to use feature.

inkhey:
Some frontend bug are found related to this new architecture:
- if app `gallery` is disabled, folder edit should not propose `gallery`.
- if app `share_content` app is disabled and `public_download_enabled` is true in workspace, frontend should not propose enabled/disable public download and should not load/display share for content.
- if app `upload_permission` app is disabled and `public_upload_enabled` is true in workspace, frontend should not propose enabled/disable public upload and should not show "Received files" dir.

metzgegu:
[...]

Issue #2326, opened 2019-08-29
Clarify role of frontend and backend in dashboard app list view


inkhey:
## Bug description & symptom

Currently use both applications endpoint and content_types endpoint to create list of buttons in dashboard. Some decision are made by backend (origin of content) but some other are done in frontend (display or not display folder depending on role). To handle trouble found in #2194, fix was made in frontend to not show share apps. 

Problem is that some part of intelligence are in frontend and some other in backend. We should decide how we deal with this. This is tricky as some buttons may refer to content_types (and app may have multiple content_type) and other like agenda may refer to other thing.

## Diagnostic

from :  https://github.com/tracim/tracim/issues/2194

> Frontend side: the dashboard button list is created from the list of alls applications installed in this tracim instance sent in by the backend (endpoint: /api/v2/system/applications). After, some elements of this list are filtered if the user role is not enough (folder case) or the application is disabled (agenda case). The share_folder is always filtered.

_Originally posted by @gbonaspetti in https://github.com/tracim/tracim/issues/2194#issuecomment-526084039_

## Expected

all intelligence in backend with api clarification ?

inkhey:
**Current spec:**

/api/v2/system/applications
```
[
  {
    "config": {},
    "fa_icon": "file-o",
    "hexcolor": "#FF0000",
    "is_active": true,
    "label": "Agenda",
    "slug": "agenda"
  }
]
```

/api/v2/system/content_types
```
[
  {
    "available_statuses": [
      {
        "fa_icon": "fa-check",
        "global_status": "open",
        "hexcolor": "#0000FF",
        "label": "Opened",
        "slug": "open"
      }
    ],
    "creation_label": "Write a document",
    "fa_icon": "fa-file-text-o",
    "hexcolor": "#FF0000",
    "label": "Text Documents",
    "slug": "pagehtml"
  }
]
```

/api/v2/workspaces (see sidebar entry)
```
[
  {
    "agenda_enabled": true,
    "created": "2020-01-06T10:35:32.507Z",
    "description": "All intranet data.",
    "is_deleted": false,
    "label": "Intranet",
    "owner": {
      "avatar_url": "/api/v2/asset/avatars/john-doe.jpg",
      "public_name": "John Doe",
      "user_id": 3
    },
    "public_download_enabled": true,
    "public_upload_enabled": true,
    "sidebar_entries": [
      {
        "fa_icon": "file-text-o",
        "hexcolor": "#F0F9DC",
[...]

Issue #1834, opened 2019-06-03
Handle other non static context params than config one.


inkhey:
## Bug description & symptom

Config CFG is a nice object but should stay just config. We do need to allow having other global  context information than config:
- App list should be separated from config and not a static object anymore
- depot storage ?
- Pluggy mecanism need to have a context about events to allow subscribe or activate them.
- IndexedContent class for elasticsearch should be pass as parameter to SearchLib.

## Expected

- wait reduce code duplication in tests : https://github.com/tracim/tracim/issues/619.
- need discussion.

inkhey:
To drop app_list and content_types_list staticness:
- remove staticness of content_type_list -> create specific lib which get app_list in entry 
- add parameters to `update_validators`

.

inkhey:
Current code force long refactoring if we need to add new object to pass to lib, for example, we will need later to pass both app_list (as it should be not static anymore) and event_dispatcher for many lib.

These lib are heavily used in controller, if we decide to add a new param, we had to pass it at many place which mean long and mostly manual refactor.

To avoid this situation, we can:
1. Use a big object that contain anything. This should work and make refactor really easy but with this we do lost some information about which data lib required
2. Use a context-based factory, similar as what is done for test. We can rely on TracimContext to add  libfactory method which will create a contextualized lib.

Example of current code:
```
    @hapic.with_api_doc(tags=[SWAGGER_TAG_SYSTEM_ENDPOINTS])
    @check_right(is_user)
    @hapic.output_body(ApplicationSchema(many=True))
    def applications(self, context, request: TracimRequest, hapic_data=None):
        """
        Get list of alls applications installed in this tracim instance.
        """
        app_config = request.registry.settings["CFG"]  # type: CFG
        app_api = ApplicationApi(app_list=app_list)
        applications_in_context = [
            app_api.get_application_in_context(app, app_config) for app in app_api.get_all()
        ]
        return applications_in_context
```

With solution 1:

```
    @hapic.with_api_doc(tags=[SWAGGER_TAG_SYSTEM_ENDPOINTS])
    @check_right(is_user)
    @hapic.output_body(ApplicationSchema(many=True))
    def applications(self, context, request: TracimRequest, hapic_data=None):
        """
        Get list of alls applications installed in this tracim instance.
        """
        current_context = request.current_context
        app_api = ApplicationApi(current_context)
        applications_in_context = [
            app_api.get_application_in_context(app) for app in app_api.get_all()
        ]
        return applications_in_context
[...]

PR #2618, opened 2020-01-13
Feat/617  reorganise apps

inkhey:
related to  #617 

## Checkpoints

- [x] Code is clear enough
- [x] Tested by quality team


backend/tracim_backend/applications/agenda/application.py
```diff
@@ -0,0 +1,114 @@
...
+            radicale_base_path=app_config.CALDAV__RADICALE__BASE_PATH,
+            radicale_user_path=app_config.CALDAV__RADICALE__USER_PATH,
+            radicale_workspace_path=app_config.CALDAV_RADICALE_WORKSPACE_PATH,
+        )
+        agenda_controller = AgendaController()
+        configurator.include(agenda_controller.bind, route_prefix=BASE_API_V2)
+        configurator.include(radicale_proxy_controller.bind)
+
+
+def create_app() -> TracimApplication:
```
inkhey: To import app, we do need to know where to find them. After discussion with @buxx about multiple solution, we decide to have a "create_app" method in application.py module of each applicaiton folder.

- Convention naming instead of automatic subclass fetching because subclass iterating, subclass iterating force to do recursive iteration and is less clear/more magic.
- Only this method outside of class here, allowing relying on explicit documentation for other methods with subclassing.
- Method instead of application object to avoid staticness (this may cause trouble during running test)

backend/tracim_backend/app_models/applications.py
```diff
@@ -1,80 +1,59 @@
...
 if typing.TYPE_CHECKING:
-    from tracim_backend.config import CFG  # noqa: F401
-    from tracim_backend.app_models.contents import ContentStatus  # noqa: F401
+    from tracim_backend.config import CFG
+    from tracim_backend.lib.utils.app import TracimApplication
+    from tracim_backend.lib.utils.app import TracimContentType  # noqa F401
 
 
-class Application(object):
+class TracimApplicationInContext(object):
```
inkhey: We do need app in context for color (color of app need cfg to be obtained.).

backend/tracim_backend/command/cleanup.py
```diff
@@ -166,36 +168,44 @@ def take_app_action(self, parsed_args: argparse.Namespace, app_context: AppEnvir
...
+                            print(
+                                'Warning: Cannot delete agenda for user "{}", agenda not found. Agenda path may be incorrect or agenda not created'.format(
+                                    user_id
+                                )
                             )
-                        )
-                        print(traceback.format_exc())
+                            print(traceback.format_exc())
+            else:
+                print("Warning ! Agenda app not enabled, agenda will not be deleted.")
```
inkhey: We do not have anymore config of app if these app are not loaded, so we can not anymore expect caldav calendar to be delete in agenda app is not loaded. Not sure how and if we do need to fix this.

backend/tracim_backend/config.py
```diff
@@ -70,13 +68,23 @@ def __init__(self, settings: typing.Dict[str, typing.Any]):
         # with object in some context
         self.settings = settings.copy()
         self.config_naming = []  # type: typing.List[ConfigParam]
-        logger.debug(self, "CONFIG_PROCESS:1: load config from settings")
+        logger.debug(self, "CONFIG_PROCESS:1: load enabled apps")
```
inkhey: We do need to load app_enabled config first, then load app and them we can do load/check like before.

backend/tracim_backend/config.py
```diff
@@ -131,10 +139,79 @@ def get_raw_config(
...
+            "{extend_apps}"
+            "share_content,"
+            "upload_permission,"
+            "gallery"
+        )
+        extend_apps = ""
+        # TODO - G.M - 2020-01-09 - remove this retrocompat code, as
+        #  CALDAV__ENABLED and COLLABORATIVE_DOCUMENT_EDITION__ACTIVATED
+        # should be deprecated.
+        self.CALDAV__ENABLED = asbool(self.get_raw_config("caldav.enabled", "false"))
```
inkhey: CALDAV__ENABLED and COLLABORATIVE_DOCUMENT_EDITION__ACTIVATED are deprecated. I create a retrocompat code, but i'm not sure that's a good idea.

backend/tracim_backend/config.py
```diff
@@ -131,10 +139,79 @@ def get_raw_config(
...
+
+    def _load_enabled_app(
+        self, enabled_app_slug_list: typing.List[str], loaded_apps: typing.List[TracimApplication]
+    ) -> None:
+
+        # TODO - G.M - 2018-08-08 - [GlobalVar] Refactor Global var
+        # of tracim_backend, Be careful app_list is a global_var
+        app_list.clear()
+        # INFO - G.M - 2020-01-09 - order of app in app_list should be:
+        # - all enabled app according to enabled_app_slug_list order
```
inkhey: This code is complex to preserve order from config file. We should probably deal with order differently. Order is directly associated to entry, so we should probably do this in https://github.com/tracim/tracim/issues/2326.

backend/tracim_backend/config.py
```diff
@@ -131,10 +139,79 @@ def get_raw_config(
...
+            do_strip=True,
+        )
+
+    def _load_enabled_app(
+        self, enabled_app_slug_list: typing.List[str], loaded_apps: typing.List[TracimApplication]
+    ) -> None:
+
+        # TODO - G.M - 2018-08-08 - [GlobalVar] Refactor Global var
+        # of tracim_backend, Be careful app_list is a global_var
+        app_list.clear()
```
inkhey: app_list and content_type_list are static object, we should probably make them non-static later, see https://github.com/tracim/tracim/issues/1834

backend/tracim_backend/lib/core/user.py
```diff
@@ -887,7 +889,8 @@ def execute_updated_user_actions(self, user: User) -> None:
         # event mecanism is ready, see https://github.com/tracim/tracim/issues/1487
         # event on_updated_user should start hook use by agenda  app code.
 
-        if self._config.CALDAV__ENABLED:
+        app_lib = ApplicationApi(app_list=app_list)
```
inkhey: This kind of code, will be moved to app specific code as soo as new event are added for this. see https://github.com/tracim/tracim/issues/1487

backend/tracim_backend/lib/utils/app.py
```diff
@@ -0,0 +1,110 @@
...
+        self.creation_label = creation_label
+        self.available_statuses = available_statuses
+        self.slug_alias = slug_alias
+        self.allow_sub_content = allow_sub_content
+        self.file_extension = file_extension
+        self.minimal_role_content_creation = minimal_role_content_creation
+        self.app = app
+
+
+class TracimApplication(ABC):
```
inkhey: TracimApplication Spec.

backend/tracim_backend/app_models/applications.py
lebouquetin: probably better to name this class something like `ConfigReadyApplication` or something like this? What do you think?

backend/tracim_backend/app_models/contents.py
```diff
@@ -105,34 +107,59 @@ def get_default_status(self) -> ContentStatus:
 # ContentType
 
 
-class ContentType(object):
+class ContentTypeInContext(object):
```
lebouquetin: same suggestion: ConfigReadyContentType? (not sure, please explain if you think this is not a good idea)

backend/tracim_backend/applications/agenda/application.py
```diff
@@ -0,0 +1,114 @@
...
+        if radicale_storage_type != "multifilesystem":
+            raise ConfigurationError(
+                '"{}" should be set to "{}"'
+                " (currently only valid value)"
+                ' when "{}" is true'.format(
+                    "caldav.radicale.storage.type", "multifilesystem", "caldav.enabled"
+                )
+            )
+
+    def import_controllers(
```
lebouquetin: I'd suggest `load_controllers()` instead of `import_controllers()`

backend/tracim_backend/applications/agenda/application.py
lebouquetin: `create_app()` is not a *method* but a *function* - it is important to use the right vocabulary to make discussion clear. I do not see why a class method would be more static than function, could you explain?

backend/tracim_backend/lib/utils/app.py
lebouquetin: As this class is a spec (specification?), you should explain in docstring how to use / inherit / reimplement it.

backend/tracim_backend/lib/utils/app.py
```diff
@@ -0,0 +1,110 @@
...
+    def __init__(
+        self, label: str, slug: str, fa_icon: str, config: typing.Dict[str, str], main_route: str
```
lebouquetin: every string parameter should have an explanation about what values can be added. If I put "hey, bos ! what do you think about this?" is it pertinent for label / slug / fa_icon / config / main route ? give what type of string is expected, give examples

backend/tracim_backend/lib/utils/app.py
```diff
@@ -0,0 +1,110 @@
...
+    ) -> None:
+        self.label = label
+        self.slug = slug
+        self.fa_icon = fa_icon
+        self.config = config
+        self.main_route = main_route
+        self.content_types = []  # typing.List[ContentType]
+        self.is_active = False
+
+    def add_content_type(
```
lebouquetin: add docstring. Should the developer overwrite this method? What kind of values are expected for str parameters?

backend/tracim_backend/lib/utils/app.py
```diff
@@ -0,0 +1,110 @@
...
+        self,
+        label: str,
+        slug: str,
+        creation_label: str,
+        available_statuses: typing.List["ContentStatus"],
+        slug_alias: typing.List[str] = None,
```
lebouquetin: `slug_alias should be `slug_aliases`

backend/tracim_backend/lib/utils/app.py
```diff
@@ -0,0 +1,110 @@
...
+            available_statuses=available_statuses,
+            slug_alias=slug_alias,
+            allow_sub_content=allow_sub_content,
+            file_extension=file_extension,
+            minimal_role_content_creation=minimal_role_content_creation,
+            app=self,
+        )
+        self.content_types.append(content_type)
+
+    def load_content_types(self):
```
lebouquetin: As a developer, if I read this I understand that I must overwrite this method. But I have no information about:

- what the method should do
- what the method should return

Please add a real docstring, with examples, expected behavior, values, etc

backend/tracim_backend/lib/utils/app.py
```diff
@@ -0,0 +1,110 @@
...
+        """load app content type"""
+        raise NotImplementedError()
+
+    def get_application_in_context(self, app_config: "CFG") -> TracimApplicationInContext:
```
lebouquetin: same remark.This kind of method MUST include a large docstring explaining how to implement it

backend/tracim_backend/lib/utils/app.py
```diff
@@ -0,0 +1,110 @@
...
+    def load_config(self, app_config: "CFG") -> None:
+        """
+        Allow to load specific config parameter, example:
+        >>> app_config.TEST__EXAMPLE_CONFIG_PARAMETER = app_config.get_raw_config("test.example_config_parameter")
+        """
+        raise NotImplementedError()
+
+    def check_config(self, app_config: "CFG") -> None:
+        """
+        Check app specific config consistency, example:
```
lebouquetin: How should I implement the error case? Is there a specific exception to raise? Please add examples of "how to implement this"

backend/tracim_backend/lib/utils/app.py
```diff
@@ -0,0 +1,110 @@
...
+        >>> app_config.check_mandatory_param("TEST__EXAMPLE_CONFIG_PARAMETER", app_config.TEST__EXAMPLE_CONFIG_PARAMETER)
+        """
+        raise NotImplementedError()
+
+    def import_controllers(
```
lebouquetin: missing complet docstring + probably rename `import` to `load` (or explain why import is better)

backend/tracim_backend/lib/utils/request.py
```diff
@@ -111,7 +111,7 @@ def candidate_workspace(self) -> Workspace:
         )
 
     @property
-    def candidate_content_type(self) -> ContentType:
+    def candidate_content_type(self) -> TracimContentType:
```
lebouquetin: what is a candidate_content_type? The docstring is not clear about what this does.

backend/tracim_backend/lib/utils/request.py
```diff
@@ -186,7 +186,9 @@ def _get_content(self, content_id_fetcher: typing.Callable[[], int]) -> Content:
             content_type=content_type_list.Any_SLUG,
         )
 
-    def _get_content_type(self, content_type_slug_fetcher: typing.Callable[[], str]) -> ContentType:
+    def _get_content_type(
+        self, content_type_slug_fetcher: typing.Callable[[], str]
```
lebouquetin: with a signature like this, this is *required* to add a docstring explaining what is expected as parameters, as result + give examples

backend/tracim_backend/lib/utils/utils.py
```diff
@@ -392,6 +393,31 @@ def email_link(self) -> str:
         return "mailto:{email_address}".format(email_address=self.email_address)
 
 
+def load_apps() -> typing.List[TracimApplication]:
```
lebouquetin: I'm not sure load_apps function should be in `tracim_backend.lib/utils.utils`... this is a critical core feature

backend/tracim_backend/lib/utils/utils.py
```diff
@@ -392,6 +393,31 @@ def email_link(self) -> str:
...
+        module = importlib.import_module("{}.application".format(app_config_path))
+        tracim_apps.append(module.create_app())
+    return tracim_apps
+
+
+def find_direct_submodule_path(module: types.ModuleType) -> typing.List[str]:
+    """
+    Get path of submodules of given module
+    :param module: module to check
+    :return: list of path of direct submodule
```
lebouquetin: please add examples. is this something like `applications.agenda`? `applications/agenda`? is a value like `applications/agenda/somesubmodule` possible?

lebouquetin: documentation should be clear enough in order to understand how to use the function without to have to read its implementation

backend/tracim_backend/tests/functional/test_account.py
```diff
@@ -984,8 +985,8 @@ def test_api__get_account_workspaces__ok_200__nominal_case(
         app_api = application_api_factory.get()
```
lebouquetin: `get()` means "the object/instance already exist", otherwise it must be `create()` or `create_instance()`

backend/tracim_backend/applications/agenda/application.py
inkhey: Problem is about naming. If we want to import class or function, we do need to know his name.
Same name class is a bad idea for debugging, so function seems a better solution.

backend/tracim_backend/app_models/applications.py
inkhey: I'm not sure that a good naming. "InContext" class (this is not the first one, we do have them for database classe too) are object use to get extend information about an object. It's needed for REST api but we do not use them to act, it's just for showing information extended by the current context.

backend/tracim_backend/app_models/contents.py
inkhey: same explanation as for app.

backend/tracim_backend/lib/utils/request.py
inkhey: not application related, will be fixed in #2626

backend/tracim_backend/lib/utils/request.py
inkhey: not application related, will be fixed in #2626

backend/tracim_backend/lib/utils/app.py
inkhey: method has been dropped, it was not used, using applicationApi.get_application_in_context instead.

backend/tracim_backend/lib/utils/app.py
inkhey: examples added.

backend/tracim_backend/lib/utils/app.py
inkhey: example added + load instead of import.

backend/tracim_backend/command/cleanup.py
inkhey: @lebouquetin is this new behavior ok ?

backend/tracim_backend/config.py
inkhey: @lebouquetin have you any idea about how to handle these old parameters ?

backend/tracim_backend/lib/utils/app.py
inkhey: done.

backend/tracim_backend/applications/agenda/application.py
inkhey: done.

backend/tracim_backend/lib/utils/app.py
inkhey: no example for config as this parameter is not used now, but done for other

backend/tracim_backend/lib/utils/app.py
inkhey: done.

backend/tracim_backend/lib/utils/utils.py
inkhey: don't really know where we should but this, as tracim_backend.lib.core is for lib like ContentApi, etc...
we could also put if tracim_backend.lib.<something_else> or at rhigher level tracim_backend.<something> like config, error, hookspec.

backend/tracim_backend/tests/functional/test_account.py
inkhey: this is test related, all factory object for test use, ".get()" this refactoring should be done something else (not app related). see new issue #2627

backend/tracim_backend/lib/utils/app.py
inkhey: done.

backend/tracim_backend/lib/utils/app.py
inkhey: done. method dropped to have similar mecanism for all application methods.

backend/tracim_backend/lib/utils/utils.py
inkhey: done.

backend/tracim_backend/app_models/applications.py
```diff
@@ -1,80 +1,59 @@
...
-        self.slug = slug
-        self.fa_icon = fa_icon
-        self.hexcolor = self._get_hexcolor_or_default(slug, app_config)
-        self.is_active = is_active
-        self.config = config
-        self.main_route = main_route
-        self.content_types = []
+    def __init__(self, app_config: "CFG", app: "TracimApplication") -> None:
+        self.app = app
+        self.app_config = app_config
```
buxx: About `app_config` naming: it looks like a config specific for this app. But it is a global Tracim config right ?

inkhey: Yes. 

This name is widely use to distinct "config" and "settings" object which exist at some point of code because of usage of some lib (pyramid as a configurator object, caldav app use a "config" dict, settings is the current setting dict used by pyramid, etc...)

Until recently there was not naming issue because application doesn't really exist.

backend/tracim_backend/applications/agenda/application.py
buxx: >  I do not see why a class method would be more static than function, could you explain?

@lebouquetin 

Other solution was not a class method but a variable named application (like in wsgi files), like this:

**backend/tracim_backend/applications/example/application.py**

```python
# [...]

class ExampleApp(TracimApplication):
    # [...]

application = ExampleApp()
```

Bad side of this solution was:

*  ̀ExampleApp` instantiate at module loading
* `application` is static

backend/tracim_backend/command/cleanup.py
```diff
@@ -166,36 +168,44 @@ def take_app_action(self, parsed_args: argparse.Namespace, app_context: AppEnvir
...
-                for workspace_id in deleted_workspace_ids:
-                    try:
-                        cleanup_lib.delete_workspace_agenda(workspace_id)
-                    except AgendaNotFoundError:
-                        print(
-                            'Warning: Cannot delete agenda for workspace "{}", agenda not found. Agenda path may be incorrect or agenda not created'.format(
-                                workspace_id
-                            )
+            app_lib = ApplicationApi(app_list=app_list)
+            if app_lib.exist("agenda"):
```
buxx: Must-use constants for "core" Tracim applications.

backend/tracim_backend/command/cleanup.py
buxx: Is it possible to "force" activate agenda application when delete users ?

backend/tracim_backend/lib/core/application.py
```diff
@@ -6,33 +6,53 @@
...
-    from tracim_backend.app_models.applications import Application  # noqa:F401
-    from tracim_backend.app_models.contents import ContentType  # noqa: F401
+    from tracim_backend.config import CFG
+    from tracim_backend.app_models.applications import TracimApplicationInContext  # noqa:F401
+    from tracim_backend.app_models.contents import ContentTypeInContext  # noqa: F401
 
 
 class ApplicationApi(object):
-    def __init__(self, app_list: List["Application"], show_all: bool = False) -> None:
+    def __init__(self, app_list: List["TracimApplication"], show_all: bool = False) -> None:
```
buxx: Can you add a docstring about parameters ? (specially for `show_all`)

backend/tracim_backend/lib/core/application.py
```diff
@@ -6,33 +6,53 @@
...
-                return app
+                if self.show_all or app.is_active:
+                    return app
         raise AppDoesNotExist("Application {app} does not exist".format(app=slug))
 
-    def get_all(self) -> List["Application"]:
+    def get_application_in_context(
+        self, app: TracimApplication, app_config: "CFG"
+    ) -> "TracimApplicationInContext":
+        from tracim_backend.app_models.applications import TracimApplicationInContext
```
buxx: Wrong place import (or there is a reason ?)

buxx: If cyclic import or goo reason: add a note

backend/tracim_backend/lib/utils/app.py
```diff
@@ -0,0 +1,216 @@
...
+        self.slug = slug
+        self.fa_icon = fa_icon
+        self.config = config
+        self.main_route = main_route
+        self.content_types = []  # typing.List[ContentType]
+        # INFO - G.M - 2020-01-16 - will be set to true automatically by Tracim
+        # if app slug in is APP_ENABLED.
+        self.is_active = False
+
+    def load_content_types(self) -> None:
```
buxx: `abc` offer a modern way to set an method as abstract, see https://docs.python.org/3/library/abc.html#abc.abstractmethod

backend/tracim_backend/lib/core/application.py
inkhey: yes, circular import, i added a comment.

backend/tracim_backend/lib/core/application.py
inkhey: renamed to show_inactive + doc.

backend/tracim_backend/lib/utils/utils.py
buxx: This can be a class function of `ApplicationApi`

buxx: Or this function can stay a util function, but take module to "explore" as parameter.

inkhey: moved to tracim_backend.apps

backend/tracim_backend/lib/utils/app.py
inkhey: done whenever possible.

backend/tracim_backend/command/cleanup.py
inkhey: done.

backend/tracim_backend/command/cleanup.py
inkhey: Yes, we can as we load app a disabled_app.
 we can do something like: 

```python
app_lib = ApplicationApi(app_list=app_list, show_disabled=True)
if app_lib.exist(AGENDA__APP_SLUG):
    agenda_app = app_lib.get(AGENDA__APP_SLUG)
    if not agenda_app.is_active:
        agenda_app.load_config(self._app_config)
```

Not sure if it make sense to do agenda_app.check_config() too.

inkhey: if we do control config :
- we will get explicit error if config is uncorrect
- we will support only fully correct agenda config

if we do not control config:
- we will get implicit error if config is uncorrect.
- we will support all config with correct agenda dir parameter (if other parameter related to caldav are uncorrect, we do not check them).

backend/tracim_backend/app_models/applications.py
buxx: ok

backend/tracim_backend/config.py
lebouquetin: what do you mean by `I create a retrocompat code,`? You did write some code which allow to re-use these parameters? What is the behavior?

backend/tracim_backend/command/cleanup.py
lebouquetin: @inkhey could we have a chat about this?

backend/tracim_backend/config.py
lebouquetin: The good way to fix this:

- default values:
    - agenda actived
    - collabora not activated
- backward compatibility:
    - add a warning if parameters are present (but do not automatically add apps)

backend/tracim_backend/command/cleanup.py
inkhey: After discussion with @lebouquetin, we decide to not force agenda deletion if agenda app is no activated.

backend/tracim_backend/config.py

inkhey:
@PhilippeAccorsi check this:
https://github.com/tracim/tracim/pull/2618/commits/0b7c801cd7ffff4dfa584dba7610426dbdf24988
- As we replaced only commented list, if used changed default config file, no change will be done.
- We add agenda to list default list.

note: This can cause trouble as if we decide to change configuration after. for example: we active agenda and then we disable it (agenda will stay).

inkhey:
:heavy_check_mark: Tested on branch with dev env (pserve/waitress) and docker context.

Issue #941, opened 2018-09-20
Refactor Group system in a better way.


inkhey:
**Describe the bug**

Currently Tracim database model use a complex system for profile, with 3 tables (2 really used) : 
- group_permission (not really used , deprecated ?)
- groups (list of groups, name/slug, and display_name)
- users_groups (groups related to user)

The role system which is the system the most similar to groups, work itself only with one table:
- user_workspace (store only role level int, all other information like slug are store only in tracim code)

There are few issue with the group system : 

- groups slug info and id are stored in 2 differents places : in database and in code (check Group.TIM_USER_GROUPNAME for example). This mean that's a bit complex to change properly profiles name, because it need to modify both database internal data (not database schema) and stored var in code (see https://github.com/tracim/tracim/issues/937 for example).
It seems more convenient, like for role, to not store anything else than id/level in database and to have slug and associated information in code which allow easily to make code related to specific profile level, to return slug of a specific group, etc...

- Groups models allow users to be part of multiple group. This doesn't make any sense related to profiles system which choose always the "max" id group and profile checking, which rely on id value to know if Admin profile means also  manager and user right (other user-group junction than max are useless).

**(not yet) Implemented solution :**
proposition here : see https://github.com/tracim/tracim/pull/2668)
- Convert profile as Enum in users table. Remove completely "group" mecanism.
- table permission and group_permission are dropped.

lebouquetin:
@inkhey we had a chat with a major client who was interested in being able to define which features are available for each profile. Fixing this issue (ie remove the "technical debt") may be in conflict with implementing the definition of features available per profile

lebouquetin:
@PhilippeAccorsi @inkhey I'd like to have a chat together before we remove this

inkhey:
after discussion with @lebouquetin, we decide to merge the fix for this and we do need to clarify how to implement a more customizable mecanism later: https://github.com/tracim/tracim/issues/2879

buxx:
I approve the [pull request](https://github.com/tracim/tracim/pull/2668)

buxx:
Manuel tests done on branch: User (admin, trusted and user) can still been created and use their account.

PR #2668, opened 2020-02-04
Refactor/941  remove group system for profile

inkhey:
related to #941 

Convert profile as Enum in users table. Remove completely "group" mecanism.

## Checkpoints

**For code reviewer**

- [x] Code is clear enough

**For developer**

**For quality team**

- [x] Tested by quality team


inkhey:
Refactoring is completed, but we need discussion (@lebouquetin) about future usage of Tracim to be sure this fix is a good idea.

inkhey:
@lebouquetin About this refactoring, i suggest to merge this for Tracim 2.6.4 or 3.0 to have a cleaner code about profile. 
Customized profile/role with some kind of group feature may be cool but i do think we need lot of search to find a proper way to include this in tracim properly. 
Current code use a "Group" table that doesn't mean anything now.

What i see currently about a group feature:

1. Add a "function" field for user that match to function of the user, these function may be highly configurable and potentially mentionnable user (will just mention the user with the function which have access to tracim).
2. The same as before but with mecanism to: 
   - set default profile and workspace role to all user of a function at user creation/ add to function
   - ui to change easily all workspace role for all user with a specify function

this way we can show organization "role" (named function) without need to deal with role access in tracim.

We can of course imagine something more configurable but has we already have two level of right (per workspace and per user), i currently do no see a simple solution.

I suggest to create an issue with the beginning of thinking about this and merge this one.

lebouquetin:
After discussion with a community of cities, there are some chances that we need to be able to configure things. So the right way to avoid technical debt would be to allow the full interface to be parameterized according to dynamic profiles and permissions.

We need to chat about this (what hard stuff you see) before to decide for this merge request; my point of view is probably more about making things more dynamic than about simplifying things.

inkhey:
After discussion with @lebouquetin, we will merge this and create a new ticket about making things more configurable.

inkhey:
1. about automatic test: there have been updated to follow the refacto, but no new test has been added. current test automatic for both migration and endpoint rely now on refactorised code to ensure no regression.
2. about manual test: rapid manual test of tracim with profile, manual test of migration code to ensure it work correctly on different database.

buxx:
Manuel tests done: ready to merge.

Issue #2674, opened 2020-02-06
when you update file you not seen filename before validate since adding multi-upload


PhilippeAccorsi:
## Bug description & symptom

When you want to update file you need to select new file. After that you not seen anything about the file you have selected. You not seen the name of the file.

## Diagnostic

*Optional - explain what has been found and what is not yet clear. If unknown, write "unknown".*

## Expected

- [x] show name of selected file (like we have in tracim <= 2.5.1)

## How to reproduce

- Try to update one file

## Screenshots

**In tracim 2.6.0_rc1**

![file_selected_is_not_visible](https://user-images.githubusercontent.com/25848257/73933173-26d94680-48dc-11ea-88fe-cd707447c115.png)

**In tracim 2.5.1**

![file_selected_is_visible](https://user-images.githubusercontent.com/25848257/73933181-2b9dfa80-48dc-11ea-8923-eea6417a642e.png)

## Extra information

PhilippeAccorsi:
Tested and verified (wait merge on develop)

PR #2689, opened 2020-02-18
[For 2.6.1] Fix/2674 filename when updating file and no preview

metzgegu:
#2674 

## Checkpoints

- [x] Code is clear enough

- [x] Tested by quality team

### DIFF
--- a/frontend_lib/test/component/FileDropzone.spec.js
+++ b/frontend_lib/test/component/FileDropzone.spec.js
@@ -1,0 +1,81 @@
+import React from 'react'
+import { expect } from 'chai'
+import { mount } from 'enzyme'
+import FileDropzone from '../../src/component/FileDropzone/FileDropzone.jsx'
+import { FILE_PREVIEW_STATE } from '../../src/helper.js'
+import sinon from 'sinon'
+require('../../src/component/FileDropzone/FileDropzone.styl')
+
+describe('<FileDropZone />', () => {
+  const onDropCallBack = sinon.spy()
+  const onClickCallBack = sinon.spy()
+
+  const props = {
+    onDrop: onDropCallBack,
+    onClick: onClickCallBack,
+    multipleFiles: true,
+    filename: 'randomPreview',
+    preview: FILE_PREVIEW_STATE.NO_FILE
+  }
+
+  const wrapper = mount(
+    <FileDropzone
+      t={(key) => key}
+      {...props}
+    />
+  )
+
+  describe('when there is no file yet', () => {
+    before(() => {
+      wrapper.setProps({ preview: FILE_PREVIEW_STATE.NO_FILE })
+    })
+
+    after(() => {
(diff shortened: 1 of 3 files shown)


frontend_lib/test/component/FileDropzone.spec.js
```diff
@@ -0,0 +1,81 @@
...
+import { FILE_PREVIEW_STATE } from '../../src/helper.js'
+import sinon from 'sinon'
+require('../../src/component/FileDropzone/FileDropzone.styl')
+
+describe('<FileDropZone />', () => {
+  const onDropCallBack = sinon.spy()
+  const onClickCallBack = sinon.spy()
+
+  const props = {
+    onDrop: onDropCallBack,
```
Skylsmoi: if you don't do any tests on these spies, you don't need them.
Replace them by an empty function

frontend_lib/test/component/FileDropzone.spec.js
```diff
@@ -0,0 +1,81 @@
...
+    onClick: onClickCallBack,
+    multipleFiles: true,
+    filename: 'randomPreview',
+    preview: FILE_PREVIEW_STATE.NO_FILE
+  }
+
+  const wrapper = mount(
```
Skylsmoi: is mount required instead of shallow ?

metzgegu: Yes because the component FileDropezone is wrapped with a react-dropzone/Dropezone so if it use shallow, we don't have have access to what Dropzone render

metzgegu:
By the request of @PhilippeAccorsi, Tested and Verified on branch.

frontend_lib/test/component/FileDropzone.spec.js
Skylsmoi: ok. Add a comment explaining why you use mount()

Skylsmoi: Btw, FileDropzone isn't wrapped by Dropzone. It has a react-dropzone/Dropzone component that wrap the jsx inside it. But FileDropzone is the hole thing.
It is more accurate to say that the the react-dropzone/Dropzone is wrapped by our FileDropzone component.

frontend_lib/test/component/FileDropzone.spec.js
```diff
@@ -0,0 +1,87 @@
...
+      wrapper.setProps({ preview: props.preview })
+    })
+
+    it('should display the preview in a img tag', () => {
+      expect(wrapper.find('.filecontent__preview > img').prop('src')).to.equal(randomPreview)
+    })
+  })
+
+  describe('click on the dropzone', () => {
+    it('should call the onClick callback', () => {
```
Skylsmoi: if you test the onclick, also test the ondrop. I told you to replace the spy by an empty function because you didn't use the spy. I should have told you this comment was for both onclick and ondrop

PR #2718, opened 2020-03-04
[For 2.6.x] fix: remove requirement to run the Black pre-commit hook with python3.6

raphj:
Pre-commit fails on a fresh install of Debian Buster because the  Black precommit hook requires Python 3.6. However, The version of Python on Debian Buster is 3.7.

The requirement is here because Black requires Python 3.6+. By removing the requirement in .pre-commit-config.yaml, we should ensure that nothing breaks. My take is that Black will fail on Python 3.5 and before anyway so removing the requirement should be safe.
    
Related to #2737.

## Checkpoints

- [x] Code is clear enough

- [x] Tested by quality team


inkhey:
@raphj @PhilippeAccorsi Is something missing for merging this issue ?

PhilippeAccorsi:
I'm not sure but (i dont have tested this issue), but Black still working correctly on debian Stretch or older without this `language_version: python3.6` ? Message still explicit on stretch or older version ?

inkhey:
> I'm not sure but (i dont have tested this issue), but Black still working correctly on debian Stretch or older without this `language_version: python3.6` ? Message still explicit on stretch or older version ?

Black doesn't work on python 3.5, which mean precommit can cause issue for dev on old system, but that's a dev problem and do not break anything for normal usage of tracim. By the way, precommit is not required, it's just a tool to help us doing clean code that will not turn ci red almost instantly.

PhilippeAccorsi:
> > I'm not sure but (i dont have tested this issue), but Black still working correctly on debian Stretch or older without this `language_version: python3.6` ? Message still explicit on stretch or older version ?
> 
> Black doesn't work on python 3.5, which mean precommit can cause issue for dev on old system, but that's a dev problem and do not break anything for normal usage of tracim. By the way, precommit is not required, it's just a tool to help us doing clean code that will not turn ci red almost instantly.


So if one dev try to use this feature on stretch, what is the solution for him to know he need python 3.6 to use Black ? At present information are available in this file. If we remove this, where we can find information in tracim project ? (maybe just add documentation about this limitation).

inkhey:
> > > I'm not sure but (i dont have tested this issue), but Black still working correctly on debian Stretch or older without this `language_version: python3.6` ? Message still explicit on stretch or older version ?
> > 
> > 
> > Black doesn't work on python 3.5, which mean precommit can cause issue for dev on old system, but that's a dev problem and do not break anything for normal usage of tracim. By the way, precommit is not required, it's just a tool to help us doing clean code that will not turn ci red almost instantly.
> 
> So if one dev try to use this feature on stretch, what is the solution for him to know he need python 3.6 to use Black ? At present information are available in this file. If we remove this, where we can find information in tracim project ? (maybe just add documentation about this limitation).

@raphj and me just add some documentation about this.

PhilippeAccorsi:
Thats great for me, we not lost information we have spend time to find.

PhilippeAccorsi:
@raphj your last merge have lost many information added by @inkhey

raphj:
After discussion with @inkhey and @PhilippeAccorsi everything is correct. All needed information is here. Let's squash & merge this! :tada:

PR #2760, opened 2020-03-12
[For 2.6.x] Fix the size of rotated images in the gallery application

### LINKED ISSUE #2537: Rotated images have incorrect sizes in the gallery application
## Bug description & symptom

When rotating an image in the gallery, it does not occupy all the available space.

## Expected

After rotate an image on the gallery, it must be resized to take upall the available space.

## How to reproduce

1. Open gallery
2. Visualize image which can be rotated
3. Rotate it

## Screenshots

rotate

## Impacted components

*Optional - list components which will be impacted.*

- *frontend*

raphj:
Related to #2537.

## Checkpoints

- [x] Code is clear enough

- [x] Tested by quality team


frontend_app_gallery/src/component/MainPreview.jsx
```diff
@@ -2,32 +2,88 @@ import React from 'react'
...
-      height: img.height,
-      width: img.width,
       imageLoaded: true
     })
   }
 
-  render () {
-    const { props, state } = this
+  fixImageSize () {
+    if (this.props.rotationAngle % 180) {
```
Skylsmoi: from my understanding, since this function is always called on component update and if props.rotationAngle is different than 0 or 180, the computing in the requestAnimatinoFrame will happen on every "compoenentDidUpdate" until rotationAngle change.

Since componentDidUpdate can be fired a lot of time (eg. especially when it has async requests related to it) I'm a bit afraid of the javascript computation amount than can be required.
Also, ref in React are attached to the DOM, not the virtual DOM. So it cannot fully use React high performance related to the virtual DOM.

Is it possible to call ifxImageSize on componendDidUpdate only when rotationAngle change ?

If not, just do some performance check on the rotation and update process. I might be wrong.

Skylsmoi: Pm me about this if I'm not clear enough

frontend_app_gallery/src/component/MainPreview.jsx
```diff
@@ -2,32 +2,88 @@ import React from 'react'
...
     this.setState({
-      height: img.height,
-      width: img.width,
       imageLoaded: true
     })
   }
 
-  render () {
-    const { props, state } = this
+  fixImageSize () {
```
Skylsmoi: I would like a more precise name for this function

I'm not entirely sure about everything it does but I would call it something like fixImageRatioForRotation

frontend_app_gallery/src/component/MainPreview.jsx
metzgegu: Yes, It is possible to call `fixImageSize` only when the rotationAngle has changed, the size has to be changed only in this case

frontend_app_gallery/src/component/MainPreview.jsx
metzgegu: Yes this function name isn't precise enough,

I think `fixImageRatioForRotation` is a good name but I would prefer `compute` to `fix` because we don't really fix something that is broken.

frontend_app_gallery/test/component/MainPreview.spec.js
```diff
@@ -13,7 +13,12 @@ describe('<MainPreview />', () => {
     handleClickShowImageRaw
   }
 
-  const wrapper = shallow(<MainPreview {...props} t={tradKey => tradKey} />)
+  // INFO - GM - 2020/04/15 - Mock some window functions used by MainPreview
+  window.requestAnimationFrame = (cb) => { cb() }
+  window.getComputedStyle = () => { return {} }
```
Skylsmoi, line 18: I think it would fit best in test/setup.js
see  window.matchMedia for example

Also, since requestAnimationFrame is supposed not to answer immediately, wrap the cb() in a setTimeout of, I don't know, 10ms should be fine. It'll reflect better what is happening in browser

frontend_app_gallery/test/setup.js
```diff
@@ -32,4 +32,7 @@ window.matchMedia = window.matchMedia || function() {
     removeListener: function() {}
   }
 }
+// INFO - GM - 2020/04/15 - Mock some window functions used by MainPreview
+window.requestAnimationFrame = (cb) => setTimeout(cb, 10)
+window.getComputedStyle = () => ({})
```
Skylsmoi, line 37: as I said, use the matchMedia example to avoid overriding the property if it isn't required

gbonaspetti:

raphj:
Thanks @Skylsmoi and @metzgegu, and sorry, I did not see this discussion.

Skylsmoi:
> Thanks @Skylsmoi and @metzgegu, and sorry, I did not see this discussion.

don't worry, you where on a different task, it was up to us to complete the pr

Issue #2693, opened 2020-02-19
Creating a workspace do not scroll to show the new workspace in sidebar


metzgegu:
## Bug description & symptom

When a new workspace is created, the workspace is opened in the sidebar but it doesn't scroll to this new workspace, it is a problem when there is a lot of different workspaces

## Diagnostic

- Sidebar item need to have html `id` in order to scroll to it

## Expected

- [x] scroll directly to created workspace in the sidebar when you creating it

## How to reproduce

- Create a workspace in a instance which have already a lot of workspaces

metzgegu:
Implemented solution is to add a id to each workspace `li` in sidebar, in order to scroll into the workspace which is open when the the workspaceList is loaded.

gbonaspetti:
Verified on branch (wait merge to close).

PR #2781, opened 2020-03-23
[for 2.6.x] Fix/2693 Scroll to workspace created in Sidebar

metzgegu:
#2693

## Checkpoints

**For code reviewer**

- [x] Code is clear enough

**For developper**

**For quality team**

- [x] Tested by quality team

### DIFF
--- a/functionnal_tests/cypress/integration/workspace/create_workspace_spec.js
+++ b/functionnal_tests/cypress/integration/workspace/create_workspace_spec.js
@@ -25,19 +25,19 @@ describe('Workspace', () => {
       .should('exist')
   })
 
+  const createOneWorkspace = (cy, workspaceName) => {
+    return cy.get('[data-cy="sidebarCreateWorkspaceBtn"]').should('be.visible').click()
+      .get('[data-cy="createcontent__form__input"]').should('be.visible').type(workspaceName)
+      .get('[data-cy="popup__createcontent__form__button"]').should('be.visible').click()
+  }
+
+  const getWorkspaceItemByName = (workspaceTitle) => (
+    cy.get(`.sidebar__content__navigation__workspace__item__name[title="${workspaceTitle}"]`)
+      .parents('li.sidebar__content__navigation__workspace__item')
+  )
+
   describe('Dashboard', () => {
     describe('Creating two new workspaces', () => {
-      const createOneWorkspace = (cy, workspaceName) => {
-        return cy.get('[data-cy="sidebarCreateWorkspaceBtn"]').should('be.visible').click()
-          .get('[data-cy="createcontent__form__input"]').should('be.visible').type(workspaceName)
-          .get('[data-cy="popup__createcontent__form__button"]').should('be.visible').click()
-      }
-
-      const getWorkspaceItemByName = (workspaceTitle) => (
-        cy.get(`.sidebar__content__navigation__workspace__item__name[title="${workspaceTitle}"]`)
-          .parents('li.sidebar__content__navigation__workspace__item')
-      )
-
       it('should display the new workspaces properly with the right workspace opened in the sidebar', () => {
         const workspaceTitle1 = 'second workspace'
         const workspaceTitle2 = 'third workspace'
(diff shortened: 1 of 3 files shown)


functionnal_tests/cypress/integration/workspace/create_workspace_spec.js
```diff
@@ -25,19 +25,19 @@ describe('Workspace', () => {
...
-    describe('Creating two new workspaces', () => {
-      const createOneWorkspace = (cy, workspaceName) => {
-        return cy.get('[data-cy="sidebarCreateWorkspaceBtn"]').should('be.visible').click()
-          .get('[data-cy="createcontent__form__input"]').should('be.visible').type(workspaceName)
-          .get('[data-cy="popup__createcontent__form__button"]').should('be.visible').click()
-      }
+  const createOneWorkspace = (cy, workspaceName) => {
+    return cy.get('[data-cy="sidebarCreateWorkspaceBtn"]').should('be.visible').click()
+      .get('[data-cy="createcontent__form__input"]').should('be.visible').type(workspaceName)
+      .get('[data-cy="popup__createcontent__form__button"]').should('be.visible').click()
```
Skylsmoi: if you are not inside a test, don't use should
Use should only if you are in an `it()`

functionnal_tests/cypress/integration/workspace/create_workspace_spec.js
```diff
@@ -62,4 +62,26 @@ describe('Workspace', () => {
       })
     })
   })
+
+  describe('Creating a workspace while a lot of workspace is already created', () => {
```
Skylsmoi:
```suggestion
  describe('Creating a workspace while a lot of workspace are already created', () => {
```

Skylsmoi:
@PhilippeAccorsi When you will be testing this PR, check my review comment.
I think it'll be nice if you do some checks on that question too.

functionnal_tests/cypress/integration/workspace/create_workspace_spec.js
metzgegu: I remove the 3 `should` because it was useless to check if it was visible before do an action to the element selected, but here the const `createOneWorkspace` is only used in test case with it() so is it still invalid ?

metzgegu:
> The implementation is good.
> One question though, the loadWorkspace() function is called by the custom event handler for the custom event REFRESH_WORKSPACE_LIST.
> this event isn't called by many places, only app workspace advanced.
> I think it can lead to unexpected behavior.
> Be sure to check these cases, I think it might only create some scroll when editing a workspace from that app which wont be a big issue. But we need to know exactly.

Yes you are right, so I moved the scroll command to only scroll to workspace in sidebar when a workspace was created eg. after loading workspace data in REFRESH_WORKSPACE_LIST

functionnal_tests/cypress/integration/workspace/create_workspace_spec.js
```diff
@@ -25,19 +25,19 @@ describe('Workspace', () => {
...
+    return cy.get('[data-cy="sidebarCreateWorkspaceBtn"]').click()
+      .get('[data-cy="createcontent__form__input"]').type(workspaceName)
+      .get('[data-cy="popup__createcontent__form__button"]').click()
+  }
 
-      const getWorkspaceItemByName = (workspaceTitle) => (
-        cy.get(`.sidebar__content__navigation__workspace__item__name[title="${workspaceTitle}"]`)
-          .parents('li.sidebar__content__navigation__workspace__item')
-      )
+  const getWorkspaceItemByName = (workspaceTitle) => (
```
Skylsmoi: give cy as a parameter as you did for createOneWorkspace.
It's better to have functions that don't use resources from outside of themselves

functionnal_tests/cypress/integration/workspace/create_workspace_spec.js
Skylsmoi: what do you mean by "invalid" ?
Is it that the function is only called once ? If yes, it's not a problem. If later we need this function, it'll be easier to refactor it at a higher level

Skylsmoi:
> > The implementation is good.
> > One question though, the loadWorkspace() function is called by the custom event handler for the custom event REFRESH_WORKSPACE_LIST.
> > this event isn't called by many places, only app workspace advanced.
> > I think it can lead to unexpected behavior.
> > Be sure to check these cases, I think it might only create some scroll when editing a workspace from that app which wont be a big issue. But we need to know exactly.
> 
> Yes you are right, so I moved the scroll command to only scroll to workspace in sidebar when a workspace was created eg. after loading workspace data in REFRESH_WORKSPACE_LIST

I'm not sure to understand, with your most recent changes, the scroll will still be done on every custom event REFRESH_WORKSPACE_LIST
I don't see how it'll be called only after a workspace creation

metzgegu:
> > > The implementation is good.
> > > One question though, the loadWorkspace() function is called by the custom event handler for the custom event REFRESH_WORKSPACE_LIST.
> > > this event isn't called by many places, only app workspace advanced.
> > > I think it can lead to unexpected behavior.
> > > Be sure to check these cases, I think it might only create some scroll when editing a workspace from that app which wont be a big issue. But we need to know exactly.
> > 
> > 
> > Yes you are right, so I moved the scroll command to only scroll to workspace in sidebar when a workspace was created eg. after loading workspace data in REFRESH_WORKSPACE_LIST
> 
> I'm not sure to understand, with your most recent changes, the scroll will still be done on every custom event REFRESH_WORKSPACE_LIST
> I don't see how it'll be called only after a workspace creation

With my most recent changes the scroll will be done on custom event REFRESH_WORKSPACE_LIST only if `data.openInSidebarId` is defined eg. when refreshing the workspace list and a workspace need to be open in sidebar, in fact currently only `PopupCreateWorkspace.jsx` use `data.openInSidebarId` when dispatching the custom event REFRESH_WORKSPACE_LIST

gbonaspetti:

Issue #2793, opened 2020-03-27
smtp error on implicit connection with port 465 visible on AWSapps


PhilippeAccorsi:
## Bug description & symptom

With awsapps smtp service, amazon use port 465 with implicit connection. When we try to send email, nothing append

## Diagnostic

*Optional - explain what has been found and what is not yet clear. If unknown, write "unknown".*

## Expected

- [x] Tracim can send email with implicit configuration
- [x] add parameter in config file to use SSL and/or STARTTLS with SMTP

## Context and version information

All Tracim version.

## How to reproduce

Try to send email with awsapps smtp service on port 465

## Impacted components

- *backend*

PhilippeAccorsi:
I not see modification in env variable list for this settings "email.notification.smtp.implicit_ssl" (settings.md documentation).

PhilippeAccorsi:
Tested and verified on branch.

Need quick test with docker image on develop branch.

PhilippeAccorsi:
Tested with docker image.

PR #2795, opened 2020-03-27
[For 2.6.3] fix(backend): add smtp implicit encryption support

inkhey:
add a boolean config parameter `email.notification.use_implicit_ssl` to active implicit encryption for smtp (ssl over smtp, most of the time on port 465) and code needed to support this feature.

related to #2793 

## Checkpoints

**For code reviewer**

- [x] Code is clear enough

**For developper**

**For quality team**

- [x] Tested by quality team

### DIFF
--- a/backend/development.ini.sample
+++ b/backend/development.ini.sample
@@ -317,6 +317,10 @@ email.notification.smtp.server = your_smtp_server
 email.notification.smtp.port = 25
 email.notification.smtp.user = your_smtp_user
 email.notification.smtp.password = your_smtp_password
+## active implicit ssl if you are using implicit smtp with encryption port like 465
+## by default, tracim will try to use explicit smtp encryption using starttls, and unencrypted
+## connection as fallback.
+email.notification.smtp.implicit_ssl = false
 
 ### Headers ###
 email.notification.from.default_label = Tracim Notifications
--- a/backend/development.ini.sample
+++ b/backend/development.ini.sample
@@ -317,9 +317,9 @@ email.notification.smtp.server = your_smtp_server
 email.notification.smtp.port = 25
 email.notification.smtp.user = your_smtp_user
 email.notification.smtp.password = your_smtp_password
-## active implicit ssl if you are using implicit smtp with encryption port like 465
-## by default, tracim will try to use explicit smtp encryption using starttls, and unencrypted
-## connection as fallback.
+# active implicit ssl if you are using implicit smtp with encryption port like 465
+# by default, tracim will try to use explicit smtp encryption using starttls, and unencrypted
+# connection as fallback.
 email.notification.smtp.use_implicit_ssl = false
 
 ### Headers ###
--- a/backend/development.ini.sample
+++ b/backend/development.ini.sample
@@ -320,7 +320,7 @@ email.notification.smtp.password = your_smtp_password
 ## active implicit ssl if you are using implicit smtp with encryption port like 465
 ## by default, tracim will try to use explicit smtp encryption using starttls, and unencrypted
 ## connection as fallback.
-email.notification.smtp.implicit_ssl = false
+email.notification.smtp.use_implicit_ssl = false
 
 ### Headers ###
 email.notification.from.default_label = Tracim Notifications
--- a/backend/doc/setting.md
+++ b/backend/doc/setting.md
@@ -311,7 +311,7 @@ to enable mail notification, smallest config is this:
     ## active implicit ssl if you are using implicit smtp with encryption port like 465
     ## by default, tracim will try to use explicit smtp encryption using starttls, and unencrypted
     ## connection as fallback.
-    email.notification.smtp.implicit_ssl = false
+    email.notification.smtp.use_implicit_ssl = false
 
 Don't forget to set `website.base_url` and `website.title` for the frontend, as some features use them to
 link the frontend in emails.
--- a/backend/doc/setting.md
+++ b/backend/doc/setting.md
@@ -4,7 +4,7 @@
 
 Most settings in Tracim are configurable using both the INI configuration file and environment variables.
 
-### Tracim 2.2 fully supported variables:
+### Tracim fully supported variables:
 
 You can set those parameters in INI configuration file (see `config_file_name`) or
 environnement variable (see `env_var_name`).
@@ -14,8 +14,8 @@ The priority order is (from less to more priority):
 - configuration file
 - environnement variables
 
-|<env_var_name>|<config_file_name>|<config_name>|
-|--------------|------------------|-------------|
+|<env_var_name>                |<config_file_name>            |<config_name>                 |
+|------------------------------|------------------------------|------------------------------|
 |TRACIM_APP__ENABLED           |app.enabled                   |APP__ENABLED                  |
 |TRACIM_SQLALCHEMY__URL        |sqlalchemy.url                |SQLALCHEMY__URL               |
 |TRACIM_DEFAULT_LANG           |default_lang                  |DEFAULT_LANG                  |
@@ -70,6 +70,7 @@ The priority order is (from less to more priority):
 |TRACIM_EMAIL__NOTIFICATION__SMTP__PORT|email.notification.smtp.port  |EMAIL__NOTIFICATION__SMTP__PORT|
 |TRACIM_EMAIL__NOTIFICATION__SMTP__USER|email.notification.smtp.user  |EMAIL__NOTIFICATION__SMTP__USER|
 |TRACIM_EMAIL__NOTIFICATION__SMTP__PASSWORD|email.notification.smtp.password|EMAIL__NOTIFICATION__SMTP__PASSWORD|
+|TRACIM_EMAIL__NOTIFICATION__SMTP__USE_IMPLICIT_SSL|email.notification.smtp.use_implicit_ssl|EMAIL__NOTIFICATION__SMTP__USE_IMPLICIT_SSL|
 |TRACIM_EMAIL__REPLY__ACTIVATED|email.reply.activated         |EMAIL__REPLY__ACTIVATED       |
 |TRACIM_EMAIL__REPLY__IMAP__SERVER|email.reply.imap.server       |EMAIL__REPLY__IMAP__SERVER    |
 |TRACIM_EMAIL__REPLY__IMAP__PORT|email.reply.imap.port         |EMAIL__REPLY__IMAP__PORT      |
--- a/backend/doc/setting.md
+++ b/backend/doc/setting.md
@@ -308,6 +308,10 @@ to enable mail notification, smallest config is this:
     email.notification.smtp.port = 1025
     email.notification.smtp.user = test_user
     email.notification.smtp.password = just_a_password
+    ## active implicit ssl if you are using implicit smtp with encryption port like 465
+    ## by default, tracim will try to use explicit smtp encryption using starttls, and unencrypted
+    ## connection as fallback.
+    email.notification.smtp.implicit_ssl = false
 
 Don't forget to set `website.base_url` and `website.title` for the frontend, as some features use them to
 link the frontend in emails.
--- a/backend/doc/setting.md
+++ b/backend/doc/setting.md
@@ -309,9 +309,9 @@ to enable mail notification, smallest config is this:
     email.notification.smtp.port = 1025
     email.notification.smtp.user = test_user
     email.notification.smtp.password = just_a_password
-    ## active implicit ssl if you are using implicit smtp with encryption port like 465
-    ## by default, tracim will try to use explicit smtp encryption using starttls, and unencrypted
-    ## connection as fallback.
+    # active implicit ssl if you are using implicit smtp with encryption port like 465
+    # by default, tracim will try to use explicit smtp encryption using starttls, and unencrypted
+    # connection as fallback.
     email.notification.smtp.use_implicit_ssl = false
 
 Don't forget to set `website.base_url` and `website.title` for the frontend, as some features use them to
--- a/backend/tracim_backend/applications/share/lib.py
+++ b/backend/tracim_backend/applications/share/lib.py
@@ -114,6 +114,7 @@ class ShareLib(object):
             config.EMAIL__NOTIFICATION__SMTP__PORT,
             config.EMAIL__NOTIFICATION__SMTP__USER,
             config.EMAIL__NOTIFICATION__SMTP__PASSWORD,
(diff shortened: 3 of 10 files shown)


backend/tracim_backend/applications/share/lib.py
```diff
@@ -164,7 +165,7 @@ def get_content_share(self, content: Content, share_id: int) -> ContentShare:
                 .filter(ContentShare.content_id == content.content_id)
                 .filter(ContentShare.share_id == share_id)
                 .one()
-            )  # type: ContentShare
```
inkhey: this typing is not needed and create weird issue using black (AST related issue), that's the reason why i drop it.

lebouquetin: @inkhey this kind of comment must be in the code. No chance to read this MR when you are in the code

inkhey: This comment is just for reviewer to understand the diff, it make no sense to have this in code because it's clearly related to a bug in black, knowing that uneeded typing here create bug in a specific black version is not really helpful for dev.

lebouquetin: I disagree because a developer may see the code, then see there is no typing and would try to add it again. So something like: "do not add typing here - there is a bug in black https://www.github.com/black/back/xxx

backend/tracim_backend/lib/mail_notifier/sender.py
```diff
@@ -63,12 +63,17 @@ def connect(self):
         if not self._smtp_connection:
             log = "Connecting to SMTP server {}"
             logger.info(self, log.format(self._smtp_config.server))
-            # TODO - G.M - 2019-01-29 - Support for SMTP SSL-only port connection
-            # using smtplib.SMTP_SSL
-            self._smtp_connection = smtplib.SMTP(self._smtp_config.server, self._smtp_config.port)
+            if self._smtp_config.implicit_ssl:
```
lebouquetin, line 66: why do you name it `IMPLICIT_SSL` while the smtplib name it `SMTP_SSL`

backend/tracim_backend/lib/mail_notifier/utils.py
```diff
@@ -1,11 +1,12 @@
 class SmtpConfiguration(object):
     """Container class for SMTP configuration used in Tracim."""
 
-    def __init__(self, server: str, port: int, login: str, password: str):
+    def __init__(self, server: str, port: int, login: str, password: str, implicit_ssl: bool):
```
lebouquetin, line 4: its not intuitive that implicit_ssl is a boolean. it should be use_implicit_ssl or something like this

lebouquetin:
some remarks on the code, plus:

- I'm not sure that implicit_ssl is the right name
- is this new parameter available through environment variable?
- if so, does it need some update on the documentation?

backend/tracim_backend/lib/mail_notifier/sender.py
inkhey: Because there is explicit and implicit ssl, we can use ssl/tls in explicit mode (aka startssl) and it's the normalized method.

inkhey: i check flask plugin which does equivalent mecanism and the option name are worst : "use_ssl", "use_tls" (tls for startssl and ssl for encrypted smtp (in fact ssl is not used anymore and all tls is used anywhere for encryption)

backend/tracim_backend/lib/mail_notifier/utils.py
inkhey: i agree.

backend/tracim_backend/lib/mail_notifier/sender.py
inkhey: default tracim behaviour is trying explicit tls encryption using starttls if we do use a new option name "use_ssl" it will make thing that defaut tracim behavior is unencrypted which is not true.

By the way, we probably, should make defaut behavior configurable to avoid "uncorrect unencrypted connexion when we expect starttls but this is out of the scope of this fix and it's complicated to find a solution which either does work and do not break old behaviour.

PhilippeAccorsi:
I not see modification in env variable list for this settings "email.notification.smtp.implicit_ssl" (settings.md documentation).

backend/tracim_backend/applications/share/lib.py
inkhey: fixed by: 
- moving typing to a working place
- creating an issue in black : https://github.com/psf/black/issues/1329
- adding comment about black issue

backend/tracim_backend/lib/mail_notifier/utils.py
```diff
@@ -1,11 +1,12 @@
 class SmtpConfiguration(object):
     """Container class for SMTP configuration used in Tracim."""
 
-    def __init__(self, server: str, port: int, login: str, password: str):
+    def __init__(self, server: str, port: int, login: str, password: str, use_implicit_ssl: bool):
```
lebouquetin, line 4: I suggest a default value to False + a docsting explaining what means use_implicit_ssl

lebouquetin:
Almost ready, I'm just not sure about the `SmtpConfiguration.__init__(...)` signature

backend/tracim_backend/lib/mail_notifier/utils.py
lebouquetin: not required at the moment, actually: a refactor will arrive soon.

PhilippeAccorsi:
@inkhey when I tried to create new account with port 465 without added smtp implicit I can see this error in backend log (after 4min) `Exception NotificationSendingFailed occured, return 500 http_code : Notification for new created account can't be send (SMTP error), new account creation aborted`. 

Why we abord creation of account ?

At present we create all the time new account, also if email is not sending. Whis this error, I can't see anything in frontend to know problem exist. I see this error in devtools :
~~~
message | Notification for new created account can't be send (SMTP error), new account creation aborted
-- | --
details | {…}
error_detail | {}
code | 2045
~~~

backend/development.ini.sample
```diff
@@ -317,6 +317,10 @@ email.notification.smtp.server = your_smtp_server
 email.notification.smtp.port = 25
 email.notification.smtp.user = your_smtp_user
 email.notification.smtp.password = your_smtp_password
+## active implicit ssl if you are using implicit smtp with encryption port like 465
+## by default, tracim will try to use explicit smtp encryption using starttls, and unencrypted
+## connection as fallback.
```
PhilippeAccorsi, line 322:
```suggestion
# active implicit ssl if you are using implicit smtp with encryption port like 465
# by default, tracim will try to use explicit smtp encryption using starttls, and unencrypted
# connection as fallback.
```

backend/doc/setting.md
```diff
@@ -308,6 +309,10 @@ to enable mail notification, smallest config is this:
     email.notification.smtp.port = 1025
     email.notification.smtp.user = test_user
     email.notification.smtp.password = just_a_password
+    ## active implicit ssl if you are using implicit smtp with encryption port like 465
+    ## by default, tracim will try to use explicit smtp encryption using starttls, and unencrypted
+    ## connection as fallback.
```
PhilippeAccorsi, line 314:
```suggestion
    # active implicit ssl if you are using implicit smtp with encryption port like 465
    # by default, tracim will try to use explicit smtp encryption using starttls, and unencrypted
    # connection as fallback.
```

backend/doc/setting.md
```diff
@@ -4,7 +4,7 @@
 
 Most settings in Tracim are configurable using both the INI configuration file and environment variables.
 
-### Tracim 2.2 fully supported variables:
+### Tracim fully supported variables:
```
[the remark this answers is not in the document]
inkhey: I do remove the 2.2 as it's not true. We update frequently the liste but we do not see that was marked as 2.2. The only way i see to see the true 2.2 parameters is checking the file at revision tagged 2.2.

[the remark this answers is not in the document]
inkhey: Not sure what we should do here.

inkhey:
> @inkhey when I tried to create new account with port 465 without added smtp implicit I can see this error in backend log (after 4min) `Exception NotificationSendingFailed occured, return 500 http_code : Notification for new created account can't be send (SMTP error), new account creation aborted`.
> 
> Why we abord creation of account ?
> 
> At present we create all the time new account, also if email is not sending. Whis this error, I can't see anything in frontend to know problem exist. I see this error in devtools :
> 
> ```
> message | Notification for new created account can't be send (SMTP error), new account creation aborted
> -- | --
> details | {…}
> error_detail | {}
> code | 2045
> ```

At present:
- if smtp failed for "Bad recipient", we accept to create the account and we return a warning in log.
- else if smtp failed for another reason, we abord the user creation, as we don't know what is happening (aka something is broken), so we return a 500 which mean "server as a problem, please contact administrator".

PhilippeAccorsi:
>     * else if smtp failed for another reason, we abord the user creation, as we don't know what is happening (aka something is broken), so we return a 500 which mean "server as a problem, please contact administrator".

Ok but I think its not user freindly if in frontend we dont know what is going on, on first 4min (no loader visible before error with sync parameter to send email). Also what is going on if I'm on async ? What is visible for user ? account is created or not ?

I understand is not good to create account is this specific case, but we need to add something to be sure is clear for admin user.
Can you check whis frontend dev to know what is possible to do to solve this problem ?

backend/doc/setting.md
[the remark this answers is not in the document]
PhilippeAccorsi: > I do remove the 2.2 as it's not true. We update frequently the liste but we do not see that was marked as 2.2. The only way i see to see the true 2.2 parameters is checking the file at revision tagged 2.2.

You right, I understand. Maybe we can just add information about "env variable feature is added in Tracim 2.2, all variable in this list are available in last Tracim version. Look in history if you need to know if some env variable are available in your Tracim version"

Something like that.

@buxx what do you think abbout that ?

inkhey:
> > ```
> > * else if smtp failed for another reason, we abord the user creation, as we don't know what is happening (aka something is broken), so we return a 500 which mean "server as a problem, please contact administrator".
> > ```
> 
> Ok but I think its not user freindly if in frontend we dont know what is going on, on first 4min (no loader visible before error with sync parameter to send email). Also what is going on if I'm on async ? What is visible for user ? account is created or not ?

With async, this should fail asynchronously, so account will be created, but mail will fail.

> I understand is not good to create account is this specific case, but we need to add something to be sure is clear for admin user.
> Can you check whis frontend dev to know what is possible to do to solve this problem ?

@Skylsmoi , @raphj It is complicated to handle this error code properly ?

PhilippeAccorsi:
> > > ```
> > > * else if smtp failed for another reason, we abord the user creation, as we don't know what is happening (aka something is broken), so we return a 500 which mean "server as a problem, please contact administrator".
> > > ```
> > 
> > 
> > Ok but I think its not user freindly if in frontend we dont know what is going on, on first 4min (no loader visible before error with sync parameter to send email). Also what is going on if I'm on async ? What is visible for user ? account is created or not ?
> 
> With async, this should fail asynchronously, so account will be created, but mail will fail.

OK
> 
> > I understand is not good to create account is this specific case, but we need to add something to be sure is clear for admin user.
> > Can you check whis frontend dev to know what is possible to do to solve this problem ?
> 
> @Skylsmoi , @raphj It is complicated to handle this error code (2045) properly ?

Error appear after 5min, so user need something in frontend to know something working in backend.

lebouquetin:
@inkhey would it be possible to check SMTP connexion at start time? This would allow to not run if SMTP config is not consistent

inkhey:
> @inkhey would it be possible to check SMTP connexion at start time? This would allow to not run if SMTP config is not consistent

1. It's possible to do some simple connexion check at startup for SMTP for both sync and async case (we can check for async synchronously at startup, it make sense).
2. This can not handle all case, as SMTP can fail after the first run.
3. It really unclear how such code should be done in tracim. We do have CFG object, but not an "initiator" class (we need to do SMTP check for all process running tracim, so one check with web tracim, one for webdav, etc…)

raphj:
@inkhey asked my point of view from an UX perspective, probably @celiamgtt should weigh in but here is my take: if a sending failure can be detected immediately in the backend, this failure should be shown in the frontend. Some mail-related failures cannot be detected immediately but this is ok (it can take a few days to send a mail).

The UI should show something like "A mail has been sent to `***@***.**`. You should receive it in a few seconds. Please check your spam folder. Click here to send the mail again.".

Moreover, it should be possible to fix the mail address if the user notices that their email address is wrong.

inkhey:
@PhilippeAccorsi The interresting part of log , to clarify what is happening, after few minutes:

```
  File "/home/user/tracim/backend/tracim_backend/lib/mail_notifier/sender.py", line 67, in connect
    self._smtp_connection = smtplib.SMTP(self._smtp_config.server, self._smtp_config.port)
  File "/usr/lib/python3.6/smtplib.py", line 251, in __init__
    (code, msg) = self.connect(host, port)
  File "/usr/lib/python3.6/smtplib.py", line 338, in connect
    (code, msg) = self.getreply()
  File "/usr/lib/python3.6/smtplib.py", line 394, in getreply
    raise SMTPServerDisconnected("Connection unexpectedly closed")
smtplib.SMTPServerDisconnected: Connection unexpectedly closed
```

backend/doc/setting.md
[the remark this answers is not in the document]
PhilippeAccorsi: I open issue about improvement of documentation. We need to add version of Tracim where we can use each env variable (at present we not explain if all env variable are available for each version or not)

PhilippeAccorsi:
Decision are:

- we add information in changelog about know issue with smtp implicit without implicit_ssl parameter activated and send mail in sync, and add link to documentation `/backend/doc/setting.md` about what we make in next release.
  - In next release we add a simple script to test smtp configuration (script started by sys admin).   Idea: if you want to make sure your smtp config working good, you can start this script and know if your smtp config working (without starting Tracim)
- issue (feature) is opened (#2822 ) about: not close automatically the flash message on error 500
- issue is opened (#2846 ) about: required version of Tracim to use each environnement variable

backend/tracim_backend/lib/mail_notifier/sender.py
lebouquetin: we need a new issue for refactor of these parameters in next major version. The point is ok for now

PhilippeAccorsi: Issue #2851 created about that

PhilippeAccorsi:

Issue #2406, opened 2019-09-17
Rich text button displayed in apps is confusing


celiamgtt:
## Bug description & symptom

Rich text button displayed in apps is confusing because it has the same level of hierarchy as the 'send' or 'cancel' buttons. 

## Expected

To discuss: display a link "text editing in advanced mode" below the comment

## Screenshots

![image](https://user-images.githubusercontent.com/50453784/65042600-7634cf80-d959-11e9-962f-e228a3995866.png)

metzgegu:
The implemented solution in my PR is to transform the `advancedEdition` button to a clickable div

gbonaspetti:
The label and the size of the button are different from that shown on the issue screenshot.

Currently:
![Screenshot from 2020-04-07 16-16-07](https://user-images.githubusercontent.com/35422692/78681801-7bee0480-78ed-11ea-9b41-cb789c40a830.png)

gbonaspetti:
Besides the label and size changes to make the button more similar to the one presented in the issue, I changed the code that made this same button hidden in small interfaces (the commit date was 2 years ago) and this for me didn't make sense with the current interface anymore. @Skylsmoi and @celiamgtt I'm open to discussion if you think it still makes sense to hide the button.

celiamgtt:
For me, it's not mandatory to display it (because it's an advance feature, and I'm not sure it will be useful on mobile), but it could be interesting for 'pro' users.

Skylsmoi:
we hid it because on mobile it is complicated to use tinymce. The screen gets too small when the mobile's keyboard appears. So we disabled the feature to keep things easy for users
Whether we should enable it should be discussed in another issue so I'm commenting your PR to roll back that change

lebouquetin:
@Skylsmoi I'm not sure what you want to rollback @Skylsmoi . If the button is about to re-appear in mobile, then the description of the issue must be updated.

I'm not sure there is good reason for having this feature on small screen because it is not really usable.

gbonaspetti:
@lebouquetin rollback the changes I had made to reappear the button. Back to the way it was.

Skylsmoi:
@lebouquetin in the related PR, @gbonaspetti removed the "display: none" on mobile. This what I want to rollback because the issue wasn't about this fix

buxx:
#2800 tested on it's branch: ok

PR #2800, opened 2020-03-30
[for 2.6.x] fix/2406 Change style to rich text button displayed in apps which is confusing

metzgegu:
#2406 

## Checkpoints

**For code reviewer**

- [x] Code is clear enough

**For developper**

**For quality team**

- [x] Tested by quality team

### DIFF
--- a/frontend_lib/src/component/Timeline/Timeline.jsx
+++ b/frontend_lib/src/component/Timeline/Timeline.jsx
@@ -138,25 +138,21 @@ class Timeline extends React.Component {
 
             <div className={classnames(`${props.customClass}__texteditor__wrapper`, 'timeline__texteditor__wrapper')}>
               <div className={classnames(`${props.customClass}__texteditor__advancedtext`, 'timeline__texteditor__advancedtext')}>
-                <button
-                  type='button'
+                <div
                   className={classnames(
-                    `${props.customClass}__texteditor__advancedtext__btn timeline__texteditor__advancedtext__btn btn outlineTextBtn`
+                    `${props.customClass}__texteditor__advancedtext__btn timeline__texteditor__advancedtext__btn`
                   )}
-                  onClick={props.onClickWysiwygBtn}
-                  disabled={props.disableComment}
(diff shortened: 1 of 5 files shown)


frontend_lib/src/component/Timeline/Timeline.jsx
```diff
@@ -138,25 +138,21 @@ class Timeline extends React.Component {
 
             <div className={classnames(`${props.customClass}__texteditor__wrapper`, 'timeline__texteditor__wrapper')}>
               <div className={classnames(`${props.customClass}__texteditor__advancedtext`, 'timeline__texteditor__advancedtext')}>
-                <button
-                  type='button'
+                <div
```
Skylsmoi, line 141: A button tag most important thing is that it is something clickable.
It's design is secondary.
Semantically, it's better to keep a button tag and adapt it's design if required.
I think it would be best to keep the button tag and remove the .btn and .outlineTextBtn class names

Skylsmoi:
about your last commit message, it's not about consistency, it's about semantic.
Consistency is doing the things the same way everywhere
Semantic is using the right tools for a given thing

gbonaspetti:
Besides the label and size changes to make the button more similar to the one presented in the issue, I changed the code that made this same button hidden in small interfaces (the commit date was 2 years ago) and this for me didn't make sense with the current interface anymore. @Skylsmoi and @celiamgtt I'm open to discussion if you think it still makes sense to hide the button.

frontend_lib/src/component/Timeline/Timeline.styl
```diff
@@ -104,6 +119,3 @@ columnTimeline()
     &__messagelist
       &__version__date
         font-size 15px
-    &__texteditor__advancedtext__btn
```
Skylsmoi, line 107: keep that for now, if you think it is not a good idea to disable it on mobile, you can create a new issue about it

gbonaspetti: Done.

Issue #697, opened 2018-07-25
Optimize Tracim Backend SQL requests


inkhey:
## Feature description

Some endpoint expectially, read status endpoint are not optimal, we do need to optimize them in order to reduce:
- the time to wait for receiving information
- the number of query (less work on database)

### Previous issue content:

**Currently:**
SQL request used by read_status endpoint, see https://github.com/tracim/tracim/blob/413085f49d1ae5930b6251fb6022fbde8b16a7f6/backend/tracim_backend/lib/core/content.py#L1291
is not optimal.

**Expected:**
Convert this database query in order to be more effective.

## Implementation
- not implemented yet
- see https://github.com/tracim/tracim/issues/697#issuecomment-611929405 for implementation proposal

lebouquetin:
@inkhey this issue is not clear, please complete the description

Skylsmoi:
front related issue: https://github.com/tracim/tracim/issues/1443

If the resolution of this issue is long/complicated, prioritize the frontend one

inkhey:
Some thought about fixing this :
- Try in raw sql first.
- Thing about using CTE query.

inkhey:
#2803 propose many database optimization, that don't directly change specific read_status part, but make read_status and many other query must faster.

inkhey:
About database support : 

Simple table of support of  recursive CTE + Debian and django 3.0 (decembre 2019) support for these database, in order to make decision on which database to retain:

| Database   | Version    | Official Debian support (debian official repos)                                   | Support recursive CTE | official support Django 3\.0 |
|------------|------------|------------------------------------------------------------|-----------------------|------------------------------|
| sqlite     | \>=3\.8\.3  | since Jessie/8                                            | yes                   | yes                         |
| postgresql | \>=9\.4     | since Jessie/8                                            | yes                   | \>=9\.5                      |
| mysql(old) | \>=5\.5     | 5\.5 in Jessie/8, dropped in Stretch/9 in favor of mariadb| no                    | \>=5\.6                      |
| mysql(new) | \>= 8\.0\.1 | No                                                        | yes                   | unclear                     |
| mariadb(old)| \>=10\.0    | Yes                                                      | no                    | \>=1\.1                      |
| mariadb(new)   | \>=10\.2\.2 | since Buster/10                                       | yes                   | yes                         |

EDIT: Mysql 5,7 and 8+ are available in official Ubuntu repositories.

| Database     |  Version |  Specificity                                                       |
[...]

PR #2803, opened 2020-03-31
Fix/697  refactoring to reduce query number

inkhey:
related to #697
Proposition of database optimization:
- Remove duplication of call to user (current user object is stored one time)
- Add current revision_id to content table to simplify query and join
- Add join in code to retrieve related object in one query instead of multiple.
- Optimize children and recursive children request (using recursive CTE for the last one if possible)
- Do not consider anymore than reading a comment mean reading his parent (not used in tracim ui and tricky mecanism, this simplify the mecanism and i do no see any reason to support this).
- Specific code to handle and to test in travis all case possible with mysql/mariadb.

Need discussion/To be decided:

- last revision deletion is tricky, current behavior is allowing empty revision_id in content table but that can lead to problem, we do need to insert value here. Easiest way may be to make batch change at the end of the process.

- Recursive CTE is not supported in old version of mysql, both postgresql and sqlite added support to this feature  a long time ago. Support in Mysql-like database need mysql > 8.0.1 or Mariadb > 10.2.2 . With mysql 8.0, these database differ and have different behavior. Supporting both old and new database version is done in this PR but it's complicated and multiplicate the number of needed test. We do need to decide what we should support.

- Need confirmation about reading a comment does not read the parent.

- We should think about a mecanism to join properly table when needed in view, currently, we do not know when we do need some related data. This should work with applications like public_upload, etc…

## Checkpoints

**For code reviewer**

- [ ] Code is clear enough

**For developper**

- [ ] Manual tests done to ensure stability on whole application layers, issue expectation follow
- [ ] Original issue have been updated with a short resume of implemented solution

**For quality team**

- [ ] Tested by quality team


inkhey:
Simple table of support of  recursive CTE + Debian and django 3.0 (decembre 2019) support for these database, in order to make decision on which database to retain:

| Database   | Version    | Official Debian support                                    | Support recursive CTE | official support Django 3\.0 |
|------------|------------|------------------------------------------------------------|-----------------------|------------------------------|
| sqlite     | \>=3\.8\.3  | since Jessie/8                                            | yes                   | yes                         |
| postgresql | \>=9\.4     | since Jessie/8                                            | yes                   | \>=9\.5                      |
| mysql(old) | \>=5\.5     | 5\.5 in Jessie/8, dropped in Stretch/9 in favor of mariadb| no                    | \>=5\.6                      |
| mysql(new) | \>= 8\.0\.1 | No                                                        | yes                   | unclear                     |
| mariadb(old)| \>=10\.0    | Yes                                                      | no                    | \>=1\.1                      |
| mariadb(new)   | \>=10\.2\.2 | since Buster/10                                       | yes                   | yes                         |

inkhey:
| Database     |  Version |  Specificity                                                       |
|--------------|----------|--------------------------------------------------------------------|
| sqlite       | >=3.8.3  | Limited support for alter table(1)                                 |
| postgresql   | >=9.4    | True Enum type, True sequence type (2)                             |
| mysql(old)   | >=5.5    |  No CTE, No Constraint, Type in alter column, Create automatic index on Foreign Keys (3) |
| mysql(new)   | >=8.0.1  | Constraint migration Issue(4) , Type in alter column, Create automatic index on Foreign Keys (3)                         |
| mariadb(old) | >=10.0   | No CTE, No Constraint, Type in alter column, Create automatic index on Foreign Keys (3)  |
| mariadb(new) | >=10.2.2 | Constraint migration Issue(4), Type in alter column, Create automatic index on Foreign Keys (3)                         |

1. need to use batch operations  : https://github.com/tracim/tracim/blob/41e7671a503ed9fec566f105d0dba85dd0a975ab/backend/doc/migration.md#sqlite-support-of-alter-table-is-limited-use-always-batch_op-

2. need **specific migration code** to handle migration modification of enum:  https://github.com/tracim/tracim/blob/41e7671a503ed9fec566f105d0dba85dd0a975ab/backend/doc/migration.md#postgresql-support-real-enum-type

3. Cannot easily add/remove index in migration, **need specific migration code** (see https://github.com/tracim/tracim/blob/41e7671a503ed9fec566f105d0dba85dd0a975ab/backend/doc/migration.md#mysql-does-already-add-index-to-foreign-keys) + Type needed in alter column (https://github.com/tracim/tracim/blob/41e7671a503ed9fec566f105d0dba85dd0a975ab/backend/doc/migration.md#mysql-require-always-type-to-be-defined-when-altering-column)
4. migration of type with constraint cause some trouble in recent mariadb/mysql version ( see https://github.com/tracim/tracim/blob/41e7671a503ed9fec566f105d0dba85dd0a975ab/backend/doc/migration.md#mysql-require-always-type-to-be-defined-when-altering-column)

inkhey:
about:
`last revision deletion is tricky, current behavior is allowing empty revision_id in content table but that can lead to problem, we do need to insert value here. Easiest way may be to make batch change at the end of the process.`

We maybe should decide to not allow last revision deletion, need to be verify but this may be possible with foreign key. This way when we delete a revision specificly , we delete only old revision,
current revision can be deleted only when deleting current content.

lebouquetin:
@inkhey all information must be put in the issue, not in the MR. MR is the result, not the place to discuss about how/what to implement

lebouquetin:
TODO :

- [ ] move all information from merge request to issue.
- [ ] decision is to support sqlite, postgresql, mariadb (new) and mysql (new) (all having recursive CTE)

inkhey:
> TODO :
> 
>     * [ ]  move all information from merge request to issue.
> 
>     * [ ]  update the issue (wrong labels, no project, no associated pull request)
> 
>     * [ ]  decision is to support sqlite, postgresql, mariadb (new) and mysql (new) (all having recursive CTE)

There does exist other blocking point to clarify :
- Last revision deletion support : how ? 
- Need confirmation about reading a comment does not read the parent.

Others point we can discuss are non-blocking, so we can postpone them in specific issue.

inkhey:
@PhilippeAccorsi as this PR remove support for old mysql/mariadb version, we do need to update properly documentation and add a breaking change somewhere in changelog.
Can you review the documentation part ?

lebouquetin:
With last fixes the MR is ok for me. @PhilippeAccorsi did you look at the documentation ? is it ok for you ?

PhilippeAccorsi:
Under reading

inkhey:
About testing, what can be done:
1. local test to verify add/update content work correctly but also content list and  dashboard view.
2. Testing clean database command (remove user/…)
3. Testing performance improvment for dashboard on a big database.

No sure, we should test all this for merge.

backend/README.md
```diff
@@ -69,18 +69,18 @@ Install the project in editable mode with its develop requirements:
 
     pip install -e ".[dev]"
 
-If you want to use PostgreSQL, MySQL or other database engines
-than the default one (SQLite), you need to install python driver for those databases
+If you want to use PostgreSQL, MySQL(8.0.1+) or MariaDB(10.3+) database engine instead of
```
PhilippeAccorsi, line 72: I think its good here to add also Postgresql version.

backend/README.md
```diff
@@ -69,18 +69,18 @@ Install the project in editable mode with its develop requirements:
...
+the default one (SQLite), you need to install python driver for those databases
```
PhilippeAccorsi, line 73: Same about sqlite I think.

backend/README.md
inkhey: For both postgresql and sqlite, version that should work are very old, so it's difficult to be sure how far we are compatible in the past.
I suggest to say we do "officially" support official debian stretch version at minimum requirement.

PhilippeAccorsi:
Need small test here. I tried to do that today.

backend/README.md
inkhey: | Database   | Tested by Travis    | Stretch version     | Buster Version       |
|------------|---------------------|---------------------|----------------------|
| Postgresql | 9.3                 | 9.6                 | 11                   |
|  Sqlite    | bundled with python | bundled with python |  bundled with python |

inkhey: Sqlite is a tricky case one as it is bundle with python and i didn't find documentation about python version meaning sqlite version.

lebouquetin: @inkhey you can write postgresql 9.3 or more.
For sqlite version, is it possible to use another version of sqlite (not the one included in python) ?

inkhey: minimal version that should (untested) work (recursive cte) are 3.8.3 for sqlite and postgresql 8.4+.

lebouquetin: You can get the sqlite version from the python interpreter: http://zetcode.com/db/sqlitepythontutorial/

You know what is the minimal python version supported, so you can get the minimal sqlite version

inkhey: changed.

backend/README.md
inkhey: changed.

PhilippeAccorsi:
> About testing, what can be done:
> 
>     1. local test to verify add/update content work correctly but also content list and  dashboard view.
> 
Done
>     2. Testing clean database command (remove user/…)
> 
Done
>     3. Testing performance improvment for dashboard on a big database.
> 
> 
> No sure, we should test all this for merge.

PhilippeAccorsi:

Issue #2812, opened 2020-04-02
Uncorrect log in Tracim (especially mailFetcher) return empty value instead of returning Traceback


inkhey:
## Bug description & symptom

Some logging return error but with empty explanation, especially in mailFetcher.
This issue as not been reported yet because these error case are something very hard to reproduce (should not happened in normal usage for exemple), they do exist to ensure daemon process like mailFetcher will not stop.

## Diagnostic

logging like : 

```python
try:
    dosomething()
except SomeException as exc:
    log = "Some log message {}"
    logger.error(self, log.format(exc.__str__()))
```

are broken, because exception may have empty str value.

## Expected

replace these logging to something like:

```python
try:
    dosomething()
except SomeException as exc:
    log = "Some log message"
    logger.exception(self, log))
```

in order to have real log

PhilippeAccorsi:
Tested and verified on branch. Wait merge to close.

PR #2825, opened 2020-04-06
[For 2.6.3] avoid logging exception string value, log traceback instead

inkhey:
Related to #2812 

Done:
- refactor Logger
- remove all exception string value logging and replace them to exc_info=true option or logger.exception() method

:warning: **Reviewers**: please take time to reviews this code, this code change code for many error case that are not well tested because they does not occured in normal working Tracim, some are needed to ensure daemon does not failed at all for example. So, please have this in mind when reviewing this code, in order to avoid "broken code" that we don't see at all.

## Checkpoints

**For code reviewer**

- [x] Code is clear enough

**For developper**

**For quality team**

- [x] Tested by quality team

### DIFF
--- a/backend/tracim_backend/lib/core/content.py
+++ b/backend/tracim_backend/lib/core/content.py
@@ -2,7 +2,6 @@
 from contextlib import contextmanager
 import datetime
 import os
-import traceback
 import typing
 
 from depot.io.interfaces import StoredFile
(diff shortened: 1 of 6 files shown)


lebouquetin:
> I would also import Any, List and Optional from the typing module instead of using them as typing.Any, typing.Optional and typing.List.

This kind of stuff is defined in coding rule. There are no conditionnal suggestions : the code must be ok with coding rules (and if coding rules should change, this is another subject)

PhilippeAccorsi:
Tested and verified on branch. tested with empty depot, wrong smtp port or wrong ssl connection.

Issue #2872, opened 2020-04-09
⬆️ Dep: Upgrade Cypress (to 12)


raphj:
We are currently using Cypress 3. This is an old version, and we should migrate to Cypress 8. An interesting feature of Cypress 4 is beta support for Firefox and automated testing of Tracim on Firefox might be great. A feature of Cypress 6 is cy.intercept(), which allows intercepting requests done with fetch. 

This should be quite easy but still needs a bit of work because Cypress 4 breaks some of our tests.

See the migration guide here: https://docs.cypress.io/guides/references/migration-guide.html

We are affected by this change: [cy.contains() ignores invisible whitespaces](https://docs.cypress.io/guides/references/migration-guide.html#cy-contains-ignores-invisible-whitespaces):

> Browsers ignore leading, trailing, duplicate whitespaces. And Cypress now does that, too.
>    ```html
>    <p>hello
>    world</p>
>    ```
>
>    ```javascript
>    cy.get('p').contains('hello world') // Fail in 3.x. Pass in 4.0.0.
>    cy.get('p').contains('hello\nworld') // Pass in 3.x. Fail in 4.0.0.
>    ```

Well, except browsers don't ignore whitespaces in textareas, and we have two tests checking that new lines are correctly being inserted in textareas:

- `cypress/integration/share/new_share_download_spec.js`
- `cypress/integration/share/new_share_upload_spec.js`

For better or worse, this feature is going away and we cannot stay on Cypress 3 eternally, so we need to discuss how we should handle this change (find a way to rewrite the tests, or sadly get rid of them).

We are affected by another change: we are using `cypress-file-upload` 3.5.3. This version is not compatible with Cypress 4. We would need migrate to version 5.0.3.
However, this version replaces its `upload` method that we use by an `attachFile` method. We use `upload`in several tests through the command `dropFixtureInDropZone`, defined in the file `cypress/support/commands.js`. So probably not a big deal.

For the record, here are the affected tests:

- cypress/integration/app_file/add_file_as_user_spec.js
- cypress/integration/helpers/file.js
- cypress/integration/share/guest_upload_form__upload_file_with_password_spec.js

There is another thing to consider: Cypress does not seem to find its modules when Yarn's PNP feature is enabled, and we may want to use it. Getting rid of `cypress-file-upload` would probably be good if it is easy to do, as well as the other module that we use, `cypress-wait-until`.

[...]

PR #2751, opened 2020-03-10
[For 2.6.x] Faster retrieval of frontend dependencies, without sudo

raphj:
This pull request fixes #2747.

- It centralises `node_modules` folders scattered across the parts for the
  frontend (frontend, frontend_lib, apps) in one place, and use symbolic links
  for the node_modules of each project to the shared node_modules folder.
  This allows speeding up the retrieval of node dependencies a lot.
    
  A script builds a super `package.json` from all the packages.json files of
  the frontend folders, ensuring that the versions of the dependencies are all
  compatible with each other. This file is stored in an hidden folder:
  `.frontend-apps-node-modules`
    
- It gets rid of `npm link` in favor a raw symbolic link, removing the need to be
   root to build Tracim (except for installing the system dependencies).

One caveat is that running `npm instal`l in one of the frontend-related folders, the `tracim_frontend_lib` symbolic link is removed from the `node_modules` folder. This is easily fixed by re-creating the symbolic link.

Related issue? #2747 

## Checkpoints

- [x] Code is clear enough
- [x] If there are FIXME in the code, then there are associated issues and issue is referenced in the FIXME
- [x] If there are TODO, NOTE or HACK in code, date and developer initials are present
- [x] Automated tests have been written (covering feature or non regression), or *at least* an issue has been created for test implementation

- [ ] Tested by quality team

Issue #2747, opened 2020-03-09
Installing frontend dependencies is slow and requires being root


raphj:
See #2916 for a comprehensive explanation about the migration to Yarn, that fixes this issue.

## Bug description & symptom

The script to install frontend dependencies runs npm install several times, leading  to duplicate downloads of dependencies.

The npm link also requires being root to build Tracim.

## Diagnostic

Run: ./install_frontend_dependencies.sh

## Expected

Each dependency is not downloaded more than once.
Root is not required to install the dependencies is all the system dependencies (Debian packages) are already there.

## Context and version information

tracim develop branch (984d624f53ea25ffa5aa16ead91b297d39165268) and up-to-date Debian Buster

## Prerequisites

This issue does not depend on any other issue 

## Impacted components

The build system

## Extra information

The following pull request fixes this issue: #2873

PhilippeAccorsi:
Can you use bug or feature template of issue for this request? 
Here we not seen what is expected, and also what is not good in script (peace of code is better to understand what is not optimized).

Some of description visible in merge request can be added here, I think.

PhilippeAccorsi:
There are some impacts in Tracim docker image?

raphj:
I edited the bug report as requested. Thanks. I am going to check the Docker image in the pull request.

Issue #1450, opened 2019-02-14
Merge dependencies list of tracim apps


Skylsmoi:
Currently, each apps has its own node_modules folder.

Almost every dependencies of apps are the same, we should mutualize them.

Yarn has the "workspace" feature that does it. See https://yarnpkg.com/blog/2017/08/02/introducing-workspaces/

See https://github.com/tracim/tracim/issues/2916 for a comprehensive explanation about the migration to Yarn.

Skylsmoi:
@raphj 
I think you will have to check every PR that have started before #2873 is merged to check if something will need to be adapted to the new architecture

PR #2873, opened 2020-04-09
[For 3.0.0] Replace NPM with Yarn

raphj:
Related to #1450, #2747 and  #2916.
___
Todos:
- [x] remove package-lock.json and npm-debug.log from .gitignore
___

Behind this short title hides a huge pull request that replaces my first attempt to fix this issue, https://github.com/tracim/tracim/pull/2751, which was less intrusive, but also not great.

This pull request:
- fixes a lot of lint issues spotted by the new version of eslint
- updates and unifies many dependencies across the different sub projects
- makes the NPM scripts use the webpack-cli binary instead of the webpack binary, which does seem not work with Yarn2. webpack calls webpack-cli anyway.
- prepares the resolution of packages under the new Plug 'n' Play (PNP) mecanism of Yarn. See https://yarnpkg.com/features/pnp/
- and eventually switches from NPM to Yarn.

The immediate advantages are:

- Simpler dependencies management.
    - `packages.json` in each project still works in the same way and should remain compatible with NPM, except for cypress tests if yarn is not installed.
   - Installing dependencies is done simply by running  `yarn install` from any folder inside the repository, instead of requiring to run `npm install` in each subproject.
   - Upgrading a dependency across the whole repository is done by running `yarn up <package name>`.
- We get rid of npm link: Yarn workspaces naturally manages what we wanted to achieve 
- Therefore, we don't need to be root anymore to install the frontend dependencies, provided that system dependencies are there.
- On a fast connection, installing the dependencies drops from around 8 minutes to 1.5 minutes. A lot less bandwidth is used, which is especially great on slow connections. It will probably make CI runs a lot faster too.
- The size of the working folder of Tracim drops from more than 2G to less than 800M.

We make use of Yarn workspaces to merge dependencies across the frontend sub-projects. This can be seen in the root `package.json` file introduced by this pull request.

Yarn introduced a new way of managing the storage of node modules in 2018 : Plug 'n' Play (PNP). Yarn 2, released in January, switched to PNP by default. This feature completely gets rid of node_modules folders and stores the modules in a `.yarn` folder at the root of the workspace, in zip files (which further saves space). Unfortunately, I had to disable this for now and use regular `node_modules` folder, because Cypress cannot seem to find its dependencies when using PNP. PNP is the future envisioned by the Yarn project, so we need to discuss whether we should get rid of these dependencies. See  issue #2872 for more information.

What needs to be done is decide how to handle the migration of existing working directories of Tracim from NPM to Yarn.

- NPM should still work in existing working directory. Developers or users do not need to switch to Yarn (but they should).
- running `./install_frontend_dependencies.sh` will install and run Yarn 2. This will work on existing working copies, but it won't clean up the duplicate node modules downloaded by NPM.

We may want to clean up the working copy in `./install_frontend_dependencies.sh`. We may detect that a developer repository has not yet migrated to Yarn 2 by looking for a known difference in one of the `node_modules` folders: things are different between Yarn 2 and NPM and the differences are easily spotted. If we succeed in enabling PNP before this pull request is merged, we will even be able to simply look for the existence of `node_module` folders: they don't exist when PNP is used, everything is stored in the folder `.cache`. However, cleaning up requires introducing a call to the `rm`, which should not be handled lightly. As an alternative, we could suggest developers and users to cleanup their directory themselves, and give a command to run manually at their own risk. This command is:

    rm **/package-lock.json node_modules **/node_modules

We should also make sure to unlink `tracim_frontend_lib` from the system or the home directory. i.e. `tracim_frontend_lib` is not in `/usr/local/lib/node_modules/` or in whatever folder NPMcould have used to link `tracim_frontend_lib`.

Reviewers should take great care. This pull request touches a lot of things. I suggest them to review it commit by commit, instead of reviewing the whole blob in one shot, which would probably be an horrible thing to do. I made each step in its own commit. I can assure you I became fluent in `git rebase` :-P. On the bright side, I made sure that all the tests pass after the first commit (linting) and after the last, in which we eventually switched to Yarn, I am pretty confident that this should not break a lot of things.

A lot of words in this description, but this is only because I am bored of waiting for my last run of the Cypress tests to end. And… *All specs passed!*. Done! Hurray!

Let's invoke @lebouquetin, @Skylsmoi and @metzgegu and @buxx at least.

P.S. I know this was planned for Tracim 3, but I needed to grasp the wonderful world of frontend management with NPM, Webpack and their friends to deeply understand how things work, and working on this was the perfect way to start digging into this… rabbit hole.

## Checkpoints

**For code reviewer**

- [x] Code is clear enough

raphj: No new feature code wise, so I checked the tests.

**For developer**

**For quality team**

- [ ] Tested by quality team


frontend/src/container/Account.jsx
```diff
@@ -105,10 +105,11 @@ export class Account extends React.Component {
     const { props } = this
     const fetchUserAgenda = await props.dispatch(getLoggedUserCalendar())
     switch (fetchUserAgenda.status) {
-      case 200:
+      case 200: {
```
Skylsmoi, line 108: is this a new linting rule ?

frontend/src/container/AppFullscreenRouter.jsx
```diff
@@ -36,79 +36,97 @@ export class AppFullscreenRouter extends React.Component {
...
-          const agendaConfig = {
-            workspaceId: null,
-            forceShowSidebar: true
-          }
-          props.renderAppFullscreen({ slug: 'agenda', hexcolor: '#7d4e24', appConfig: agendaConfig }, props.user, null, {})
-          return null
-        }} />
+        <Route
+          exact
+          path={PAGE.ADMIN.WORKSPACE} render={() => {
```
Skylsmoi, line 41: this doesn't seem consistent to me.
Did a new eslint rule stage that the old syntax is wrong ?
About the new version, to be consistent, render props should be at a new line

frontend/src/container/SearchResult.jsx
```diff
@@ -230,21 +229,18 @@ class SearchResult extends React.Component {
...
               breadcrumbsList={props.breadcrumbs}
             />
 
             <PageContent parentClass='searchResult'>
-              <div className='folder__content' data-cy={'search__content'}>
-                {currentNumberSearchResults > 0 &&
-                  <ContentItemHeader showSearchDetails />
-                }
+              <div className='folder__content' data-cy='search__content'>
+                {currentNumberSearchResults > 0 && <ContentItemHeader showSearchDetails />}
```
Skylsmoi, line 243: is
```suggestion
{currentNumberSearchResults > 0 && (
  <ContentItemHeader showSearchDetails />
)}
```
fine for the new eslint rules ?

frontend_app_agenda/src/container/Agenda.jsx
```diff
@@ -117,7 +117,7 @@ class Agenda extends React.Component {
   }
 
   loadAgendaList = async workspaceId => {
-    const { state } = this
+    const { state, props } = this
```
Skylsmoi, line 120: you added props but I don't see any usage, shouldn't this create a linting error ?

frontend_app_file/src/container/File.jsx
```diff
@@ -745,7 +747,7 @@ class File extends React.Component {
           customClass={`${state.config.slug}`}
           i18n={i18n}
         >
-          <div /* this div in display flex, justify-content space-between */>
+          <div> {/* this div in display flex, justify-content space-between */}
```
Skylsmoi, line 750: was this firing a linting error ?

frontend/src/component/Dashboard/AgendaInfo.jsx
```diff
@@ -18,6 +18,7 @@ export const AgendaInfo = props => (
             className='agendaInfo__content__text__help primaryColorFont primaryColorFontDarkenHover'
             href='https://github.com/tracim/tracim/issues/1673'
             target='_blank'
+            rel='noopener noreferrer'
```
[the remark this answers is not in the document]
raphj: Standard rule jsx-no-target-blank: https://github.com/standard/eslint-config-standard-jsx/blob/master/eslintrc.json#L35

See https://github.com/yannickcr/eslint-plugin-react/blob/HEAD/docs/rules/jsx-no-target-blank.md

frontend/src/container/Account.jsx
raphj: Yes : https://eslint.org/docs/rules/no-case-declarations

frontend/src/container/AppFullscreenRouter.jsx
raphj: Yes: https://github.com/standard/eslint-config-standard-jsx/blob/master/eslintrc.json#L27
https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-first-prop-new-line.md

But you are right, render should be at a new line and I will thix this

frontend_app_file/src/container/File.jsx
raphj: Yes: react/jsx-no-comment-textnodes

https://github.com/standard/eslint-config-standard-jsx/blob/0fc0f43a8feaec180ecd0f258b03c081df9b47ef/eslintrc.json#L33

raphj: Yes: react/jsx-tag-spacing

https://github.com/standard/eslint-config-standard-jsx/blob/0fc0f43a8feaec180ecd0f258b03c081df9b47ef/eslintrc.json#L39

frontend_app_agenda/src/container/Agenda.jsx
raphj: `props` is used on lines 132 and 135, this caused a linting error that was not raised because standard was not enabled for app agenda before.

frontend/webpack.config.js
```diff
@@ -45,7 +45,7 @@ module.exports = {
     rules: [{
       test: /\.jsx?$/,
       enforce: 'pre',
-      exclude: [/node_modules/],
+      exclude: [/node_modules/, /frontend_lib/],
```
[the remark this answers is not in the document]
raphj: It is not the case when yarn is used in PNP mode: frontend_lib is not in a node_modules folder because there aren't node_modules folders anymore in this mode. This is why I put the additional regular expression. The regular expression `/node_modules/` is going to be useless. I suggest keeping both regular expressions for now so Tracim is compatible with both PNP mode and regular `node_modules` folders and we just have to figure out Cypress to enable PNP.

frontend/package.json
```diff
@@ -20,64 +20,69 @@
...
+    "eslint-import-resolver-node": "^0.3.3",
+    "history": "^4.10.1",
     "i18next": "10.5.0",
     "js-cookie": "^2.2.0",
     "lodash": "^4.17.11",
     "prop-types": "^15.7.2",
     "query-string": "^6.8.1",
     "radium": "^0.25.2",
-    "react": "16.7.0",
+    "react": "^16.13.1",
```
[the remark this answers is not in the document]
raphj: Okay, let's talk about it and we can revert to the old react if necessary and decide to upgrade later :-)

frontend/webpack.config.js
```diff
@@ -79,8 +81,20 @@ module.exports = {
     }]
   },
   resolve: {
+    plugins: [
+      PnpWebpackPlugin,
+    ],
+    alias: {
+      // Make ~tracim_frontend_lib work in stylus files
+      '~tracim_frontend_lib': path.dirname(path.dirname(require.resolve('tracim_frontend_lib')))
```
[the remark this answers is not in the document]
raphj: I reused the notation used in the doc of stylus-native-loader: https://www.npmjs.com/package/stylus-native-loader

Look for the following in this page:

>  ```javascript
>  resolve: {
>    // All aliases are used for Stylus @import and @require path resolution
>    alias: {
>      // Maps @import('~styl/*') to @import('/path/to/src/styl/*')
>      '~styl': path.join(__dirname, 'src/styl'),
>  
>      // Maps @import('mixins') to @import('/path/to/src/styl/mixins')
>      'mixins$': path.join(__dirname, 'src/styl/mixins'),
>    },
>  },
>  ```

See also the alias declaration section that defines it properly in tracim's own webpack configs. 

This seems a standard notation to denote an import from a node module in stylus, this will help a new dev looking at the stylus file/

We have no choice: the old way hard codes the location of tracim_frontend_lib, but with yarn the location is not in node_modules anymore. The old way does not work anymore and maybe worked by luck (it was dependent on NPM's specific behavior). The new way works with regular NPM though :-)

raphj:
So, we need to decide on which way we should format the following piece of code, of which several instances are present in the Tracim frontend code, now rejected by standard:

```js
<div>
  {condition
    ? <Elem />
    : <OtherElem />
  }
</div>
```

The problem is: the closing brace is on its own line whereas the opening brace isn't, which is not consistent. In this PR, I settled on:

```js
<div>
  {condition
    ? <Elem />
    : <OtherElem />}
</div>
```

I don't like it so much but I'm okay with it.

Standard is not fine with the following:

```js
<div>{
   condition
     ? <Elem />
     : <OtherElem />
}</div>
```

Standard does not like the following neither:

```js
<div>
{
  condition
    ? <Elem />
    : <OtherElem />
}
</div>
```
Standard is, however, fine with this one:

```js
<div>
  {
    condition
      ? <Elem />
      : <OtherElem />
  }
</div>
```

If you are not fine with the first but fine with the last, I will make the necessary edits (@Skylsmoi :-))

frontend/src/container/AppFullscreenRouter.jsx
raphj: And I just did.

frontend/webpack.config.js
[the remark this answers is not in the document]
Skylsmoi: Ok for the 2 path.dirname(). Since it's a bit confusing, I'd advise a comment to explain it

About the "~", if it's in the module documentation, I'm fine with it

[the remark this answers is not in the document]
Skylsmoi: you can resolve that conversation

frontend/webpack.config.js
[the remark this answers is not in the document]
Skylsmoi: Ok, as long as frontend_lib is not listed as local dependency in package.json, it should be fine.

I just don't understand how yarn handles it, we'll can talk about it later.

Afaik, we currently don't use pnp mode because of cypress. So we need to keep /node_modules/. When it'll not be required, we will have to remove it though

frontend/package.json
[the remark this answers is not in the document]
Skylsmoi: I would say to keep it since, from my first tests, I couldn't reproduce the error.

If the error reappear we most likely keep the old react version.

If there isn't the error, that'll be great, we'll can keep being up to date with react version

frontend/src/component/Dashboard/AgendaInfo.jsx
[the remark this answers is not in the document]
Skylsmoi: :+1:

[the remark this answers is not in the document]
Skylsmoi: you can resolve the issue

frontend/src/container/Account.jsx
Skylsmoi: ok, good. You can resolve this conversation

frontend_app_agenda/src/container/Agenda.jsx
Skylsmoi: ok, you can resolve

frontend_app_file/src/container/File.jsx
Skylsmoi: ok you can resolve

Skylsmoi:
about the new standard rule for ternary operations in jsx:
- I don't like the solution in this PR because I think that not having the opening et the closing bracket on the same column in confusing
- the only other solution standard is fine with seems good to me. It'll add lines of code but it is consistent

So, I'd say we settle for your last example.

This creates another question:
Should we do the same for codes like so:
``` jsx
<div>
  { shouldDisplayButton && (
    <Button />
  )}
</div>
```
From my understanding, by using parenthesis, standard is fine with this syntax. But it isn't possible to use it for ternary operations.
Should the above also be refactored to:
``` jsx
<div>
  {
    shouldDisplayButton && (
      <Button />
    )
  }
</div>
```
Even though it isn't necessary ?

___
About you not agreeing with standard rules, we'll can discuss it. I personally agree with possibly all of them.

Skylsmoi:
I think this PR shouldn't be for 2.6.x because there are too much changes.
I'd say we should put it for 3.0 and merge it asap since it'll improve development efficiency, mostly by speeding up ci
This should be raph's priority as soon as he is available for Tracim

raphj:
> About you not agreeing with standard rules, we'll can discuss it. I personally agree with possibly all of them.

I'm fine with standard or whatever rule set has been decided on the project, no worry :-)

raphj:
About the packages lock files, I did not find them in the repository. What should we do?

Skylsmoi:
it is in gitignore indeed.
remove it from gitingore then, and also for npm-debug.log

raphj:
About your last question on the standard rule, the standard rul is not specific to ternary condition, we will need to reformat those too ;-)

Skylsmoi:
You mean this:
``` jsx
<div>
  { shouldDisplayButton && (
    <Button />
  )}
</div>
```

will fire an error with standard ?
I'm confused I though it wouldn't

raphj:
> You mean this:
> 
> ```js-jsx
> <div>
>   { shouldDisplayButton && (
>     <Button />
>   )}
> </div>
> ```
> 
> will fire an error with standard ?
> I'm confused I think it didn't when I tried to build on your branch

You are right, it will work :-)

raphj:
So, I applied this across the code base and it also works for ternaries.

frontend/package.json
```diff
@@ -50,6 +50,10 @@
     "webpack": "^4.42.1",
     "webpack-cli": "^3.3.11"
   },
+  "peerDependencies": {
```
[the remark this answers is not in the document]
raphj: So I haven't taken the time to take a deep look at this yet, but I just tried to remove one of these peer dependencies block and get these warnings when running Yarn as a result:

```
➤ YN0000: ┌ Resolution step
➤ YN0002: │ tracim_app_admin_workspace_user@workspace:frontend_app_admin_workspace_user doesn't provide cheerio@0.19.x || 0.20.x || 0.22.x || ^1.0.0-0 requested by chai-enzyme@npm:1.0.0-beta.1
➤ YN0002: │ tracim_app_admin_workspace_user@workspace:frontend_app_admin_workspace_user doesn't provide cheerio@^1.0.0 requested by tracim_frontend_lib@workspace:frontend_lib
```

I wanted to check that this was still necessary in Yarn 2 since I made those changes with Yarn 1 (because of mistake).


Putting them in `dependencies` or `devDependencies` yields:

```
➤ YN0000: ┌ Resolution step
➤ YN0001: │ Error: No candidate found for cheerio@npm:^1.0.0
    at C (/home/raph/dl/tracim/.yarn/releases/yarn-berry.js:42:105034)
    at process._tickCallback (internal/process/next_tick.js:68:7)
➤ YN0000: └ Completed in 0.35s
➤ YN0000: Failed with errors in 0.35s
```

We probably shouldn't put them in `dependencies` or `devDependencies` anyway even if we could, to make it clear that our code does not directly depend on them.

We should both take a look at how peer dependencies work. I'm not sure if we should block this pull request in the meantime tough. What do you think?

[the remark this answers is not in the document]
raphj: So I haven't taken the time to take a deep look at this yet, but I just tried to remove one of these peer dependencies block and get these warnings when running Yarn as a result:

```
➤ YN0000: ┌ Resolution step
➤ YN0002: │ tracim_app_admin_workspace_user@workspace:frontend_app_admin_workspace_user doesn't provide cheerio@0.19.x || 0.20.x || 0.22.x || ^1.0.0-0 requested by chai-enzyme@npm:1.0.0-beta.1
➤ YN0002: │ tracim_app_admin_workspace_user@workspace:frontend_app_admin_workspace_user doesn't provide cheerio@^1.0.0 requested by tracim_frontend_lib@workspace:frontend_lib
```

I wanted to check that this was still necessary in Yarn 2 since I made those changes with Yarn 1 (because of mistake).


Putting them in `dependencies` or `devDependencies` yields:

```
➤ YN0000: ┌ Resolution step
➤ YN0001: │ Error: No candidate found for cheerio@npm:^1.0.0
    at C (/home/raph/dl/tracim/.yarn/releases/yarn-berry.js:42:105034)
    at process._tickCallback (internal/process/next_tick.js:68:7)
➤ YN0000: └ Completed in 0.35s
➤ YN0000: Failed with errors in 0.35s
```

We probably shouldn't put them in `dependencies` or `devDependencies` anyway even if we could, to make it clear that our code does not directly depend on them.

We should both take a look at how peer dependencies work. I'm not sure if we should block this pull request in the meantime tough. What do you think?

frontend/src/container/WorkspaceContent.jsx
```diff
@@ -749,9 +756,10 @@ class WorkspaceContent extends React.Component {
                     rootContentList={rootContentList}
                     isLast={isWorkspaceEmpty || isFilteredWorkspaceEmpty}
                     t={t}
-                  />}
+                  />
+                )}
 
-                {state.contentLoaded && (isWorkspaceEmpty || isFilteredWorkspaceEmpty)
+                {((state.contentLoaded && (isWorkspaceEmpty || isFilteredWorkspaceEmpty))
```
Skylsmoi, line 784: I think there is one not required opening and closing parenthesis here

raphj: I think it is clearer with the additional parentheses to clearly delimitate the condition of the ternary, but I can remove them if you think it is better without. What do you think?

Skylsmoi: I'm fine with it, I thought it was an error

PhilippeAccorsi:
Not merge before release 2.7.0

Issue #2917, opened 2020-04-24
Feat: Update redux for username feature


gbonaspetti:
## Feature description

Add username in the `user` object of the redux state.

## Context

Linked to #2829 

## Impacted components
- frontend

gbonaspetti:
For this issue a `username` has been added on `user` at Redux . To make the difference clear, the functions linked to `public_name` that used the USER_NAME notation have been updated to USER_PUBLIC_NAME and new `username` fuctions have been added as USER_USERNAME.

metzgegu:
Verified on branch (wait merge to close).

PR #2928, opened 2020-04-27
[For 3.0.0] Feature/2917 adding username to user at redux

gbonaspetti:
#2917 

## Checkpoints

**For code reviewer**

- [x] Code is clear enough

**For developper**

**For quality team**

- [x] Tested by quality team

### DIFF
--- a/frontend/test/src/container/Account.spec.js
+++ b/frontend/test/src/container/Account.spec.js
@@ -11,7 +11,7 @@ import {
   UPDATE,
   USER_AGENDA_URL,
   USER_EMAIL,
-  USER_NAME,
+  USER_PUBLIC_NAME,
   USER_WORKSPACE_DO_NOTIFY,
   WORKSPACE_LIST_MEMBER,
   ADD,
@@ -34,7 +34,7 @@ import {
 
 describe('<Account />', () => {
   const setWorkspaceListMemberListCallBack = sinon.spy()
-  const updateUserNameCallBack = sinon.spy()
+  const updateUserPublicNameCallBack = sinon.spy()
   const newFlashMessageWarningCallBack = sinon.spy()
   const updateUserEmailCallBack = sinon.spy()
   const updateUserWorkspaceSubscriptionNotifCallBack = sinon.spy()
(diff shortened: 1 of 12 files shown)


gbonaspetti:
Automated tests have not been written because it is not a visible feature that can be tested with our current infrastructure.

frontend/test/src/container/Account.spec.js
```diff
@@ -115,12 +115,12 @@ describe('<Account />', () => {
     })
 
     describe('handleSubmitNameOrEmail', () => {
-      it('updateUserNameCallBack should be called when the function handleSubmitNameOrEmail() is called with a new Name', (done) => {
+      it('updateUserPublicNameCallBack should be called when the function handleSubmitNameOrEmail() is called with a new Name', (done) => {
```
Skylsmoi, line 118: take advantage of this PR to rename this test. It should start with "should"

gbonaspetti: ok, I'll rename all the tests in that file to be consistent

frontend/test/src/reducer/user.spec.js
```diff
@@ -0,0 +1,55 @@
...
+  logged: null,
+  auth_type: '',
+  timezone: '',
+  profile: PROFILE.user,
+  email: '',
+  is_active: true,
+  avatar_url: null,
+  created: '',
+  public_name: '',
+  lang: getBrowserLang(),
```
Skylsmoi, line 22: I don't think you should be using getBrowerLang().
The unit tests aren't run inside a browser, I don't know what to expect from getBrowerLang() but I'm assuming it is undefined or a default value.

frontend/test/src/reducer/user.spec.js
```diff
@@ -0,0 +1,55 @@
...
+import userReducer from '../../../src/reducer/user.js'
+import {
+  UPDATE,
+  USER_USERNAME,
+  USER_PUBLIC_NAME
+} from '../../../src/action-creator.sync'
+import { getBrowserLang } from '../../../src/helper.js'
+import { PROFILE } from 'tracim_frontend_lib'
+
+const defaultUser = {
```
Skylsmoi, line 11: your test user should have actual values.
There is a user in the fixture folder, you can use it.

Check that this user has all the properties and update it if required

gbonaspetti: The purpose of defaultUser was to test that the reducer "should return the default state", it is equal to the default state, so it has no values.
For the other tests I will use the user in the fixture folder, but for this "it" test I still think it is important to keep the defaultUser.

Skylsmoi: I see. Then you need to get the defaultUser variable from the reducer's file

Skylsmoi: export it if it is required

metzgegu:

Issue #2918, opened 2020-04-24
Bug: Cannot install frontend dependencies with yarn ≠ 1.21


gbonaspetti:
## Bug description & symptom

Cannot install_frontend_dependencies with yarn version 1.19

## Screenshots

![Screenshot from 2020-04-24 16-33-00](https://user-images.githubusercontent.com/35422692/80225254-1ae36200-864b-11ea-80ad-7c9bdad88a49.png)

## Fix

The fix consists in using command `yarn policies set-version berry` instead of `yarn set version berry`, and checking that Yarn is not already version 2.

The fix supports installing and setting up the frontend dependencies using any version of Yarn but we still only use Yarn 2 after installation.

For the record and since the migration to Yarn, .yarnrc / .yarnrc.yml are not commited and are in .gitigore. If we commited these files, ./install_frontend_dependencies would fail because yarn would try to execute a non existent file in .yarn. These files therefore need to be created at install time.

raphj:
The following command only seems to work with yarn 1.22.

    yarn set version berry

The following command seems to work with older versions as well as yarn 1.22:

    yarn policies set-version berry 

I am going to open a pull request with a fix.

The fix consists in replacing the former by the latter in `install_frontend_dependencies.sh` (line 166).

Issue #2916, opened 2020-04-24
Switch from NPM to Yarn


raphj:
We are switching to Yarn.

# Context

This switch solves two issues:
 - #1450: dependencies are in multiple places across the repository
 - #2747: dependencies are downloaded multiple times, and npm link required to build Tracim, requires being root

Related pull request: #2873 

# Impacted components

- frontend
- continuous integration
- docker images

# How to migrate

### To do once

Make sure your working folder is clean to ensure that you will not lose any work.

Switch to `develop` and pull the latest changes:

```
git checkout develop
git pull
```

Then, remove NPM-related files in the root of your working directory. ⚠ This is an `rm` command, triple check what you are doing! `**` means: any subdirectory.

```
rm **/package-lock.json node_modules **/node_modules
```

Also make sure you remove the link to tracim_frontend_lib on your system in `/usr/local/lib/node_modules/` or somewhere in your `HOME` if it is not there.

### To install frontend dependencies

Run the following command, as usual:

```
./install_frontend_dependencies.sh
```

You should be all set!

### `npm` commands → `yarn` commands
Any `npm` command becomes a `yarn` command. Roughly:

Before:
```bash
npm XXXXX
npm install
````

After:
```bash
yarn XXXXX
yarn install # yarn without any argument also works  
````

### Importing  files from a module in stylus
In stylus files, Import files from `tracim_frontend_lib` using a tilde.

Before:

```stylus
@import "../../../../node_modules/tracim_frontend_lib/src/css/Variable.styl"
```

After:

```stylus
@import "~tracim_frontend_lib/src/css/Variable.styl"
```

Note: the later should work on NPM too :-)

#### [jsx-curly-newline](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-curly-newline.md):

Before:

```jsx
<div>
  {condition
    ? <Elem />
    : <OtherElem />
  }

  {condition &&
    <Elem />
  }
</div>
```

After:

```jsx
<div>
  {(condition
    ? <Elem />
    : <OtherElem />
  )}

  {condition && (
    <Elem />
  )}
</div>
```

For long conditions, prefer surrounding with parentheses (this is not enforced by standard

Correct:

```jsx
<div>
  {(condition1 && condition2 && condition3
    ? <Elem />
    : <OtherElem />
  )}
</div>
```

My advice:

```jsx
<div>
  {((condition1 && condition2 && condition3)
    ? <Elem />
    : <OtherElem />
[...]

PR #2929, opened 2020-04-27
fix(dep): support any installed version of yarn

raphj:
Related to #2918.
This allows using any installed version of yarn on the developer's system.

The fix consists in using command `yarn policies set-version berry` instead of `yarn set version berry` and checking that Yarn is not already version 2.

## Checkpoints

- [x] Code is clear enough

**For developer**

**For quality team**

- [x] Tested by quality team

### DIFF
--- a/install_frontend_dependencies.sh
+++ b/install_frontend_dependencies.sh
@@ -121,9 +121,9 @@ debian_install() {
 }
 
 setup_yarn() {
-    if [ ! -f ".yarnrc.yml" ]; then
-        log "Setting up yarn (yarn set version berry)"
-        yarn set version berry && \
+    if ! yarn -v | grep '^2\.' > /dev/null; then
+        log "Setting up yarn (yarn policies set-version berry)"
+        yarn policies set-version berry && \
             loggood "Yarn is correctly set up." || \
             logerror "Failed to set up yarn."
     fi
@@ -132,7 +132,7 @@ setup_yarn() {
         logerror "We expected Yarn 2, we got $(yarn --version)."
     fi
 
-    if ! grep -F 'nodeLinker: node-modules' .yarnrc.yml; then
+    if ! grep -F 'nodeLinker: node-modules' .yarnrc.yml > /dev/null; then
         log "Setting yarn nodeLinker to node-modules mode."
         echo 'nodeLinker: node-modules' >> .yarnrc.yml
     fi
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ previews
 
 #ignore yarn
 .yarn
+.yarnrc
 .yarnrc.yml
 node_modules
 


install_frontend_dependencies.sh
```diff
@@ -121,9 +121,9 @@ debian_install() {
 }
 
 setup_yarn() {
-    if [ ! -f ".yarnrc.yml" ]; then
-        log "Setting up yarn (yarn set version berry)"
-        yarn set version berry && \
+    if ! yarn -v | grep '^2\.' > /dev/null; then
```
Skylsmoi, line 124: looking at the related issue, you are talking about yarn version 1.22, it says here yarn > 2.0.
Is it ok ?
Update diff or issue accordingly

Skylsmoi:
> You need to add .yarnrc to .gitignore

Should we ?
yarnrc file is configuration about yarn feature.
I'm not sure about what it does exactly, but I think that ignoring it allows to have different configuration on different repository instances.
I don't it is what we expect
@raphj need your thoughts on that please, what .yarnrc does ?

install_frontend_dependencies.sh
raphj: It's okay. We need to support any version of yarn (1 or 2) when installing the dependencies, including setting up yarn 2. After that, we only want to support Yarn 2 after installation.

raphj:
.yarnrc is like .yarnrc.yml used by latest versions of yarn, on older versions.
This file is created when setting up yarn. @gbonaspetti is right, we should add this file to gitignore in my opinion.

Skylsmoi:
> .yarnrc is like .yarnrc.yml used by latest versions of yarn, on older versions.
> This file is created when setting up yarn. @gbonaspetti is right, we should add this file to gitignore in my opinion.

@raphj Confirm this please:
- .yarnrc.yml is only created when installing tracim with yarn 2
- .yarnrc is only created when installing tracim with yarn 1.22
- since tracim will then only use yarn 2, .yarnrc is useless because it will only use .yarnrc.yml

gbonaspetti:
> > .yarnrc is like .yarnrc.yml used by latest versions of yarn, on older versions.
> > This file is created when setting up yarn. @gbonaspetti is right, we should add this file to gitignore in my opinion.
> 
> @raphj Confirm this please:
> 
> * .yarnrc.yml is only created when installing tracim with yarn 2
> * .yarnrc is only created when installing tracim with yarn 1.22
> * since tracim will then only use yarn 2, .yarnrc is useless because it will only use .yarnrc.yml

I did the test if the ./install_frontend_dependencies.sh command would still work without the .yarnrc file, and I have the following error 
Screenshot from 2020-05-04 11-17-23

raphj:
@Skylsmoi 

 - .yarnrc is supposed to be created when install_frontend_dependencies.sh is ran while yarn < 1.22 (e.g. 1.19) is installed. .yarn
 - .yarnrc.yml is supposed to be created when install_frontend_dependencies.sh is ran while yarn 1.22 is installed
 - Nothing is done when yarn 2 is already installed globally, which is not recommended by the Yarn project but still possible anyway (I somehow ended in this situation on my setup)

.yarnrc and .yarnrc.yml are needed so yarn 1 finds and run yarn 2 inside the tracim folder. Without them, yarn 1 is used.

@gbonaspetti I'm going to check this

raphj:
@gbonaspetti I don't reproduce your problem on this branch. I:
 - installed a fresh Debian buster container
 - installed yarn 1.19.1,
 - cloned the repository
 - ran git checkout fix/2918__support_any_version_of_yarn
 - ran install_frontend_dependencies, which succeeds
 - removed .yarnrc
 - ran install_frontend_dependencies, which succeeds again

I expect this to fail on develop with yarn 1.19 though. What are the steps you did to produce this error?

raphj:
@Skylsmoi if we commit .yarnrc / .yarnrc.yml, ./install_frontend_dependencies will fail because yarn will try to run a non existent file in `.yarn`, these files need to be created at install time and to be git ignored in my opinion.

gbonaspetti:
> @gbonaspetti I don't reproduce your problem on this branch. I:
> 
> * installed a fresh Debian buster container
> * installed yarn 1.19.1,
> * cloned the repository
> * ran git checkout fix/2918__support_any_version_of_yarn
> * ran install_frontend_dependencies, which succeeds
> * removed .yarnrc
> * ran install_frontend_dependencies, which succeeds again
> * removed .yarnrc and .yarnrc.yml
> * ran install_frontend_dependencies, which succeeds again
> 
> I expect this to fail on develop with yarn 1.19 though. What are the steps you did to produce this error?

- git checkout develop
- git pull
- ran rm -rf **/package-lock.json node_modules **/node_modules
- change the file install_frontend_dependencies.sh, line 126
- ran ./install_frontend_dependencies.sh, it works
- remove .yarnrc
- ran  ./install_frontend_dependencies.sh, error
- restore .yarnrc
- ran ./install..., it works again

Basically I followed the switch steps (#2916), the first time it gave the issue error, so I made the change in the install file to `yarn policies set-version berry` and ran the install_frontend_dependencies command. My computer froze, I restarted and ran the install_frontend_dependencies again, it worked and at some point the .yarnrc file was created.
Having the .yarnrc file, if I do `yarn --version` I have `2.0.0-rc.32` and the install_frontend_dependencies works.
When deleting the .yarnrc file, if I do `yarn --version` I get `1.19.1` and the installation error from the screenshot.

raphj:
>     git checkout develop

That's why it fails. You are testing against develop, not this branch. You need to type this instead:

    git checkout fix/2918__support_any_version_of_yarn

let me know if it works for you :-)

gbonaspetti:
> > ```
> > git checkout develop
> > ```
> 
> That's why it fails. You are testing against develop, not this branch. You need to type this instead:
> 
> ```
> git checkout fix/2918__support_any_version_of_yarn
> ```
> 
> let me know if it works for you :-)
> 
> (and don't edit file `install_frontend_dependencies`, this branch is about making this edit unecessary)

Yes, it works, my bad, sorry.

(don't worry, I won't commit this file or anything, I just made the change to be able to do the installation and continue with my development while this pr is not merge)

raphj:
I force-pushed a squashed commit in the hope Travis would pick up the right branch. It seems to have worked. Waiting for the CI to end and I will merge.

Issue #2927, opened 2020-04-27
Feat: allow to login with username


metzgegu:
## Feature description

In login component :
- [x] Update placeholder, input icon and flash messages with the errors with username
- [x] Check if autocomplete of login still works with username

## Context

Linked to #2829

## Impacted components

- *frontend*

## Extra information

Translations :
- Invalid email or username: 
    - fr: "Adresse email ou identifiant invalide"
    - pt: "E-mail ou identificador inválido"
- Email address or username: 
    - fr: "Adresse email ou identifiant"
    - pt: "E-mail ou identificador"
- Invalid credentials: 
    - fr: "Identifiants invalide"
    - pt: "Credenciais inválidas"

gbonaspetti:
Verified on branch (wait merge to close).

metzgegu:
In order to login with `username` the login endpoint need to have a `username` param, so the frontend need to check the type of login to add the correct param.
To do that the frontend use an existing function in `frontend_lib`

PR #2934, opened 2020-04-27
[For 3.0.0] Feature/2927 allow login with username

metzgegu:
#2927 

## Checkpoints

**For code reviewer**

- [x] Code is clear enough

**For developper**

**For quality team**

- [x] Tested by quality team


frontend/src/container/Login.jsx
```diff
@@ -240,14 +240,14 @@ class Login extends React.Component {
...
               <InputGroupText
-                parentClassName='loginpage__card__form__groupemail'
+                parentClassName='loginpage__card__form__groupelogin'
                 customClass='mb-3 mt-4'
-                icon='fa-envelope-open-o'
-                type='email'
-                placeHolder={props.t('Email Address')}
-                invalidMsg={props.t('Invalid email')}
+                icon='fa-at'
+                type='text'
```
Skylsmoi, line 246: I don't think this is required. Keeping type email seems good to me.
If the user wants to use his username, since username shouldn't have any special characters, the email optimized keyboard in good.
If the user wants to use his email, email keyboard is better.

I'm talking about mobile. On desktop it shouldn't change a lot.

If the change is about email validation by the html, I think you can disabled it since we do out own email validation. Confirm we do our own, I'm not sure

Btw, please confirm taht username cannot have special chars

metzgegu: Allowed chars for username is `azAZ09-_` so errors can appears with `type=email`

Skylsmoi: why that ?
`-` and `_` are email allowed characters

metzgegu: After verification, yes you're right, in fact the browser show me an error (email invalid) because the field must contain a `@`

In FIrefox :
image

Skylsmoi: indeed, I'd say remove the html validation on this field and we'll use our custom one

Skylsmoi: but keep "email" type because it's easier for mobile users

frontend/src/container/Login.jsx
```diff
@@ -110,14 +114,19 @@ class Login extends React.Component {
...
 
-    if (email.value === '' || password.value === '') {
+    if (login.value === '' || password.value === '') {
       props.dispatch(newFlashMessage(props.t('Please enter a login and a password'), 'warning'))
       return
     }
 
-    const fetchPostUserLogin = await props.dispatch(postUserLogin(email.value, password.value, state.inputRememberMe))
+    const credentials = {
+      ...(checkEmailValidity(login.value) ? { email: login.value } : { username: login.value }),
```
raphj, line 125: Are we sure we want to do this like that?
Are people allowed to use "@" in their username? This behavior would probably be annoying for them. A discussion needs to happen with the backend crowd on this I think.
Another solution would be to have two inputs but it would probably be a bit annoying. Or a checkbox that gets checked automatically, and the user can revert the check.

metzgegu: Allowed chars in username are `azAZ09-_`, 
Yes other front solution are possible but I think it would be way better to keep just one input for both (email and username)

gbonaspetti:

PR #2989, opened 2020-05-05
fix(backend): allow none value for username at user creation

buxx:


lebouquetin:
@buxx we should not accept both `username` NULL and `email` NULL. My feeling is this is not managed...

buxx:
@lebouquetin These changes only allow to give a `null` like this:

```json
{
    "username": "bob",
    "email": null,
    # ...
}
```

because before these changes, only this was accepted:


```json
{
    "username": "bob",
    # ...
}
```

[This code](https://github.com/traacim/tracim/blob/feature/2829__unique_user_username/backend/tracim_backend/lib/core/user.py#L894) check there is one of them.

An additional check can be done at Schema level in `UserCreationSchema`.

buxx:
@lebouquetin 

> we should never accept query like this:
> [...]
> This is not in our standard structure based API 

All (almost ? @inkhey) optional api parameter can currently be omitted. I don't find this information in our api specs: https://algoo.tracim.fr/ui/workspaces/11/contents/html-document/2479 (maybe i miss it) Can you update it if it is not ? We must open a technical debt issue for update our api in this way.

> (and it may result in side effects)

Like what ?

> your link is broken

My link was about:

```python
    def create_minimal_user(
        self,
        email: typing.Optional[str] = None,
        username: typing.Optional[str] = None,
        profile: typing.Optional[Profile] = None,
        save_now=False,
    ) -> User:
        """Previous create_user method"""
        if not email and not username:
            raise EmailOrUsernameRequired("Email or username is required to create an user")
        # ...
```

> if the code is not in the schema, where is the documentation about "what combination is allowed or not?". Isn't it possible to implement the check at the schema description layer? It would be automatically documented

Yes it is possible to add the check at Schema level too. But it cannot generate documentation about that because it is an dynamic check (with python execution code). I can add information in docstring of schema, and it will be visible in generated doc (i'm doing it).

inkhey:
> @lebouquetin
> 
> > we should never accept query like this:
> > [...]
> > This is not in our standard structure based API
> 
> All (almost ? @inkhey) optional api parameter can currently be omitted. I don't find this information in our api specs: https://algoo.tracim.fr/ui/workspaces/11/contents/html-document/2479 (maybe i miss it) Can you update it if it is not ? We must open a technical debt issue for update our api in this way.

As far i know this is unspecified. I agree we should have a clear rule here. 
I just do some verification about documentation:
Not required field can be autodocumented correctly with swagger, but nullable field can't until we support openapi 3 see https://stackoverflow.com/questions/53639439/swagger-server-node-js-post-with-integer-and-or-null?noredirect=1&lq=1

> 
> > (and it may result in side effects)
> 
> Like what ?

I see a semantical difference between pushing null which may mean "set to empty" and get the default value, that why i see case where distinction may be needed. Don't remember such case in tracim.

> 
> > your link is broken
> 
> My link was about:
> 
> ```python
>     def create_minimal_user(
>         self,
>         email: typing.Optional[str] = None,
>         username: typing.Optional[str] = None,
>         profile: typing.Optional[Profile] = None,
>         save_now=False,
>     ) -> User:
>         """Previous create_user method"""
>         if not email and not username:
>             raise EmailOrUsernameRequired("Email or username is required to create an user")
>         # ...
> ```
> 
> > if the code is not in the schema, where is the documentation about "what combination is allowed or not?". Isn't it possible to implement the check at the schema description layer? It would be automatically documented
> 
> Yes it is possible to add the check at Schema level too. But it cannot generate documentation about that because it is an dynamic check (with python execution code). I can add information in docstring of schema, and it will be visible in generated doc (i'm doing it).

Issue #2828, opened 2020-04-07
Add Tracim Events and Tracim Live Message related tables in database and fill them


Skylsmoi:
## Feature description

To support notification feature of tracim, we do need two new database tables:

- [x] `event`

| name       | type     | comment                                                     |
|------------|----------|-------------------------------------------------------------|
| event_id   | integer  |                                                             |
| event_type | string   |                                                             |
| date       | datetime |                                                             |
| fields     | JSON     | information needed to display a message (related ids+names) |

- [x] `message` : same as event but related to someone and with a read/unread status.

| name        | type     | comment                                                   |
|-------------|----------|-----------------------------------------------------------|
| event_id    | integer  | foreign key to event                                      |
| receiver_id | integer  | user id that is the recipient of the message              |
| sent_date   | datetime | when the message was sent                                 |
| read_date   | datetime | initially NULL, used to confirm the message has been read |

And a generic mechanism  in Tracim for creating events in the database :

- [x] it will observe the database entities changes and generate events based on those
- [x] the contents of the `fields` column shall be enough for the frontend to display the messages without needing to query the backend (messages are "self-sufficient")
- [x] The generation and publication of messages (using #2837) will be delegated to a RQ job called "TLMBuilder"
- [x] the new mechanism has to take #1589 into account: it is not needed to do it in this issue, but it must provide a satisfactory solution for it.

## Implemention summary

- entity events are generated when their respective objects are flushed to the database by using sqlalchemy's event mecanism. This solution has been chosen because it ensures all changes are captured at the right time as the various api classes have rather complex behavior regarding save parameters (see #3035). When this complexity has been refactored and an access to the plugin manager has been provided to the apis (#1834) we can think of moving those hook directly in the api classes.
- `pluggy` hooks are used to propagate events (`on_user_created`/... hooks)
- Event creation is done by one class `EventBuilder` which also post a job in RQ for message creation & dispatch
- Message creation is done by `LiveMessageBuilder`, currently not integrated in a RQ worker (waiting for #2991).
- Endpoint for retrieving messages has been written (`MessageController` class).

Skylsmoi:
@inkhey Can you edit this issue description please ?

inkhey:
Suggestion of database table:

For Tracim Event:

| Field Name | Type                                  | Others caracteristics                                       | goal                                                    |
|------------|---------------------------------------|-------------------------------------------------------------|---------------------------------------------------------|
[...]

PR #3022, opened 2020-05-11
2828  event tlm database

grignards:
Refs #2828 

## Checkpoints

**For code reviewers**

**For developers**

**For the quality team**


backend/tracim_backend/lib/core/user.py
```diff
@@ -148,7 +148,7 @@ def get_all(self) -> typing.Iterable[User]:
 
     def get_user_ids_from_profile(self, profile: Profile) -> typing.Iterable[int]:
         query = self._apply_base_filters(self._session.query(User.user_id))
-        return query.filter(User.profile == profile).all()
+        return [res[0] for res in query.filter(User.profile == profile)]
```
inkhey, line 151: Not sure to understand the meaning of this change. Can you explain @grignards ?

grignards: SQLAlchemy returns tuples, even when only one column is asked for (like it is here), so I extract the ids from their tuples.

inkhey: ok

backend/tracim_backend/command/__init__.py
```diff
@@ -63,6 +63,9 @@ def take_action(self, parsed_args: argparse.Namespace) -> None:
                         self.take_app_action(parsed_args, app_context)
         except Exception as exc:
             logger.exception(self, exc)
+            import traceback
```
inkhey, line 66: logger.exception(self, exc) is not enought to get traceback ?

backend/tracim_backend/lib/event.py
```diff
@@ -0,0 +1,287 @@
...
+            receiver_ids.add(event.user["user_id"])
+        return receiver_ids
+
+    @classmethod
+    def _get_workspace_event_receiver_ids(cls, event: Event) -> typing.Set[int]:
+        user_api = UserApi(current_user=None, session=cls._session(), config=cls._config())
+        administrators = user_api.get_user_ids_from_profile(Profile.ADMIN)
+        role_api = RoleApi(current_user=None, session=cls._session(), config=cls._config())
+        workspace_members = role_api.get_workspace_member_ids(event.workspace["workspace_id"])
+        return set(administrators).union(set(workspace_members))
```
raphj, line 287:
```suggestion
        return set(administrators).union(workspace_members)
```

```suggestion
        return set(administrators + workspace_members)
```
(works with Python lists, not sure about the types of `administrators` and `workspace_members`)

backend/tracim_backend/lib/event.py
```diff
@@ -0,0 +1,287 @@
...
+        self._current_user = None
+
+    # User events
+    @hookimpl
+    def on_user_created(self, user: User, db_session: TracimSession) -> None:
+        self._create_user_event(OperationType.CREATED, user, db_session)
+
+    @hookimpl
+    def on_user_modified(self, user: User, db_session: TracimSession) -> None:
+        if user.is_deleted:
```
raphj, line 98: Are we sure there aren't edge cases where a deleted user is modified, making this function behave erratically (if I understand its logic correctly)? 

Same remark for all these functions using `something.is_deleted`

backend/tracim_backend/command/__init__.py
grignards: the logs do not appear in the tests, that's why I added this and I forgot it. I didn't check if logs are displayed while using `tracimcli` interactively.

backend/tracim_backend/models/event.py
```diff
@@ -0,0 +1,123 @@
+# -*- coding: utf-8 -*-
+"""
+Live messages/events related model.
+
+It's perfectly fine to re-use this definition in the tracim application,
+though.
```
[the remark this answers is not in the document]
grignards: Copy/paste… I'll remove this comment as it does not make sense here.

backend/tracim_backend/lib/event.py
inkhey: @grignards , is there any way to have some kind of diff between previews and new User at this level ?

grignards: Yes it is possible through [the attribute state](https://docs.sqlalchemy.org/en/13/orm/internals.html#sqlalchemy.orm.state.AttributeState.history)

grignards: Attribute state works as expected, I modified my code to ensure behavior (plus a unit test for the corner case of is_deleted==True then modified).

backend/tracim_backend/lib/event.py
grignards: I didn't know that `union` worked with a list as parameter, thank you for the tip.

backend/tracim_backend/lib/event.py
```diff
@@ -0,0 +1,287 @@
...
+        query = (
+            self._session.query(Message)
+            .filter(Message.receiver_id == user_id)
+            .options(joinedload(Message.event))
+        )
+        if read_status == ReadStatus.READ:
+            query = query.filter(Message.read != null())
+        elif read_status == ReadStatus.UNREAD:
+            query = query.filter(Message.read == null())
+        return query.all()
```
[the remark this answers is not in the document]
grignards: Else means `ALL` here and `ALL` means no filtering (so no action to take on the query).
I added a comment to explain it (plus an assert).

inkhey:
Found a  probably 3rd issue:
- [ ] standard http message endpoint is not paginated, need to discuss with frontend, but this is probably needed. Don't think this is a blocking point but a specific issue is needed, if need is confirmed by frontend (@Skylsmoi)

PR #3040, opened 2020-05-13
[For 2.7.4] Log authenticated user

grignards:
This PR adds a log when an HTTP request is made by an authenticated user.
The log includes the request method, path and user id.

Related to #3043

## Checkpoints

**For code reviewers**

**For developers**

**For the quality team**

### DIFF
--- a/backend/tracim_backend/lib/utils/authorization.py
+++ b/backend/tracim_backend/lib/utils/authorization.py
@@ -16,6 +16,7 @@ from tracim_backend.exceptions import InsufficientUserRoleInWorkspace
 from tracim_backend.exceptions import TracimException
 from tracim_backend.exceptions import UserGivenIsNotTheSameAsAuthenticated
 from tracim_backend.exceptions import UserIsNotContentOwner
+from tracim_backend.lib.utils.logger import logger
 from tracim_backend.lib.utils.request import TracimContext
 from tracim_backend.models.auth import Profile
 from tracim_backend.models.roles import WorkspaceRoles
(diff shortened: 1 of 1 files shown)


lebouquetin:
@grignards I suggest you put the data in a JSON format, what do you think ? It may be easier for parsing logs?

PhilippeAccorsi:
I'm not sure but all new log starting with date like other log, or not ?

inkhey:
> I'm not sure but all new log starting with date like other log, or not ?

It's the same logger as other tracim, so yes, similar format with date.

PhilippeAccorsi:
> > I'm not sure but all new log starting with date like other log, or not ?
> 
> It's the same logger as other tracim log, so yes, similar format with date.

Ok, great.

grignards:
> @grignards I suggest you put the data in a JSON format, what do you think ? It may be easier for parsing logs?

Here the proper solution would be to have structured logs, but as the current tracim logging outputs lines of text I think parsing the lines will be enough.
They are output in this form:
```
2020-05-13 09:00:52 INFO  [tracim:45][waitress] [TracimRequest] GET /api/v2/users/me/workspaces/1/contents/read_status from authenticated user 1
```

lebouquetin:
@grignards ok. I suggest at least a specific keyword which will allow to distinguish these logs from others (to be sure not parsing some other log that match the pattern).

For example something like: 

```
2020-05-13 09:00:52 INFO  [tracim:45][waitress] [TracimRequest] GET /api/v2/users/me/workspaces/1/contents/read_status from authenticated user 1 #USERACTIVITY`
```

PhilippeAccorsi:
Logging have same syntax with uwsgi ? (production use uwsgi on docker image).

grignards:
> Logging have same syntax with uwsgi ? (production use uwsgi on docker image).

Yes from what I gathered with the log file you provided me.

grignards:
In fact the `[TracimRequest]` part is already specific to those logs so we can use it to differentiate them.

lebouquetin:
> In fact the [TracimRequest] part is already specific to those logs so we can use it to differentiate them.

I missed this. :+1:

PhilippeAccorsi:
Test not to be necessary for this fix.

PhilippeAccorsi:
@grignards you can squash and merge. Add `Ref: #3043` in message commit.

Issue #2837, opened 2020-04-07
Implement TLM server side with message bus


Skylsmoi:
## Feature description

TLM means Tracim Live Message. TLMs are the way the server will notify frontend of new contents. This will be based on long polling / websocket or server-side events.
We do need a technology allowing a "message bus" where the server can directly push the information to the client.

- [x] choose a technology allowing a message bus, most probably a library or a server software allowing multiple technology at the same time among polling, websocket and server-sent events for better compatibility
- [x] define object received by the frontend (we most probably do not need to define a bidirectionnal communication protocol for websocket if we do rely on improved technology that does this for us)
- [X] implement it

## Context

The frontend needs to know as soon as possible when there is a new notification or a new todo. It does not need to have to ask for it.

## Impacted components

backend

## Extra information

First step is to define the structure of the message. This step is critical as it is blocker for frontend implementation
Regarding the communication protocol, I see 2 main messages:
- new notification
- new todo

# Implemented Solution
see PR https://github.com/tracim/tracim/pull/3041

- use pushpin as TLM bus using SSE protocol and standard tracim auth authentification mecanism.
- documentation about using pushpin for dev env and to run standalone pushpin
- automatic test with and without pushpin for sse endpoint
- sse command tester for debugging.

This Pr implement also simple implementation in frontend.

:warning: This make sse stream working if pushpin is enabled, but does not make sse stream working in production and docker,
few issue are related to better integrate and documentation this feature in backend to make it default:
- https://github.com/tracim/tracim/issues/2901
- https://github.com/tracim/tracim/issues/3053
- https://github.com/tracim/tracim/issues/3062
- https://github.com/tracim/tracim/issues/3055

inkhey:
About technical choice for TracimLiveMessage bus:

> 
> In the way, i see the process, we do need a piece of software that is notified that some connected user need to receive a notification and return this information into the correct socket/sse/other.
> 
> We can implement something custom here using asyncio but it may be complex. Another solution can be using Nchan. Nchan is a pubsub server, that do support itself, websocket, sse and many other way to communicate event efficiently, it is an extension from nginx, so we do need nginx.
> https://nchan.io/
> 
> Pro:
> - scalable
> - less code
> - no need to worry in backend about integrating complicate stuff
> - efficient
> - mature
> - seems not too complicate
> - Handle itself many way to communicate, so out of the box support for websocket, sse and more.
> - existing lib for frontend: https://www.npmjs.com/package/nchan
[...]

PR #3041, opened 2020-05-13
Feature/2837  implement pushpin for events

inkhey:
:warning: Wait for simple implementation in frontend before merging this.

related to #2837 

Only for simple code review now, need sync with https://github.com/tracim/tracim/pull/3022 PR first.

## Checkpoints

**For code reviewers**

**For developers**

**For the quality team**


backend/tracim_backend/views/core_api/user_controller.py
```diff
@@ -542,6 +543,26 @@ def disable_workspace_notification(self, context, request: TracimRequest, hapic_
         wapi.save(workspace)
         return
 
+    @hapic.with_api_doc(tags=[SWAGGER_TAG__USER_CONTENT_ENDPOINTS])
+    @check_right(has_personal_access)
+    def live_messages(self, context, request: TracimRequest, hapic_data=None) -> Response:
+        """
+        Tracim Live Message Events as ServerSide Event Stream
+        """
+        user_channel_name = "user_{}".format(request.candidate_user.user_id)
```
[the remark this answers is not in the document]
inkhey: it should be "candidate user" as this endpoint could be used by an administrator:
candidate_user -> user from user_id in the endpoint.
current_user -> user authenticated.

To clarify:
- /users/me/live_messages -> should always use current_user as candidate_user doesn't exist
- /users/1/live_messages -> should always use candidate_user as current_user may be user 2 which is administrator.

pushpin_dev.conf
```diff
@@ -0,0 +1,63 @@
+[global]
```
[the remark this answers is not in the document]
inkhey: There is very few difference. I'm not sure about the strategy to use there. I managed to have different port/dir usage for :
- default pushpin config.
- pushpin dev config. 
- pushpin test config.

This way we are sure to not have independant process for pushpin daemon, pushpin dev usage and pushpin process runned automatically for test.

backend/tracim_backend/views/core_api/schemas.py
```diff
@@ -1309,3 +1309,7 @@ class ConfigSchema(marshmallow.Schema):
     workspace_size_limit = marshmallow.fields.Integer()
     workspaces_number_per_user_limit = marshmallow.fields.Integer()
     instance_name = marshmallow.fields.String()
+
+
```
raphj, line 1313: Maybe add a TODO here?

inkhey: done.

inkhey:
> I can't really review this, I don't have the required knowledge.
> The part that outputs event streams seems correct and the code looks good though.
> 
> However, I think you need to:
> 
>     * [ ]  check the Accept HTTP header from the client, it needs to accept text/event-stream
> 
>     * [ ]  close the connection with the correct HTTP 40X status if it's not present in the Accept header
> 
>     * [ ]  add a test for this

see #3042

backend/tracim_backend/lib/core/live_messages.py
```diff
@@ -0,0 +1,54 @@
...
+    def _convert_to_sse_string(self):
+        if self.event:
+            self._buffer += "event: {}\n".format(self.event)
+        for row in self.data:
+            self._buffer += "data: {}\n".format(row)
+
+        self._buffer += "\n"
+
+    def __str__(self):
+        return self._buffer
```
grignards, line 35: This should be simpler with a simple function, isn't it? It could also do the 'json.dumps()' call with the right arguments to have only one `data:` line.

inkhey: Idea here, is just to have an simple unopiniated implementation of sse protocol.

I don't really like the idea of having a method that does thing at multiple protocol layer at once but that can be discussed as the protocol is really simple here.

inkhey: i adjust the code a bit.

backend/tracim_backend/lib/core/event.py
```diff
@@ -215,15 +216,17 @@ def _publish_event(self, db_session: TracimSession, instance: object) -> None:
...
-        redis_connection = get_redis_connection(self._config)
-        # TODO S.G 2020-05-12: change is_async to use workers
-        queue = get_rq_queue(redis_connection, RQ_QUEUE_NAME, is_async=False)
-        logger.debug(self, "publish event {} to RQ queue {}".format(event, RQ_QUEUE_NAME))
         LiveMessageBuilder.session = db_session
         LiveMessageBuilder.config = self._config
-        queue.enqueue(
-            LiveMessageBuilder.publish_messages_for_event, self._event_schema.dump(event).data,
-        )
+        # TODO - G.M - 2020-05-15 - this parameter should be renamed, it's not anymore email related
```
raphj, line 221:
```suggestion
        # TODO - G.M - 2020-05-15 - this parameter should be renamed, it's not email-related anymore
```

backend/tracim_backend/lib/core/live_messages.py
raphj: Uh oh, there is a bug here that I didn't see!
why is data a list? What if there is a string with new lines in this list? Why not just have a data string?

raphj: but I agree that this code should not do anything related to json here

inkhey:
:warning: warning Wait for simple implementation in frontend before merging this.

backend/tracim_backend/lib/core/live_messages.py
inkhey: There is a data list to allow multiline sse, because it's an unopiniated implementation, but i can remove this, if needed.

raphj: IMHO you should accept a regular string and split it by new lines

Skylsmoi:
This PR has been merged with known issues and failing tests.

It will be fixed soon, we required merging to allow frontend dev to start working on the live message refactors

Issue #3051, opened 2020-05-14
Feat: Create Account endpoint for live_messages


inkhey:
## Feature description

We need to add "/users/me" account endpoint for lives messages features aside of existing "/users/{user_id}" 

This can be fixed either with code duplication or solving technical debt issue #1478.

grignards:
What are the pros of those `me` endpoints compared to the `{user_id}` ones ?
Because:
- they duplicate code in backend
- they aren't RESTful (for example it wouldn't be possible to cache them as their results depend on the current user)

lebouquetin:
My feeling is `/users/me` may be a redirect to `/users/{user_id}`, no?

lebouquetin:
@grignards the reason for having /users/me is "ease of use". APIs are to be used by not-really-developers and the best way to publish an API is to allow to easily go to your target.

For exemple, if you are not yet authenticated, you can't access your profile so you do not know your user id

inkhey:
tried refactoring /me endpoint using 307 redirection:
https://github.com/tracim/tracim/pull/3085

Some unclear questions:
- how frontend handle redirection 307, does it try to repost get parameters, post parameters, body ? depending on how frontend deal with it, it can break current behavior. After rapid check it's look like get parameters are not passed, so probably the same for others.
- should we convert get parameters into location using redirect 307 ?

That's why, i do suggest frontend to be refactored in order to not use /me endpoint if we follow this idea.

grignards:
> tried refactoring /me endpoint using 307 redirection:
> #3085
> 
> Some unclear questions:
> 
>     * how frontend handle redirection 307, does it try to repost get parameters, post parameters, body ? depending on how frontend deal with it, it can break current behavior. After rapid check it's look like get parameters are not passed, so probably the same for others.

The [Mozilla doc](https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections) on this topic seems to suggest that the body and method of the request are preserved when executing the redirection, this would mean the full URI (including query parameters) should be returned in the `Location` header.

I think the redirection *should* be handled transparently by the browser, without needing javascript code for this, but ask @Skylsmoi to be sure.

>     * should we convert get parameters into location using redirect 307 ?
> 
> 
> That's why, i do suggest frontend to be refactored in order to not use /me endpoint if we follow this idea.

Skylsmoi:
I have tested it and 307 redirections are handled nicely by browser. Front doesn't have to do anything.

[...]

Issue #1478, opened 2019-03-14
refactor account endpoint code


inkhey:
We have both `/api/v2/users/{user_id}/... `and `/api/v2/users/me/...` endpoints we should see how we can avoid duplication of code there : 

**1rst solution : reroute to same internal code:**
- need 2 right checker for both url, there are also different actually (checker need to be more complex if we want only one there)
- need to be explicit for all input and output schema if we want check !
- implicit redirection

example (untested code) :
```
   def _route_calendars(context, request, hapic_data, user):
        app_config = request.registry.settings['CFG']
        calendar_api = CalendarApi(
            current_user=user,
            session=request.dbsession,
            config=app_config,
        )
        return calendar_api.get_user_calendars(
            user,
            workspaces_ids_filter=hapic_data.query.workspace_ids
        )

    @hapic.with_api_doc(tags=[SWAGGER_TAG__USER_CALENDAR_ENDPOINTS])
    @check_right(has_personal_access)
    @hapic.input_path(UserIdPathSchema())
    @hapic.input_query(CalendarFilterQuerySchema())
    @hapic.output_body(CalendarSchema(many=True))
    def user_calendars(self, context, request: TracimRequest, hapic_data=None):
        user = request.candidate_user
        return self._route_calendar(context,request, hapic_data, user)

    @hapic.with_api_doc(tags=[SWAGGER_TAG__USER_CALENDAR_ENDPOINTS])
    @check_right(is_user)
    @hapic.input_query(CalendarFilterQuerySchema())
    @hapic.output_body(CalendarSchema(many=True))
    def account_calendars(self, context, request: TracimRequest, hapic_data=None):
        user = request.current_user
        return self._route_calendar(context,request, hapic_data, user)
```

**2nd solution : use redirection :** 

- no need to check right here
- no need to explain input and output schema.
- able to document input schema if we want to document better
- output schema always the same.
- explicit redirection
- able to redirect many endpoints with only one "catch all redirection" if documentation is not needed here.

example with already existing endpoint:
```
    @hapic.with_api_doc(tags=[SWAGGER_TAG__CONTENT_ENDPOINTS])
    @hapic.input_path(ContentIdPathSchema())
    @hapic.output_body(NoContentSchema(), default_http_code=HTTPStatus.FOUND)  # nopep8
    def get_content(
            self,
            context,
            request: TracimRequest,
            hapic_data=None,
    ) -> None:
        """
[...]

PR #3085, opened 2020-05-19
refactor account me endpoint with redirection

inkhey:
related to #1478 and #3051
## Checkpoints

**For code reviewers**

**For developers**

**For the quality team**


backend/tracim_backend/views/core_api/account_controller.py
```diff
@@ -50,10 +28,8 @@ def account_route_get(self, context, request: TracimRequest, hapic_data: HapicDa
         without giving directly user id.
         This route generate a HTTP 307 with the right url
         """
-        return redirect_account_route(
-            user_id=request.current_user.user_id,
-            path_suffix=request.matchdict.get("path_suffix"),
-            get_params=request.GET,
+        return HTTPTemporaryRedirect(
+            request.url.replace("/me", "/{user_id}".format(user_id=request.current_user.user_id), 1)
```
inkhey, line 32: Just converted the redirection from a "rebuild url" model to a more simpler approach with replace. @grignards @buxx @raphj which approach did you prefer here, and did you see any issue with this approach (possibly proxy related) ?

inkhey:
:warning: #3089 should probably be merged before this one to simplify merge.

backend/tracim_backend/views/core_api/account_controller.py
```diff
@@ -52,300 +21,28 @@
...
-    @hapic.input_body(SetEmailSchema())
-    @hapic.output_body(UserSchema())
-    def set_account_email(self, context, request: TracimRequest, hapic_data=None):
+    @hapic.input_path(PathSuffixSchema())
+    def account_route_get(self, context, request: TracimRequest, hapic_data: HapicData):
         """
-        Set user Email
+        Convenient route allowing to use PUT /api/v2/users/{user_id}/* endpoint with authenticated user
+        without giving directly user id.
+        This route generate a HTTP 307 with the right url
```
raphj, line 29:
```suggestion
        without directly giving a user id.
        This route generates an HTTP 307 response with the right URL
```

backend/tracim_backend/views/core_api/account_controller.py
raphj: You should probably replace the whole prefix, to avoid risky random unrelated replacements in the url, no?

backend/tracim_backend/views/core_api/account_controller.py
```diff
@@ -52,300 +21,28 @@
...
         """
-        app_config = request.registry.settings["CFG"]  # type: CFG
-        api = ContentApi(
-            show_archived=True,
-            show_deleted=True,
-            current_user=request.current_user,
-            session=request.dbsession,
-            config=app_config,
+        return HTTPTemporaryRedirect(
+            request.url.replace("/me", "/{user_id}".format(user_id=request.current_user.user_id), 1)
```
raphj, line 44: Duplicate code (including the comment), maybe perform this replacement in a dedicated function?

backend/tracim_backend/views/core_api/account_controller.py
inkhey: you are right, according to the code, as we add route with` /users/me{path_suffix:.*} ` (and the prefix for api is done in another place,
The best of both approach seems:

```
request.url.replace("/users/me", "/users/{user_id}".format(user_id=request.current_user.user_id), 1)
```
This way we don't mix thing that are not related (prefix for api for example), and we do maximize risk.

backend/tracim_backend/views/core_api/account_controller.py
inkhey: That's not completely duplicated code, one method is for put and one other for get. This is currently required to do this way (2 method) to make hapic documentation work correctly. Not sure a dedicated function is a good idea here as we gain just few line of code, and this make code less straightforward.

backend/tracim_backend/views/core_api/account_controller.py
grignards: Or even replace `/users/me/` to be sure (even if it is unlikely we'll use a route like `/users/meow` :wink:

inkhey: That's not really possible as `/users/me` endpoint exist.

backend/tracim_backend/views/core_api/account_controller.py
```diff
@@ -52,300 +21,32 @@
...
-        before_content = None
-        if content_filter.before_content_id:
-            before_content = api.get_one(
-                content_id=content_filter.before_content_id,
-                workspace=workspace,
-                content_type=content_type_list.Any_SLUG,
+        return HTTPTemporaryRedirect(
+            request.url.replace(
+                "/users/me", "/users/{user_id}".format(user_id=request.current_user.user_id), 1
             )
```
raphj, line 48: Still not convinced.

What about something like (seems more robust):

```suggestion
        redirection = "/users/{user_id}".format(user_id=request.current_user.user_id)
        if request.url = "/users/me":
            newurl = redirection
        else
           newurl = re.sub(r"^/users/me/", redirection + "/", request.url)
        return HTTPTemporaryRedirect(newurl, 1)
```

(:warning this suggestion requires importing the module `re`)

inkhey: Don't sure to understand your proposal, but this cannot work.
request.url does contain the full url, not just the ending path.

inkhey:
for quality test:
- just rapid test of frontend and check if no error during usage.

Issue #3055, opened 2020-05-14
Feat: Use pushpin in docker in order to support live messages


inkhey:
## Feature description

Add support for live messages feature in docker too: add pushpin into the docker architecture.
This probably require #3054 and #3053  to be clarify before.

Issue #2785, opened 2020-03-25
Prepare new release 3.0.0


PhilippeAccorsi:
## Feature description

- [x] update frontend version
- [x] update backend version
- [x] update changelog

PhilippeAccorsi:
Add some information about databse support. https://github.com/tracim/tracim/issues/697

inkhey:
:warning: 
Api breaking changes
- remove of /v2 :  https://github.com/tracim/tracim/issues/3052
- /users/me endpoint are now real redirection (return http 307):

Other big change:
- pushpin mandatory, this will change multiples things in the way to setup tracim for both development and production environments, to be clarify when we fix all the documentation/implementation issue about using pushpin in production.

grignards:
:warning: 
Backend configuration and deployment changes:

- renaming of `email.processing_mode` and `email.async.*` configuration parameters to `jobs.xxx` : #3100 
- `async` mode needs a new process to be executed, details in https://github.com/tracim/tracim/blob/develop/backend/README.md#run-daemons (the `rq worker ...` line).

grignards:
:warning: 

- apache/ configuration change to route live messages endpoint through pushpin:

```
ProxyPassMatch /api/(v2/users/.*/live_messages)$ http://127.0.0.1:7999/api/$1
```

PhilippeAccorsi:
> ProxyPassMatch /api/(v2/users/.*/live_messages)$ http://127.0.0.1:7999/api/$1

We need to chnage this with fix about remove/change `v2/` ?

grignards:
> > ProxyPassMatch /api/(v2/users/.*/live_messages)$ http://127.0.0.1:7999/api/$1
> 
> We need to change this when fix about remove/change `v2/` to other thing is merge ?

I updated my comment (and the docker PR).

PhilippeAccorsi:
> * renaming of `email.processing_mode` and `email.async.*` configuration parameters to `jobs.xxx` : #3100
> 
> * `async` mode needs a new process to be executed, details in https://github.com/tracim/tracim/blob/develop/backend/README.md#run-daemons (the `rq worker ...` line).

Add breaking change (I think) with parameter name change from `email.....` to `jobs.....`.

PhilippeAccorsi:
Before 3.0.0, docker just show `tail -f /var/log/dpkg.log` when running so docker logs size not increase when running tracim.

[...]

PR #3131, opened 2020-06-03
feat(backend): integrate pushpin and custom rq worker for TLMs

grignards:
Also improved docker build (split in several RUN commands)

Refs #3055.

## Checkpoints

**For code reviewers**

**For developers**

**For the quality team**

### DIFF
--- a/backend/doc/apache.md
+++ b/backend/doc/apache.md
@@ -41,6 +41,7 @@ Create a file named `/etc/apache2/sites-available/tracim.conf` containing:
         ProxyPass /ui http://127.0.0.1:6543/ui
         ProxyPassReverse /ui http://127.0.0.1:6543/ui
 
+        ProxyPassMatch /api/(v2/users/.*/live_messages)$ http://127.0.0.1:7999/api/$1
         ProxyPass /api http://127.0.0.1:6543/api
         ProxyPassReverse /api http://127.0.0.1:6543/api
 
--- a/backend/doc/apache.md
+++ b/backend/doc/apache.md
@@ -41,7 +41,9 @@ Create a file named `/etc/apache2/sites-available/tracim.conf` containing:
         ProxyPass /ui http://127.0.0.1:6543/ui
         ProxyPassReverse /ui http://127.0.0.1:6543/ui
 
-        ProxyPassMatch /api/(v2/users/.*/live_messages)$ http://127.0.0.1:7999/api/$1
+        ProxyPassMatch /api/users/(.*/live_messages)$ http://127.0.0.1:7999/api/users/$1
+        ProxyPassReverse /api/users/(.*/live_messages)$ http://127.0.0.1:7999/api/users/$1
+
         ProxyPass /api http://127.0.0.1:6543/api
         ProxyPassReverse /api http://127.0.0.1:6543/api
 
--- a/tools_docker/Debian_Uwsgi/Dockerfile
+++ b/tools_docker/Debian_Uwsgi/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:stretch
+FROM debian:buster
 MAINTAINER contact@tracim.fr
 
 ARG TAG=""
@@ -13,22 +13,12 @@ ENV LANG C.UTF-8
 ENV LANGUAGE C.UTF-8
 ENV LC_ALL C.UTF-8
 
-
 RUN if ([ "$TAG" != "" ] && [ "$BRANCH" != "" ]); \
     then echo "Its not possible to use ARG TAG and ARG BRANCH in same time." && exit 1;fi
 
-RUN \
-# Create command to clone repo
-    get_source_from_github() { \
-    if ([ "$TAG" != "" ] && [ "$BRANCH" = "" ]); \
-    then git clone -b $TAG --single-branch "$REPO" tracim; \
-    elif ([ "$TAG" = "" ] && [ "$BRANCH" != "" ]); \
-    then git clone "$REPO" tracim && cd tracim/ && git checkout $BRANCH && cd -; \
-    else git clone "$REPO" tracim; \
-    fi \
-    } \
 # Install required packages
-    && apt update && apt upgrade -qy \
+RUN \
+    apt update && apt upgrade -qy \
     && apt install -qy \
     apache2 \
     build-essential \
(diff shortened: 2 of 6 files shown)


tools_docker/Debian_Uwsgi/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM debian:stretch
+FROM debian:buster
```
inkhey, line 1: Is this required ?
I agree to do this, but this also mean that when stretch support will break we will not be able to know it.
I suggest discuss this with @PhilippeAccorsi .

tools_docker/Debian_Uwsgi/supervisord_tracim.conf
```diff
@@ -9,13 +9,21 @@ logfile=/var/tracim/logs/supervisord.log ; (main log file;default $CWD/superviso
 pidfile=/var/run/supervisord.pid ; (supervisord pidfile;default supervisord.pid)
 childlogdir=/var/tracim/logs/            ; ('AUTO' child log dir, default $TEMP)
 
+
+[program:xvfb]
```
inkhey, line 13: Nice idea :+1:

tools_docker/Debian_Uwsgi/supervisord_tracim.conf
```diff
@@ -9,13 +9,21 @@ logfile=/var/tracim/logs/supervisord.log ; (main log file;default $CWD/superviso
...
+
 ; email notifier (if async email notification is enabled)
 [program:tracim_mail_notifier]
 directory=/tracim/backend/
 command=python3 /tracim/backend/daemons/mail_notifier.py
-stdout_logfile =/var/tracim/logs/mail_notifier.log
+stdout_logfile=/var/tracim/logs/mail_notifier.log
 redirect_stderr=true
-autostart=false
+autostart=true
```
inkhey, line 26: Not sure to understand, this mean that  mail notifier is activated even if mail notification is disabled ?

PhilippeAccorsi:
- Need also information for changelog about apache conf need to be updated.

tools_docker/Debian_Uwsgi/supervisord_tracim.conf
PhilippeAccorsi: We start all the time this service and I think we not stop it if email is not activated (why ? I think its a mistake)

PhilippeAccorsi:
>     * Need also information for changelog about apache conf need to be updated.

Add information here about apache modification : https://github.com/tracim/tracim/issues/2785

grignards:
> @grignards
> 
>     * why you add multiple RUN ? Its increase size of docker image without any interest.

It avoids to rebuild everything each time you modify one line in the Dockerfile as the build cache will be used. But yes this leads to an increased image size (I read the best practices too quickly). I will change this.

>     * I'm not remember that we have decided to change stretch to buster now (issue exist about that I think)

Fair remark, I didn't check if there was an issue. In general I prefer to keep the production environment relatively up-to-date regarding to development/testing environment as most of the tests are done in those (for example travis uses ubuntu bionic which is based on debian buster). If you think this is too early/that this change will cost you time I'll revert it (didn't take too long anyway).

>     * If I remember correctly, installing nodejs and npm like this not work like expected `nodejs npm` (without `curl -sL https://deb.nodesource.com/setup_10.x | bash -`)

With stretch the nodejs version was really too old (4.8.2), buster provides a 10.19 version which should be ok I think (the image builds and runs).

>     * what is going on if we use environnement variable to set `basic_setup.website_base_url`, pushpin can take value of environnement variable ?

As pushpin is configured, it will redirect any request to tracim's web application so I think the `base_url` parameter has no effect on pushpin's configuration. Furthermore pushpin is only used for the `live_messages` endpoint.

tools_docker/Debian_Uwsgi/supervisord_tracim.conf
grignards: This service was always started in the develop version of the entrypoint.sh script, so I didn't change this behavior. The optional service was/is `tracim_mail_fetcher` (only launched if the `REPLY_BY_EMAIL` environment variable is set).

grignards:
> > ```
> > * Need also information for changelog about apache conf need to be updated.
> > ```
> 
> Add information here about apache modification : #2785

Done.

tools_docker/Debian_Uwsgi/supervisord_tracim.conf
inkhey: Ok, fix this here if this is not too complicated or create an issue about this problem.

backend/doc/apache.md
```diff
@@ -41,6 +41,7 @@ Create a file named `/etc/apache2/sites-available/tracim.conf` containing:
         ProxyPass /ui http://127.0.0.1:6543/ui
         ProxyPassReverse /ui http://127.0.0.1:6543/ui
 
+        ProxyPassMatch /api/(v2/users/.*/live_messages)$ http://127.0.0.1:7999/api/$1
```
inkhey, line 44: Does proxyPassMatch does same as ProxyPass + ProxyPassReverse ?

inkhey: note: should be updated later (/v2 will be removed)

grignards: > Does proxyPassMatch does same as ProxyPass + ProxyPassReverse ?

It is not clear if the ProxyPassReverse directive works with regular expressions (and there is no ProxyPassMatchReverse): http://httpd.apache.org/docs/2.2/mod/mod_proxy.html#proxypassreverse

I just found an [article](https://botproxy.net/docs/how-to/how-to-write-regex-for-apache-proxypassmatch-to-reverse-proxy-api-calls/) indicating that it would work so we can try. I'll try.

inkhey: Ok, please notify if there is some change to be done, we do need to update tracim apache documentation too:
https://github.com/tracim/tracim/pull/3137

tools_docker/Debian_Uwsgi/supervisord_tracim.conf
grignards: I can, but what should be the fix ? Starting `tracim_mail_notifier` only when `REPLY_BY_EMAIL` is set ?

inkhey: No, reply by email is another feature, you should start tracim_mail_notifier when email_notification is enabled (and also normally when notification is in async mode).

tools_docker/Debian_Uwsgi/Dockerfile
```diff
@@ -1,5 +1,15 @@
-FROM debian:buster AS builder
-MAINTAINER contact@tracim.fr
```
inkhey, line 2: any reason to remove maintainer ?

tools_docker/Debian_Uwsgi/Dockerfile
grignards: Reverted to using stretch.

tools_docker/Debian_Uwsgi/supervisord_tracim.conf
grignards: Created issue #3142 for this.

tools_docker/Debian_Uwsgi/entrypoint.sh
```diff
@@ -143,4 +136,4 @@ if [ "$START_CALDAV" = "1" ]; then
     cd /tracim/backend/
     tracimcli caldav sync -c /etc/tracim/development.ini
 fi
-tail -f /var/log/dpkg.log
+tail -f /var/log/uwsgi/app/tracim_web.log
```
PhilippeAccorsi, line 139: Why you change this ?

Uwsgi log is not store in container. We have done this `tail -f /var/log/dpkg.log` to make container running. all necessary log is outside container.

grignards: > Why you change this ?
> 
> Uwsgi log is not store in container. We have done this `tail -f /var/log/dpkg.log` to make container running. all necessary log is outside container.

I think it makes more sense to block displaying the main applications's log file rather than an unrelated file. When running the container in foreground you directly see the log and it makes it available with `docker logs`.

PhilippeAccorsi: You're right. Maybe its more simple to understand that this log is available in volume if we make `tail -f /var/tracim/logs/tracim_web.log`, what do you think ?

PhilippeAccorsi: > You're right. Maybe its more simple to understand that this log is available in volume if we make `tail -f /var/tracim/logs/tracim_web.log`, what do you think ?

`/var/tracim/` in mounted when you start docker (docker run or docker-compose).

grignards: You're right, I will change to `/var/tracim/logs/tracim_web.log`.

inkhey:
There is currently no automated test on docker, so i checked. An issue about the fact does exist.

Issue #3102, opened 2020-05-29
Add unit tests for TLM handlers at frontend_app_html-document


gbonaspetti:
## Bug description & symptom

Need to add some unit tests for the TLM handlers at frontend_app_html-document.

## Prerequisites

#3066 

## Impacted components

- frontend_app_html-document

PR #3138, opened 2020-06-04
Fix/3102  add unit tests to frontend_app_html_document

gbonaspetti:
#3102 

## Checkpoints

**For code reviewers**

**For developers**

**For the quality team**


frontend_app_html-document/test/container/HtmlDocument.spec.js
```diff
@@ -0,0 +1,220 @@
...
+              ...contentHtmlDocument.htmlDocument,
+              label: 'newContentName'
+            }
+          }
+
+          before(() => {
+            wrapper.instance().handleContentModified(tlmData)
+          })
+
+          it('should be updated with the content modified', () => {
```
Skylsmoi, line 107: what should be updated ? I don't get this test name as I read it like so:
handleContentModified
  modify the content name should be updated with the content modified

I think it should be something like `it('should change the property content of the state', () => {`

Skylsmoi: do the same for similar tests bellow

Issue #2850, opened 2020-04-08
Refactor frontend update with new real time events feature


Skylsmoi:
### TLM implementation DONE
- Account.jsx
- AdminAccount.jsx
- Dashboard.jsx
- Header.jsx
- Home.jsx
- Login.jsx (entry point of Tracim, must connect to SSE)
- ReduxTlmDispatcher.jsx (new component)
- Sidebar.jsx (workspace.modified, workspace_role...)
- SearchResult.jsx
- Tracim.jsx
- WorkspaceContent.jsx

### new unit tests
- reducer/currentWorkspace.js
- TLM for Dashboard
- reducer/user.js
- internal functions for Account
- frontend_lib/test/helper.spec.js (add serialize test)
- internal functions for Header
- internal functions for SearchResult

## Feature description

EDIT: this issue will handle the folder frontend/ only

Example with the save of the new comment in a content:
Currently: When saving a comment, after the api has returned 200, the front will reload the list of comments.
Expected: When saving a comment, after the api has returned 200, the front do nothing. And the backend will send a message through the real time feature that will tell the app to load the given comment. The app will then display it

Do this process for:
- new comment
- new version of content (revision)
- new content in a workspace
- new workspace
- role change in a workspace

**Actually, all data modified on server must be updated in the user interface after the TLM message is received.**

Frontend must have only one way for updating data shown on the interface.

The general behavior should/must be the following one:

- on first load / first view, get data from the API
- then all refresh of the data shown on the user interface should be triggered by TLM reception

## Impacted components

- frontend
- frontend_lib

NOTE: apps will be handled in specific issues

## Extra information

~~This feature is not the priority, we might choose to implement this incrementally~~
This feature is the base for switching to live update. It must be fully released for V3.0.0

## Prerequisite 
https://github.com/tracim/tracim/issues/2842

inkhey:
@Skylsmoi Before refactoring this, we do need some clarification about a potential "fallback" mode of tracim without real time message. If we do expect such fallback, it may be better do not delete the old update mecanism code yet.

lebouquetin:
@Skylsmoi @inkhey did you discuss about a fallback? This is a major point to know which technology we'll use for live messaging.

The fallback should not be an implementation of something very different - I mean the fallback should be very localized in code.

To be discussed in priority

@Skylsmoi the refactor about update from TLM instead of API request result must not be progressive. I changed the extra information. We can talk about it if you need it.

lebouquetin:
[...]

PR #3140, opened 2020-06-04
Feature/2850 handle TLM in workspace content list

metzgegu:
#2850

## Checkpoints

**For code reviewers**

**For developers**

**For the quality team**


frontend/src/container/WorkspaceContent.jsx
```diff
@@ -83,45 +88,118 @@ class WorkspaceContent extends React.Component {
...
-        this.buildBreadcrumbs()
-        if (!state.appOpenedType) this.setHeadTitle(this.getFilterName(qs.parse(props.location.search).type))
-        break
+  handleContentModified = data => {
+    if (this.props.currentWorkspace.id !== data.workspace.workspace_id) return
+    if (data.content.content_namespace === 'upload') {
+      this.props.dispatch(updateWorkspaceShareFolderContentList([
+        {
+          ...data.content,
+          ...(data.content.content_type === CONTENT_TYPE.FOLDER && { parent_id: SHARE_FOLDER_ID })
```
Skylsmoi, line 189: while this works, it is too technical because it uses a very advanced knowledge of the spread operator behavior

try to write it in a simpler way

frontend/src/container/WorkspaceContent.jsx
```diff
@@ -83,45 +88,118 @@ class WorkspaceContent extends React.Component {
...
+    this.setHeadTitle()
+  }
+
+  handleContentCreated = data => {
+    if (this.props.currentWorkspace.id !== data.workspace.workspace_id) return
+    if (data.content.content_namespace === 'upload') {
+      this.props.dispatch(addWorkspaceShareFolderContentList([
+        {
+          ...data.content,
+          ...(data.content.content_type === CONTENT_TYPE.FOLDER && { parent_id: SHARE_FOLDER_ID })
```
Skylsmoi, line 170: while this works, it is too technical because it uses a very advanced knowledge of the spread operator behavior

try to write it in a simpler way

frontend/src/reducer/currentWorkspace.js
```diff
@@ -160,7 +160,7 @@ export default function currentWorkspace (state = defaultWorkspace, action) {
         ...state,
         contentReadStatusList: state.contentReadStatusList.filter(id => id !== action.unreadId),
         recentActivityList: [
-          state.recentActivityList.find(content => content.id === action.unreadId),
+          ...(state.recentActivityList.find(content => content.id === action.unreadId) || []),
```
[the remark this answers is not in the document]
metzgegu: Yes true, this line was here to prevent `state.recentActivityList.find(content => content.id === action.unreadId` -> null, in order to prevent this I suggest to use `...state.recentActivityList.filter(content => content.id === action.unreadId)` so it will return [] if it can not find it

frontend/src/reducer/currentWorkspace.js
```diff
@@ -160,7 +160,7 @@ export default function currentWorkspace (state = defaultWorkspace, action) {
         ...state,
         contentReadStatusList: state.contentReadStatusList.filter(id => id !== action.unreadId),
         recentActivityList: [
-          ...(state.recentActivityList.find(content => content.id === action.unreadId) || []),
+          ...state.recentActivityList.filter(content => content.id === action.unreadId),
```
Skylsmoi, line 163: I'm confused, this seems to mean that recentActivityList can have the same contend id multiple times. I don't think it does

Skylsmoi: So I don't understand the purpose of this

metzgegu: The purpose was to return an empty array if the content is not in the `recentActivityList` instead of `null`
But I removed it and fix the problem with adding reducer in `currentWorkspace` for share content

metzgegu:
> I'm not sure to understand how adding those reducers fix the previous issue


This issue occurred only with shared contents, when a share content is created it triggers reducer in `workspaceShareFolderContentList` in order to add the content in redux but it also trigger the reducer `${REMOVE}/${WORKSPACE_READ_STATUS}` from `currentWorkspace`, this reducer manipulate the `recentActivityList` in order to move the content updated to the top of the list with the function `filter` but it return null because it was not added (recentActivity start with a null value which cause error in `Dashboard.jsx`).
It was not added because the file `currentWorkspace` don't have reducer for share content but only for normal content, so in order to have the same behaviour as normal content, I added those reducer for share content.

Skylsmoi:
> This issue occurred only with shared contents, when a share content is created it triggers reducer in `workspaceShareFolderContentList` in order to add the content in redux but it also trigger the reducer `${REMOVE}/${WORKSPACE_READ_STATUS}` from `currentWorkspace`, this reducer manipulate the `recentActivityList` in order to move the content updated to the top of the list with the function `filter` but it return null because it was not added (recentActivity start with a null value which cause error in `Dashboard.jsx`).
> It was not added because the file `currentWorkspace` don't have reducer for share content but only for normal content, so in order to have the same behaviour as normal content, I added those reducer for share content.

Ok, I understand.
Actually, the issue was in the currentWorkspace reducer, the actions handling the recentActivities couldn't add a content through TLM if it wasn't already in the list.
I have fixed it in the PR https://github.com/tracim/tracim/pull/3194
This PR will have to be upated once https://github.com/tracim/tracim/pull/3194 is merged though

frontend/src/container/ReduxTlmDispatcher.jsx
```diff
@@ -74,7 +83,17 @@ export class ReduxTlmDispatcher extends React.Component {
   }
 
   handleContentCreated = data => {
-    this.props.dispatch(addWorkspaceContentList([data.content], data.workspace.workspace_id))
+    if (data.content.content_namespace === 'upload') {
+      this.props.dispatch(addWorkspaceShareFolderContentList(
+        [{
+          ...data.content,
+          parent_id: data.content.content_type === CONTENT_TYPE.FOLDER ? SHARE_FOLDER_ID : data.content.parent_id
```
Skylsmoi, line 90: I'm confused here
shouldn't the folder have the proper parent_id already ?
I think the constant SHARE_FOLDER_ID's value should be the same as the backend one (which is -1 I think).
Is this test here only for safety ?

frontend/src/reducer/currentWorkspace.js
```diff
@@ -169,36 +168,6 @@ export default function currentWorkspace (state = defaultWorkspace, action) {
         recentActivityList: state.recentActivityList.filter(c => !action.workspaceContentList.some(cc => c.id === cc.content_id))
       }
 
-    case `${ADD}/${WORKSPACE_CONTENT_SHARE_FOLDER}`:
```
Skylsmoi: why removing all this ?
shouldn't the sharefolder contents appears in the recent activity list of the dashboard ?

frontend/src/reducer/workspaceContentList.js
```diff
@@ -29,53 +30,84 @@ export const serializeContent = c => ({
   created: c.created
 })
 
-export default function workspaceContentList (state = [], action) {
+export default function workspaceContentList (state = {}, action) {
```
Skylsmoi, line 33: add a defaultState constant to initialize the state with the properties workpaeId and contentList

frontend/src/reducer/workspaceShareFolderContentList.js
```diff
@@ -4,58 +4,93 @@ import {
...
+  WORKSPACE_CONTENT_SHARE_FOLDER, ADD,
+  WORKSPACE_CONTENT_SHARE_FOLDER_ARCHIVED,
+  WORKSPACE_CONTENT_SHARE_FOLDER_DELETED,
+  REMOVE,
+  RESTORE
 } from '../action-creator.sync.js'
 import { serializeContent } from './workspaceContentList.js'
 
-export default function workspaceShareFolderContentList (state = [], action) {
+export default function workspaceShareFolderContentList (state = {}, action) {
```
Skylsmoi, line 15: add a defaultState as in workspaceContentList reducer

frontend/src/container/ReduxTlmDispatcher.jsx
```diff
@@ -74,7 +83,17 @@ export class ReduxTlmDispatcher extends React.Component {
   }
 
   handleContentCreated = data => {
-    this.props.dispatch(addWorkspaceContentList([data.content], data.workspace.workspace_id))
+    if (data.content.content_namespace === 'upload') {
```
Skylsmoi, line 86: I think this should be in the reducer.
for workspaceContentList: `id (action.content.content_namespace !== 'content') return`
for workspaceShareFolderContentList: `id (action.content.content_namespace !== 'upload') return`
and for currentWorkspace, I'm sure there is something to handle

Do you think it can be problematic ?

Also, we avoid using hard coded strings like 'upload'. You must add a constant somewhere (util/helper can be a good choice, maybe even frontend_lib) with the values

metzgegu: For now, the actions type for content type (content and shareFolderContent) are separated, so the idea is to use only one action type for both type of content ?
Only `[(ADD)(SET)(UPDATE)(DELETE)(RESTORE)]/WORKSPACE_CONTENT` for example

frontend/src/container/ReduxTlmDispatcher.jsx
metzgegu: If the folder is a sharedFolder the backend set the `parent_id` to `null` (API and TLM), I just followed this current implementation, [see](https://github.com/tracim/tracim/blob/6483195859bdbf76bee7092a937a186b592159a5/frontend/src/container/WorkspaceContent.jsx#L603)

frontend/src/reducer/currentWorkspace.js
metzgegu: Yes, it shouldn't have been removed.

frontend/src/container/ReduxTlmDispatcher.jsx
Skylsmoi: indeed.
It's correct then
you can resolve

frontend/src/reducer/currentWorkspace.js
```diff
@@ -168,6 +169,37 @@ export default function currentWorkspace (state = defaultWorkspace, action) {
         recentActivityList: state.recentActivityList.filter(c => !action.workspaceContentList.some(cc => c.id === cc.content_id))
       }
 
+    case `${RESTORE}/${WORKSPACE_CONTENT_SHARE_FOLDER}`:
+    case `${ADD}/${WORKSPACE_CONTENT_SHARE_FOLDER}`:
+      return {
+        ...state,
+        recentActivityList: [
+          ...action.workspaceShareFolderContentList.map(c => serializeContent(c)),
```
Skylsmoi, line 177: the serialize process has been updated, you need to use the new process

see:
- the new function: https://github.com/tracim/tracim/blob/develop/frontend_lib/src/helper.js#L375
- usage example: https://github.com/tracim/tracim/blob/develop/frontend/src/reducer/user.js#L50

See with gbonaspetti for questions

frontend/src/container/ReduxTlmDispatcher.jsx
Skylsmoi: That's the idea yes.
I have looked closely into your changes and I don't see any downside for having one action action "contentModfied" and having the reducers doing the intelligence I wrote in my previous comment

Skylsmoi: Because, currently, we have 2 ways of separating them:
- the content.content_namespace
- the action name

While only one is required

gbonaspetti:
some problems found during the quality tests:
- the read status does not update to the first clicked content after a change (but as soon as other content is read, the list updates itself)
read
- the content is marked as "not read" even if the change was made by the user who is logged in.

frontend/src/container/WorkspaceContent.jsx
```diff
@@ -113,6 +125,18 @@ export class WorkspaceContent extends React.Component {
     this.updateUrlTitleBreadcrumbs()
   }
 
+  handleContentModified = data => {
+    if (data.author.user_id === this.props.user.userId) {
+      this.props.dispatch(addWorkspaceReadStatus(data.content, data.workspace.workspace_id))
```
Skylsmoi, line 130: couldn't this be in the ReduxTlmDispatcher ?
Is there a reason why you put it in WorkspaceContent component ?

metzgegu: Yes it should be in `ReduxTlmDispatcher`

frontend/src/container/ReduxTlmDispatcher.jsx
```diff
@@ -91,10 +92,16 @@ export class ReduxTlmDispatcher extends React.Component {
     if (response.status !== 200) return
 
     this.props.dispatch(removeWorkspaceReadStatus(response.json, data.workspace.workspace_id))
+    if (data.author.user_id === this.props.user.userId) {
+      this.props.dispatch(addWorkspaceReadStatus(data.content, data.workspace.workspace_id))
+    }
```
Skylsmoi, line 97: this part doesn't make a lot of sense
it literally says:
1) remove read status
2) if you are the author
3) add read status

Also, removeWorkspaceReadStatus() use the data from `response.json` and addWorkspaceReadStatus() from the TLM `data.content`

you are in a comment creation, the content is the comment, not the parent's content

And finally, from what I said yesterday, you are not supposed to add intelligence in here, if you really have to, tell me why please

(by "intelligence", I'm only talking about the if)

frontend/src/container/ReduxTlmDispatcher.jsx
```diff
@@ -91,10 +92,16 @@ export class ReduxTlmDispatcher extends React.Component {
...
   }
 
   handleContentModified = data => {
     this.props.dispatch(updateWorkspaceContentList([data.content], data.workspace.workspace_id))
+    if (data.author.user_id === this.props.user.userId) {
```
Skylsmoi, line 102: same comment about intelligence

frontend/src/container/ReduxTlmDispatcher.jsx
metzgegu: Yes sorry, I updated the pull request according to preview comment too quickly

I put the 'intelligence' here because we must have the current user ID in order to know if the author is the current user connected to read the new revision. I put there because 2 components use `contentReadStatusList` but we could put additional TLM handlers in these 2 components

Skylsmoi: ok, the reason to put the intelligence here because you need the connected user seems fine to me

Issue #3062, opened 2020-05-15
Feat: Ensure an event is commited before publishing it to mail/tlm dispatchers


grignards:
## Feature description

An event should be published only when the transaction creating it has been committed to the database.

## Context

As done currently, an event is published in the context of the tracim's webapp SQL transaction.
This means that if:
  - the transaction is rolled back
  - takes a long time to commit
we can have a job trying to dispatch an event that
  - will never really exist
  - will be written in the database after the job tries to process it

This is not a problem currently as the dispatching is handled directly within tracim's webapp. But the plan is to handle the dispatch in a RQ job.

## Prerequisites

None

## Impacted components

- *backend*

## Extra information

Can be done using [SQL Alchemy's session events](https://docs.sqlalchemy.org/en/13/orm/events.html#session-events), especially `after_commit` and possibly `after_soft_rollback`.

Issue #3064, opened 2020-05-18
Bug: functional tests failing with SQLite


grignards:
## Bug description & symptom

Some frontend functional tests are failing on develop

See here for details https://travis-ci.org/github/tracim/tracim/jobs/687078785

## Diagnostic

Related to a SQLite connection used in another thread:

```
message": "(sqlite3.ProgrammingError) SQLite objects created in a thread can only be used in that same thread. The object was created in thread id 139992464492288 and this is thread id 139992481277696.\n[SQL: SELECT users.password AS users_password, users.user_id AS users_user_id, users.email AS users_email, users.display_name AS users_display_name, users.created AS users_created, users.is_active AS users_is_active, users.is_deleted AS users_is_deleted, users.imported_from AS users_imported_from, users.timezone AS users_timezone, users.auth_type AS users_auth_type, users.lang AS users_lang, users.auth_token AS users_auth_token, users.auth_token_created AS users_auth_token_created, users.reset_password_token_hash AS users_reset_password_token_hash, users.reset_password_token_created AS users_reset_password_token_created, users.allowed_space AS users_allowed_space, users.profile AS users_profile \nFROM users \nWHERE users.user_id = ?]\n[parameters: [{'%(139992500389536 param)s': 1}]]\n(Background on this error at: http://sqlalche.me/e/f405)
```

## Expected

Functional tests must be green.

## Impacted components

- *backend*

inkhey:
Can't reproduce it, either in travis or in local runned test.
Just found an unrelated (without currently any consequences) issue in travis config doing this: https://github.com/tracim/tracim/commit/6ce80ebc749f5afd759011bf29343b9f6f67785e

Skylsmoi:
I could reproduce this issue when running cypress locally.

The test that was repeatedly reproducing it (not 100% of the time though) is: `app_file/add_file_as_user_spec.js` [here](https://github.com/tracim/tracim/blob/develop/functionnal_tests/cypress/integration/app_file/add_file_as_user_spec.js)

adding `email.processing_mode = async` to cypress_test.ini resolved the issue

inkhey:
to reproduce the issue waitress threads should not be set to 1 :
```
[server:main]
use = egg:waitress#main
listen = %(basic_setup.listen)s
threads = 1  #this line need to be removed to use default 4 threads
```

PhilippeAccorsi:
> adding `email.processing_mode = async` to cypress_test.ini resolved the issue

[...]

PR #3141, opened 2020-06-05
Feature/3062  publish event after commit

grignards:
Correct possible incoherence between database state and published events when the events are published asynchronously.

Refs #3062.

## Checkpoints

**For code reviewers**

**For developers**

**For the quality team**


backend/tracim_backend/lib/core/event.py
```diff
@@ -98,19 +99,43 @@ class EventBuilder:
 
     def __init__(self, config: CFG) -> None:
         self._config = config
-        self._current_user = None  # type: typing.Optional[User]
+        # NOTE S.G. 2020-06-04: need thread local storage as EventBuilder()
+        # is created only once/wsgi application
+        self._thread_local = threading.local()
```
inkhey, line 104: I'm not sure to understand why we do need thread local code, was it no more simple to ensure we hare not sharing object between thread ?  It maybe not possible but it's unclear to me yet.

Anyway, do this may solve issue #3064 ?

backend/tracim_backend/lib/core/event.py
```diff
@@ -98,19 +99,43 @@ class EventBuilder:
...
     def on_current_user_set(self, user: User) -> None:
-        self._current_user = user
+        self._thread_local.current_user = user
 
     @hookimpl
     def on_request_session_created(self, request: TracimRequest, session: TracimSession) -> None:
-        event.listen(session, "pending_to_persistent", self._publish_event)
+        commit_event = "after_flush"
+        if self._config.JOBS__PROCESSING_MODE == self._config.CST.ASYNC:
+            # We need after commit event when processing in async
```
inkhey, line 123: Can't we do both sync and async with "after commit" ?

backend/tracim_backend/tests/functional/test_live_messages.py
```diff
@@ -123,3 +123,39 @@ def test_api__user_live_messages_endpoint_with_GRIP_proxy__ok__user_update__asyn
...
         assert result["event_type"] == "user.modified"
         assert event1.event == "message"
+
+    @pytest.mark.parametrize(
+        "config_section", [{"name": "functional_async_live_test"}], indirect=True
+    )
+    def test_api__user_live_messages_endpoint_with_GRIP_proxy__ok__content_creation__async(
+        self, pushpin, app_config, rq_database_worker
+    ):
+        headers = {"Accept": "text/event-stream"}
```
inkhey, line 133: Can you just put a docstring here to clarify what is tested here.

backend/tracim_backend/lib/core/event.py
grignards: Unfortunately not as the session is un-useable `after_commit` (I tried).

backend/tracim_backend/tests/functional/test_live_messages.py
grignards: I think I will remove this test: I put it here since I saw more "missed" events when creating contents during my firts manual tests, but after more testing this is not especially contents which are concerned (as expected).

backend/tracim_backend/lib/core/event.py
grignards: There is no thread local part anymore: pending events are now stored on the given TracimContext instance.

backend/tracim_backend/lib/core/event.py
```diff
@@ -60,6 +60,8 @@
 
 RQ_QUEUE_NAME = "event"
 
+_PENDING_EVENTS = "pending_events"
```
inkhey, line 63: This seems not used or do i miss something ?

backend/tracim_backend/lib/core/event.py
```diff
@@ -299,19 +315,34 @@ def _create_role_event(
...
-                context=session.context
+                context=context
             )  # type: BaseLiveMessageBuilder
         else:
-            message_builder = SyncLiveMessageBuilder(context=session.context)
-        message_builder.publish_messages_for_event(event.event_id)
+            message_builder = SyncLiveMessageBuilder(context=context)
+
+        # We only publish events that have an event_id from the DB.
+        # we can have `new` events here as we add events in the session
```
inkhey, line 333: Not sure to understand why we can have unpublishable event, do you have an exemple ?

backend/tracim_backend/lib/core/event.py
grignards: Ah yes, I forgot this.

backend/tracim_backend/lib/core/event.py
grignards: The event creation process is done like this (example for user):

1. user_api.update(user)
2. (during session `after_flush`  number 1):
   - on_user_modified
   - create_user_event, add it to session
   - `_publish_events` is also called as it is a `post_flush` operation when processing mode is `sync`. At this point the `user.modified` event does not have its id set (it has not been flushed yet, only added to the session)
3. (during session `after_flush` number 2, eg. just before transaction.commit):
  - `_publish_events` is called again, `user.modified` pending event has its id set so can be published.

What's a bit unfortunate is that we cannot do any SQL with the session's in `after_commit`: at this point we are sure that all pending events have an id, but we cannot use the session to do SQL anymore.

inkhey: ok.

Issue #3129, opened 2020-06-03
Feat: Have a request identification given by the associated TLM (frontend part)


Skylsmoi:
## Feature description

When frontend POST or PUT something, it will add an id is the request (Body ? Header ?).
And, when the backend will generate the TLM associated with this modification, it will give back it.

## Todo list
- [x] specify this id (header, body, name, type ...)
- [x] backend implementation
- [x] frontend implementation

## Context

This will allow to handle this specific case https://github.com/tracim/tracim/issues/3066#issuecomment-637611391

## Impacted components

backend
frontend
apps

## (future) Implemented solution:
https://github.com/tracim/tracim/pull/3144
- Frontend can send `X-Tracim-ClientToken` Header for any request to backend.
- This header will set "client_token" field of TLM created from these request with the header value.

grignards:
Like we discussed the id should be unique for a (user, browser, tab) tuple. making it unique/request is simpler but has the default of being unpredictable for frontend components that have not originated the request.
I don't know if the case to have an special case in a separate component is realistic though…

grignards:
In backend this shouldn't be a big work as we already have access to the current (== authentified) user when creating the event/message. So we "just" have to add this id to the current user object.

lebouquetin:
what about a uuid token generated at JS load time? IE in the first line of Tracim javascript code executed? Or when the store is setup?

Maybe something like `<ip_md5>-<userid>-<browser-string-md5>-<uuid>` would allow to easier investigate in case of bug? is it overkill?

grignards:
> Maybe something like `<ip_md5>-<userid>-<browser-string-md5>-<uuid>` would allow to easier investigate in case of bug? is it overkill?
> 
`user_id` is already present in the TLMs (in author field). Including the ip address hash could have subtle effects if the same client's computer/phone changes IP while using Tracim (provided the id is updated in these cases).

Storing the id in the frontend's store will lead to losing it if the tab is reloaded but this doesn't seem a problem for our need.

grignards:
>  specify this id (header, body, name, type ...)

I would be in favor of a request header for transmitting the id as it is orthogonal to what it's done in the request (kinda like the authentication cookie).
Also this will be much simpler to implement than transmitting it through the request's body/query parameters.

I would use a string type which will be "opaque" for the back-end: we will only copy it to the messages without interpreting it or enforcing any kind of validation.

As for its name… `client_id` in the author field of TLMs ?
[...]

PR #3144, opened 2020-06-05
add client token support for tracim web request

inkhey:
related to #3129 

## Checkpoints

**For code reviewers**

**For developers**

**For the quality team**

### DIFF
--- a/backend/tracim_backend/lib/utils/request.py
+++ b/backend/tracim_backend/lib/utils/request.py
@@ -49,14 +49,14 @@ class TracimContext(ABC):
         # Candidate content_type found in request body
         self._candidate_content_type = None
         # Client Id, useful to permit link between request and TLM response
-        self._client_id = None  # type: typing.Optional[str]
+        self._client_token = None  # type: typing.Optional[str]
 
     # INFO - G.M - 2018-12-03 - Useful property of Tracim Context
     def set_user(self, user: User):
         self._current_user = user
 
-    def set_client_id(self, client_id: str):
-        self._client_id = client_id
+    def set_client_token(self, client_token: str):
+        self._client_token = client_token
 
     @property
     @abstractmethod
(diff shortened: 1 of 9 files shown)


backend/tracim_backend/lib/utils/request.py
```diff
@@ -255,6 +260,11 @@ def set_user(self, user: User):
         super().set_user(user)
         self.plugin_manager.hook.on_current_user_set(user=user, request=self)
 
+    def set_client_id(self, client_id: str):
+        """Overload TracimContext method to add plugin hook call."""
+        super().set_client_id(client_id)
+        self.plugin_manager.hook.on_current_client_id_set(client_id=client_id, request=self)
+
```
grignards, line 267: I think we can define this method only on `TracimRequest` as it has a meaning only for HTTP requests coming from Tracim's frontend.

inkhey:
@Skylsmoi Can you test this branch, set "X-Tracim-ClientId" Header and check TLM received ?

lebouquetin:
I suggest:

- `_id` suffix -> should be used for int values
- `_token` or`_slug` suffix for str values.

It make things more clear.

inkhey:
X-Tracim-ClientTo

> I suggest:
> 
>     * `_id` suffix -> should be used for int values
> 
>     * `_token` or`_slug` suffix for str values.
> 
> 
> It make things more clear.
> 
> edit: in the current case, this is not a slug (which implies a "static" value), so i suggest `_token`

So :
```
Header: X-Tracim-ClientToken
tlm field name: client_token
```

lebouquetin:
@Skylsmoi yes, this is my suggestion. @inkhey @grignards do you agree with this?

inkhey:
> @Skylsmoi yes, this is my suggestion. @inkhey @grignards do you agree with this?

I found "token" a bit less clear about the usage than id (a token is an id, but when i do think token, i mostly think about authentication), but that's not a big deal.

The reason i do use str here, is that i do think it's easier for frontend to create unique id here, maybe should we use uuid ? @Skylsmoi Did you suggest something specific for identificator here  ?

lebouquetin:
> a token is an id, but when i do think token, i mostly think about authentication

I disagree that token is related to authentication; it is related to session for sure.

Using a string is required. You do not need to know in backend how it is used, actually. This is a feature usefull for clients but it has nothing to do with backend (it only allows to link API request with associated TLM)

grignards:
> @Skylsmoi yes, this is my suggestion. @inkhey @grignards do you agree with this?

`Token` is ok for me.

Skylsmoi:
I don't think I suggested anything

Since it's a header, it will be a string though. I don't plan to cast it to int, string is fine

Skylsmoi:
@inkhey This is a full backend PR, what should I do exactly ?

inkhey:
quality test here does no make sense : i checked.

Issue #3145, opened 2020-06-05
Bug: Events/messages aren't generated with webdav requests


grignards:
## Bug description & symptom

The webdav changes do not create events/messages.

## Diagnostic

Missing initialization of context in webdav application.

## Expected

Webdav should generate events/messages as in the web application.

## Context and version information

On develop.

## Screenshots

*Optional - add screenshots to help explain your problem.*

*Please consider sharing an [animated GIF screencast generated with peek](https://github.com/phw/peek)*

## Impacted components

- *backend*
- *webdav*

## Extra information

*Optional - add any useful information to understand and develop the feature*

lebouquetin:
@grignards I'm afraid by this issue. The event mechanism must be in tracim core features. This issue let me think that it is not implemented like this...

PR #3149, opened 2020-06-08
Bugfix/3145  webdav tlms

grignards:
Refs #3145.

## Checkpoints

**For code reviewers**

**For developers**

**For the quality team**


backend/tracim_backend/models/tracim_session.py
```diff
@@ -12,13 +16,34 @@ class TracimSession(Session):
     def __init__(self, *args, **kwargs) -> None:
         super().__init__(*args, **kwargs)
         self._allow_revision_deletion = False  # type: bool
+        self._context = None  # type: typing.Optional[weakref.CallableProxyType]
+
+    # TODO S.G 2020-06-08: this is a temporary setup until #1834 is done
```
inkhey, line 21: That's effectively not really nice, reason to do this is for getting the proper plugin_manager ?

grignards: Yes, to be able to do the verification in XxxApi classes.

grignards: Since #1854 will likely lead to give `TracimContext` (or something similar) to XxxApi instead of session/current user/config/..., I think it is acceptable.

backend/tracim_backend/lib/rq/worker.py
```diff
@@ -69,7 +69,7 @@ def worker_context() -> typing.Generator[TracimContext, None, None]:
         yield context
         transaction.commit()
     except Exception:
-        transaction.rollback()
+        transaction.abort()
```
inkhey, line 72: What the difference between abort and rollback ?

grignards: `transaction` uses `abort()` as I found out (`rollback()` doesn't exist).

PR #3150, opened 2020-06-08
Feature/2850  handle tlm in account

gbonaspetti:
#2850 

## Checkpoints

**For code reviewers**

**For developers**

**For the quality team**


frontend/test/src/container/Account.spec.js
```diff
@@ -31,7 +31,7 @@ import {
   mockPutMyselfName200,
   mockPutMyselfPassword204, mockPutMyselfPassword403
 } from '../../apiMock'
-
+// TODO update tests
```
Skylsmoi, line 34: if you plan on updating the tests in that PR, no problem.
If you plan on doing it later with another issue, don't forget the link to the issue

frontend/src/reducer/user.js
```diff
@@ -33,25 +48,27 @@ export default function user (state = defaultUser, action) {
...
 
-    case `${UPDATE}/${USER_PUBLIC_NAME}`:
-      return { ...state, public_name: action.newName }
+    case `${UPDATE}/${USER}`:
+      return {
+        ...state,
+        ...serializeUser(action.newUser),
+        agendaUrl: serializeUser(action.newUser).agendaUrl ? serializeUser(action.newUser).agendaUrl : state.agendaUrl,
+        logged: serializeUser(action.newUser).logged ? serializeUser(action.newUser).logged : state.logged
+      }
```
Skylsmoi, line 67: avoid multiple calls to the same function and try using `||` for ternaries that returns the condition
```suggestion
    case `${UPDATE}/${USER}`:
      const serializedUser = serializeUser(action.newUser)
      return {
        ...state,
        ...serializedUser,
        agendaUrl: serializedUser.agendaUrl || state.agendaUrl,
        logged: serializedUser.logged || state.logged
      }
```

frontend/src/reducer/user.js
```diff
@@ -33,25 +48,27 @@ export default function user (state = defaultUser, action) {
     case `${SET}/${USER_CONNECTED}`:
       return {
         ...state,
-        ...action.user,
-        lang: action.user.lang ? action.user.lang : state.lang
+        ...serializeUser(action.user),
+        lang: serializeUser(action.user).lang ? serializeUser(action.user).lang : state.lang
```
Skylsmoi, line 52: same comment as `case ${UPDATE}/${USER}:`

frontend/test/src/container/Account.spec.js
gbonaspetti: in this PR, just a self reminder :smile:

frontend/src/reducer/user.js
gbonaspetti: Changed

frontend/src/reducer/user.js
gbonaspetti: Changed

frontend/src/reducer/user.js
```diff
@@ -46,11 +62,7 @@ export const defaultUser = {
 export default function user (state = defaultUser, action) {
   switch (action.type) {
     case `${SET}/${USER_CONNECTED}`:
-      return {
-        ...state,
-        ...serializeUser(action.user),
-        lang: serializeUser(action.user).lang || state.lang
-      }
+      return serializeUser({ ...unserializeUser(state), ...action.user })
```
raphj, line 65: Would something like this work?

```suggestion
      return {...state, ...action.user }
```

gbonaspetti: No, because `action.user` does not have all the properties of `state` and we want overwrite only the properties that it has, but when it goes through the `serializeUser` these properties are set as `undefined` and overwrite all the properties of `state`.

raphj: Ah right. This is a bit frustrating to unserialize and then serialize almost the same thing and this looks like a code that could be rewritten without thinking about what you said. Could you add a comment to warn about this?

gbonaspetti: Done

Skylsmoi, line 67: can't you write it like this:
``` jsx
return { ...state, ...serializeUser(action.user) }
```

Btw, for the SET, I think we should not keep the old state, we should discard it because it is used, for example, when logging off and logging in with a new user.

It's different from UPDATE user where we only change some properties of the same user and where we must keep the old state

frontend/src/reducer/user.js
```diff
@@ -59,12 +71,7 @@ export default function user (state = defaultUser, action) {
       return { ...state, lang: action.lang }
 
     case `${UPDATE}/${USER}`:
-      return {
-        ...state,
-        ...serializeUser(action.newUser),
-        agendaUrl: serializeUser(action.newUser).agendaUrl || state.agendaUrl,
-        logged: serializeUser(action.newUser).logged || state.logged
-      }
+      return serializeUser({ ...unserializeUser(state), ...action.newUser })
```
Skylsmoi, line 76: same question as above, could this be enough ?
``` jsx
return { ...state, ...serializeUser(action.newUser) }
```

frontend/src/reducer/user.js
Skylsmoi: it would end up like this 
``` jsx
return { ...serializeUser(action.user) }
```

raphj: See https://github.com/tracim/tracim/pull/3150#discussion_r437982602 @Skylsmoi

frontend_lib/test/helper.spec.js
```diff
@@ -150,15 +151,14 @@ describe('helper.js', () => {
...
     }
     const objectToSerialize = {
-      avatar_url: undefined,
       email: null,
       user_id: 0,
-      public_name: 'a_name',
+      public_name: '',
       lang: 'pt',
-      username: ''
+      username: undefined
```
raphj, line 158: I understand that you want to test that the username key will be present after serialization (even if `undefined`).
Then the test bellow is not sufficient. You need to test `Object.prototype.hasOwnProperty.call(serializedObj, 'username')` or something like that.

If you don't care whether the key is actually present, feel free to ignore this comment and resolve the conversation :-)

Issue #3107, opened 2020-05-29
Feat: Handle TLM for app Gallery


Skylsmoi:
## Feature description

Refactor the app gallery with TLM usage

## Context

The v3.0 is the live update version. Apps must trigger update through this process

## Impacted components

App gallery

## Design

(discussed with @celiamgtt):

- when a currently displayed image is deleted, the previous image is automatically changed. If this was the first picture, we move on to the "new first"
- when an unprinted image is deleted, the thumbnail is silently removed
- when an unprinted image is modified, the thumbnail is silently updated
- when a displayed image is modified, the display is updated
- if an image is added (or undeleted), the thumbnail is silently added

@celiamgtt says that people are used to content that automatically update, provided it is not disruptive (e.g. it does not loose work)

## Extra info

I'm not sure there is a lot to do here

raphj:
This comment explains the implementation design for this refactor.

With the live messages, we are going to update the list of pictures as well as the currently displayed picture according to the live messages we receive.
 A first approach was to just reload the list of pictures from the server to ensure that we are always consistent with the list the backend sends. However, this did not end well.
 
 If several pictures are uploaded / modified in a short time, the list of pictures was downloaded several times before the state of the component had time to change even once, leading to crashes and weird behavior. Even if it worked, this would be an easy but inefficient way of handling
 updates.
 
 Imagine a folder with 1000 pictures. Are we handling this list again and again if 10 pictures are added? Come on, what was I thinking.
 
 So we are applying the harder but more correct approach: updating the list using the TLM.
 
 We still want to ensure consistency with what would be displayed if we downloaded the list of pictures from the backend though.
 
 This list built there at the time of writing this comment:
 
    backend/tracim_backend/views/core_api/workspace_controller.py
 
 In the workspace_content method of the WorkspaceController class we can see that the contents of a folder is **sorted by label**.
 
 That's what we are going to do.
 
 This is bad news though: if we are currently watching a picture that is renamed, it will move in the list.
 
 What is the correct behavior?
1. Not changing the current picture index if a picture is renamed.  This will not change the position, but this will make the currently picture change "out of nowhere"
2. Changing the position does not has this downside, but the user might lose track of its picture browsing session.
 
Since anything could happen with the list of pictures anyway (imagine someone taking the time to sort their Tracim folder and renaming random pictures for instance). We cannot really avoid the downsides of 2) anyway. We will at least free the user from liberal image jumping.
 
[...]

PR #3146, opened 2020-06-05
add TLM implementation for app Thread + unit tests

Skylsmoi:
https://github.com/tracim/tracim/issues/3109

## Checkpoints

The goal of this section is to help code reviewers and the merge request author to do the required checks on this pull request. Please don't edit this list.
If one or more of these checkpoints are out of scope, please write below why they are.

**For code reviewers**

- [x] The code is clear enough
- [x] If there are FIXMEs in the code, then there are associated issues and issue is referenced in the FIXME
- [x] If there are TODOs, NOTEs or HACKs in code, the date and the developer initials are present
- [x] Automated tests have been written (covering the feature or non regression), or *at least* an issue has been created to implement the test

**For developers**

- [x] Manual tests done to ensure stability on whole application layers, issue expectation follow
- [x] The original issue has been updated with a short summary of the implemented solution

**For the quality team**

- [x] Tested by the quality team

PR #3153, opened 2020-06-08
Feature/3107 handle TLM for app Gallery

raphj:
#3107 

## Checkpoints

**For code reviewers**

**For developers**

**For the quality team**


frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -244,14 +302,31 @@ export class Gallery extends React.Component {
...
+          }
+        }
+
+        this.setState({ imagesPreviews, displayedPictureIndex, imagesPreviewsLoaded: true })
 
         break
       }
-      default: this.sendGlobalFlashMessage(props.t('Error while loading content list'))
+      default:
+        console.log("fetchContentList", fetchContentList)
```
gbonaspetti, line 328: missing console.log?

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -502,10 +577,19 @@ export class Gallery extends React.Component {
     if (state.autoPlay) this.displayReactImageLightBoxArrows(false)
   }
 
+  // note: returns undefined if there is no current picture
```
gbonaspetti, line 580:
```suggestion
  // INFO - 2020-06-09 - RJ this function returns undefined if there is no current picture
```

frontend_app_gallery/test/container/Gallery.spec.js
```diff
@@ -0,0 +1,94 @@
...
+      previewUrlForThumbnail: [picture.filename],
+      rotationAngle: 0,
+      rawFileUrl: [picture.filename]
+    }))
+  }
+
+  const wrapper = shallow(<Gallery {...props} />)
+
+  describe('TLM handlers', () => {
+    describe('handleContentDeleted', () => {
```
gbonaspetti, line 65: Why do you only have tests for `handleContentDeleted` and not `handleContentCreatedOrModifiedOrUndeleted`?

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -67,41 +72,94 @@ export class Gallery extends React.Component {
...
+      { entityType: TLM_ET.CONTENT, coreEntityType: TLM_CET.DELETED, optionalSubType: TLM_ST.FILE, handler: this.handleContentDeleted },
+      { entityType: TLM_ET.CONTENT, coreEntityType: TLM_CET.UNDELETED, optionalSubType: TLM_ST.FILE, handler: this.handleContentCreatedOrModifiedOrUndeleted }
+    ])
+
     this.lightboxRotation = new LightboxRotation()
+  }
 
-    document.addEventListener(CUSTOM_EVENT.APP_CUSTOM_EVENT_LISTENER, this.customEventReducer)
+  // TLM Handlers
+  handleContentCreatedOrModifiedOrUndeleted = data => {
```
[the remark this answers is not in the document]
raphj: A picture can be created with another app. Does that answer your question?

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -67,41 +72,94 @@ export class Gallery extends React.Component {
...
+    // A picture has been added or modified - let's reload the gallery list.
```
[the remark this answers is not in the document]
raphj: if that not handled in this line:

```js
    props.registerLiveMessageHandlerList([
      { entityType: TLM_ET.CONTENT, coreEntityType: TLM_CET.MODIFIED, optionalSubType: TLM_ST.FILE, handler: this.handleContentCreatedOrModifiedOrUndeleted }, 
```

with  `optionalSubType`?

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -124,7 +182,7 @@ export class Gallery extends React.Component {
       this.loadGalleryList(state.config.appConfig.workspaceId, state.folderId)
       const contentDetail = await this.loadContentDetails()
       this.buildBreadcrumbs(contentDetail.workspaceLabel, contentDetail.folderDetail, false)
-    } else if (prevState.fileSelected !== state.fileSelected || prevState.imagesPreviewsLoaded === !state.imagesPreviewsLoaded || prevState.breadcrumbsLoaded === !state.breadcrumbsLoaded) {
+    } else if (prevState.displayedPictureIndex !== state.displayedPictureIndex || prevState.imagesPreviewsLoaded === !state.imagesPreviewsLoaded || prevState.breadcrumbsLoaded === !state.breadcrumbsLoaded) {
```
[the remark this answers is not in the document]
raphj: Sure.

```suggestion
    } else if (
      prevState.displayedPictureIndex !== state.displayedPictureIndex ||
      prevState.imagesPreviewsLoaded === !state.imagesPreviewsLoaded ||
      prevState.breadcrumbsLoaded === !state.breadcrumbsLoaded
    ) {
```

frontend_app_gallery/test/container/Gallery.spec.js
raphj: Because handleContentCreatedOrModifiedOrUndeleted reuses the logic already in Gallery. We should discuss this.

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -67,41 +72,94 @@ export class Gallery extends React.Component {
...
-          this.setState({
-            config: data.config,
-            folderId: newFolderId
-          })
+    let newDisplayedPictureIndex = state.displayedPictureIndex
+    let newImagesPreviews = state.imagesPreviews
+
+    // We remove it from the list of previews
+    for (let i = 0; i < state.imagesPreviews.length; i++) {
+      if (state.imagesPreviews[i].contentId === data.content.content_id) {
```
Skylsmoi, line 106: to get the index of an element in an array of object, you can use the `findIndex` method
see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex

frontend_app_gallery/src/container/Gallery.jsx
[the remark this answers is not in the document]
gbonaspetti: I'll give you the example of app_html-doc because I think it will be easier to understand. In the app you can create content such as html-document, but you can also create content of the type comment. The optionalSubType is used to differentiate what type of content is being created/modified/deleted/restored, because for specific type of content we may have specific fields that are useless for others (for example, the file type needs the "number of pages" field that is only useful for this type of content).

What I was referring to in the comment is that if you receive a request to delete a file type content, you should check the id of that content before deleting to make sure you are deleting the correct one. If I'm viewing a content X and another user has deleted another content Y, I will receive TLM for that deletion, but it shouldn't influence the content I'm viewing at the moment.

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -656,7 +740,7 @@ export class Gallery extends React.Component {
                 onClose={this.handleCloseDeleteFilePopup}
               >
                 <div className='gallery__delete__file__popup__body'>
-                  <div className='gallery__delete__file__popup__body__msg'>{props.t('Are you sure ?')}</div>
+                  <div className='gallery__delete__file__popup__body__msg'>{props.t('Are you sure?')}</div>
```
Skylsmoi, line 747: if you change a translation key, don't forget to run the `build-transaltion` script

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -244,14 +306,31 @@ export class Gallery extends React.Component {
...
+
+        let displayedPictureIndex = state.displayedPictureIndex
+        const currentPictureId = this.currentPictureId()
+        if (currentPictureId) {
+          for (let i = 0; i < imagesPreviews.length; i++) {
+            if (imagesPreviews[i].contentId === currentPictureId) {
+              displayedPictureIndex = i
+              break
+            }
+          }
```
Skylsmoi, line 324: using `findIndex` this would be
`displayedPictureIndex = imagesPreviews.findIndex(image => image.contentId === currentPictureId)`

frontend_app_gallery/src/container/Gallery.jsx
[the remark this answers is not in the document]
raphj: I think I will not receive events for comments because of `entityType: TLM_ET.CONTENT`.

For the gallery, we need to update the thumbnails for any image shown, so I cannot just handle the currently displayed picture.

Is this okay or is there something I am missing there?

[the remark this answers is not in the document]
gbonaspetti: :thinking: I think it's okay then, I hadn't noticed that detail

frontend_app_gallery/src/container/Gallery.jsx
raphj: The equivalent code using findIndex is:

```suggestion
        let displayedPictureIndex = -1

        const currentPictureId = this.currentPictureId()
        if (currentPictureId) {
          displayedPictureIndex = imagesPreviews.findIndex(image => image.contentId === currentPictureId)
        }

        if (displayedPictureIndex === -1) {
          displayedPictureIndex = state.displayedPictureIndex
        }
```

I prefer the for version, but feel free to commit my suggestion if you really prefer this version.

frontend_app_gallery/src/container/Gallery.jsx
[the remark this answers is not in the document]
gbonaspetti: Not really, it's more a question of naming (i think), because you have a handleContent**Created**OrModifiedOrUndeleted if there is no handle for it, like:

```
{ entityType: TLM_ET.CONTENT, coreEntityType: TLM_CET.CREATED, optionalSubType: TLM_ST.FILE, handler: this.handleContentCreatedOrModifiedOrUndeleted }
```

If it is created by another app, it will be handled by it. Right?

frontend_app_gallery/src/container/Gallery.jsx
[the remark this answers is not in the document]
Skylsmoi: I think you, at least, need to check that the TLM is about the same workspace and the same parent_id as the folder you are opening the gallery on

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -581,11 +665,11 @@ export class Gallery extends React.Component {
             >
               <div ref={modalRoot => (this.reactImageLightBoxModalRoot = modalRoot)} />
 
-              {state.displayLightbox && (
+              {state.displayLightbox && this.currentPicture() && (
```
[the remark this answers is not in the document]
raphj: Yes it can if the last picture is deleted, if I understand the code correctly. Can you confirm?

frontend_app_gallery/test/container/Gallery.spec.js
gbonaspetti: If the logic itself is already tested, you can at least test "if I receive a tlm, the loadGalleryList function is called".

frontend_app_gallery/src/container/Gallery.jsx
raphj: I messed you commiting the json files, I'll commit them. Thanks.

frontend_app_gallery/src/container/Gallery.jsx
Skylsmoi: I'm not sure I understand, your suggestion's addition seems to do more than its deletion

The findIndex is only meant to replace the for, why do you need some more variable initialization using findIndex ?

frontend_app_gallery/src/container/Gallery.jsx
raphj: indeed

frontend_app_gallery/src/container/Gallery.jsx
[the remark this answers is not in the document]
Skylsmoi: you are correct, you can resolve after reading this

frontend_app_gallery/src/container/Gallery.jsx
Skylsmoi: :+1:
I resolve

frontend_app_gallery/src/container/Gallery.jsx
raphj: sorry, my snippet is for replacing lines 316 and 317 too:

```
        let displayedPictureIndex = state.displayedPictureIndex
        const currentPictureId = this.currentPictureId()
```
findIndex is nice but it needs a bit of work here because it can return -1 and I need to add a check for that. The for loop was easier and the whole code with the for loop is more concise. I you really prefer I'd use findIndex here I'll commit an equivalent code using findIndex but the improvement by doing that is not clear to me.

Skylsmoi: What's the difference between "displayed"Picture and "current"Picture ?

Skylsmoi: I do prefer the use of `findIndex` but I'd write it with a little difference:
``` jsx
let displayedPictureIndex = -1
const currentPictureId = this.currentPictureId()

if (currentPictureId) {
  displayedPictureIndex = imagesPreviews.findIndex(image => image.contentId === currentPictureId)
  if (displayedPictureIndex === -1) displayedPictureIndex = state.displayedPictureIndex
}
```
The 2 reasons for me to prefer this are:
- it explicitly tells what we are doing if `currentPictureId` isn't in `imagesPreviews`
- in Tracim, we tend to use array manipulation functions over for loop

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -67,41 +72,94 @@ export class Gallery extends React.Component {
...
+    let newImagesPreviews = state.imagesPreviews
+
+    // We remove it from the list of previews
+    for (let i = 0; i < state.imagesPreviews.length; i++) {
+      if (state.imagesPreviews[i].contentId === data.content.content_id) {
+        // We found the picture to delete
+
+        // We remove it in the new list of images
+        newImagesPreviews = state.imagesPreviews.slice()
+        newImagesPreviews.splice(i, 1)
```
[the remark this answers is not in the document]
raphj: I removed my use of splice. I used it because I wanted to avoid looping twice on the same array by using filter, but I found another solution. splice modified an array that belonged to the previous state, which is not good I guess.

frontend_app_gallery/src/container/Gallery.jsx
raphj: > use most recent javascript features

This is not my goal. My goal is to write efficient and legible code, and for loops often achieve both goals for me, the latter being more than subjective though. I'm fond of recent Javascript feature and will gladly use them when needed.
Functional-style array manipulation functions invite developers to write code that loops over arrays more times than necessary and I don't think this is a good thing too.

They also force the usage of callback functions that get called for every elements of arrays, and calling a function is slow against running the body of a for loop.

Comment taken though, I'll try to avoid for loops and use functional-style programming instead in Tracim, but I'm not sure it is a good solution in every cases.

I agree with you about being explicit though.

In any case, I updated the code to use findIndex :-)

frontend_app_gallery/src/container/Gallery.jsx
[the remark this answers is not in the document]
raphj: I rewrote this PR, I think I fixed this.

frontend_app_gallery/src/container/Gallery.jsx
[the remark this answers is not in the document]
raphj: You are right. I did this.

frontend_app_gallery/test/container/Gallery.spec.js
raphj: Right, I made more extensive tests.

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -67,43 +71,223 @@ export class Gallery extends React.Component {
...
+    this.setState({ imagesPreviews, displayedPictureIndex })
+  }
+
+  handleShowApp = data => {
+    console.log('%c<Gallery> Custom event', 'color: #28a745', data)
+    const newFolderId = qs.parse(data.config.history.location.search).folder_ids
+    if (data.config.appConfig.workspaceId !== this.state.config.appConfig.workspaceId || newFolderId !== this.state.folderId) {
+      this.setState({
+        config: data.config,
+        folderId: newFolderId
```
raphj, line 273: Check that the folderId is initialized somewhere

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -243,51 +431,102 @@ export class Gallery extends React.Component {
...
+      fileName: file.filename,
+      lightBoxUrlList,
+      previewUrlForThumbnail,
+      rotationAngle: 0,
+      rawFileUrl
+    }
+  }
 
-    return (await Promise.all(images.map(async (image) => {
+  loadPreviews = async (imageContentIds) => {
```
raphj, line 514: change to loadPreviewList

frontend_app_gallery/test/container/Gallery.spec.js
```diff
@@ -0,0 +1,147 @@
...
+      loggedUser: {
+        ...defaultDebug.loggedUser
+      }
+    },
+    t: key => key
+  }
+
+  const stateMock = {
+    displayedPictureIndex: 1,
+    imagesPreviews: pictures.map(picture => ({
```
raphj, line 46: Try to use the method from the component if not too complicated

frontend_app_gallery/test/container/Gallery.spec.js
```diff
@@ -0,0 +1,147 @@
...
+    describe('handleContentDeleted', () => {
+      describe('delete a picture', () => {
+        it('should not be in the picture list anymore', () => {
+          wrapper.setState(stateMock)
+          wrapper.instance().handleContentDeleted({content: pictures[1]})
+          expect(wrapper.state().imagesPreviews.every(image => image.contentId !== pictures[1].content_id))
+          expect(wrapper.state().imagesPreviews.length).to.equal(stateMock.imagesPreviews.length - 1)
+        })
+
+        it('should go to the next picture', () => {
```
raphj, line 70: add a describe per case

frontend_app_gallery/test/container/Gallery.spec.js
```diff
@@ -0,0 +1,147 @@
...
+          wrapper.setState(stateMock)
+          wrapper.instance().handleContentDeleted({ content: { ...pictures[0], workspace_id: 2 } })
+          expect(wrapper.state().imagesPreviews.length).to.equal(stateMock.imagesPreviews.length)
+        })
+      })
+    })
+
+    describe('handleContentModified', () => {
+      describe('modify the current picture', () => {
+        it('should not keep the old label in the picture list anymore and but the new one, yes after rename', () => {
```
raphj, line 104:
```suggestion
        it('should not keep the old label in the picture list anymore but the new one, yes after rename', () => {
```

frontend_app_gallery/test/container/Gallery.spec.js
```diff
@@ -0,0 +1,147 @@
...
+          wrapper.setState(stateMock)
+          wrapper.instance().handleContentModified({ content: { ...pictures[1], label: 'betterversion' } })
+          expect(wrapper.state().imagesPreviews.every(image => image.label !== pictures[1].label))
+          expect(wrapper.state().imagesPreviews.some(image => image.label === 'betterversion'))
+        })
+
+        it('should be sorted', () => {
```
raphj, line 111: better description

frontend_app_gallery/test/container/Gallery.spec.js
```diff
@@ -0,0 +1,147 @@
...
+          wrapper.setState(stateMock)
+          wrapper.instance().handleContentModified({ content: { ...pictures[1], label: 'betterversion' } })
+          let sortedImagesPreviews = wrapper.state().imagesPreviews.slice()
```
raphj, line 114:
```suggestion
          let sortedImagesPreviews = [... wrapper.state().imagesPreviews]
```

frontend_app_gallery/test/container/Gallery.spec.js
raphj: I'm updating the descriptions so they are clearer, but I'm keeping the describe / it as is because I'm not so sure that this should changed after all

frontend_app_gallery/src/container/Gallery.jsx
raphj: ok, it is set in the constructor of the app. Let's just remove the irrelevant showApp custom event.

frontend_app_gallery/i18next.scanner/en/translation.json
```diff
@@ -18,6 +18,7 @@
   "Open raw file": "Open raw file",
   "There isn't any previewable content at that shared space's root.": "There isn't any previewable content at that shared space's root.",
   "There isn't any previewable content at that folder's root.": "There isn't any previewable content at that folder's root.",
+  "There isn’t any picture to delete.": "There isn’t any picture to delete.",
```
gbonaspetti, line 21:
```suggestion
  "There isn't any picture to delete.": "There isn't any picture to delete.",
```

frontend_app_gallery/i18next.scanner/fr/translation.json
```diff
@@ -18,6 +18,7 @@
   "Open raw file": "Ouvrir le fichier original",
   "There isn't any previewable content at that shared space's root.": "Il n'y a pas de contenu affichable à la racine de cet espace partagé.",
   "There isn't any previewable content at that folder's root.": "il n'y a pas de contenu affichable à la racine de ce dossier.",
+  "There isn’t any picture to delete.": "Il n’y a pas d’image à supprimer.",
```
gbonaspetti, line 21:
```suggestion
  "There isn't any picture to delete.": "Il n'y a pas d'image à supprimer.",
```

frontend_app_gallery/i18next.scanner/pt/translation.json
```diff
@@ -18,6 +18,7 @@
   "Open raw file": "Abrir a documentação não processada (raw)",
   "There isn't any previewable content at that shared space's root.": "Não existe nenhum conteúdo pré-visualizável na raiz desse espaço partilhado.",
   "There isn't any previewable content at that folder's root.": "Não existe nenhum conteúdo pré-visualizável na raiz dessa pasta.",
+  "There isn’t any picture to delete.": "Não há nenhuma imagem a apagar.",
```
gbonaspetti, line 21:
```suggestion
  "There isn't any picture to delete.": "Não há nenhuma imagem a apagar.",
```

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -53,55 +58,147 @@ export class Gallery extends React.Component {
...
+
+  // TLM Handlers
+
+  // NOTE - RJ - 2020-06-10
+  // a comment about the handling of live messages here:
+  // https://github.com/tracim/tracim/issues/3107#issuecomment-643994410
 
-    document.addEventListener(CUSTOM_EVENT.APP_CUSTOM_EVENT_LISTENER, this.customEventReducer)
+  liveMessageNotRelevant (data, state) {
+    return Number(data.content.workspace_id) !== Number(state.config.appConfig.workspaceId) ||
```
gbonaspetti, line 95: Is there a case where the `data.content.workspace_id` will not be a number?

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -53,55 +58,147 @@ export class Gallery extends React.Component {
...
+    Number(data.content.parent_id) !== Number(state.folderId)
```
gbonaspetti, line 96: Is there a case where the `data.content.parent_id` will not be a number?

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -53,55 +58,147 @@ export class Gallery extends React.Component {
...
+  }
+
+  // TLM Handlers
+
+  // NOTE - RJ - 2020-06-10
+  // a comment about the handling of live messages here:
+  // https://github.com/tracim/tracim/issues/3107#issuecomment-643994410
 
-    document.addEventListener(CUSTOM_EVENT.APP_CUSTOM_EVENT_LISTENER, this.customEventReducer)
+  liveMessageNotRelevant (data, state) {
```
gbonaspetti, line 94: Why do you need to receive the state here, won't it always be the component state?

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -343,65 +495,70 @@ export class Gallery extends React.Component {
...
   handleOpenDeleteFilePopup = () => {
-    this.setState({
-      displayPopupDelete: true
-    })
+    if (this.state.imagePreviewList.length) {
+      this.setState({
+        displayPopupDelete: true
+      })
+    } else {
+      this.sendGlobalFlashMessage(this.props.t('There isn’t any picture to delete.'))
```
gbonaspetti, line 537:
```suggestion
      this.sendGlobalFlashMessage(this.props.t("There isn't any picture to delete."))
```

frontend_app_gallery/src/container/Gallery.jsx
raphj: @Skylsmoi wanted this function to be pure. If you both decide that I should revert, I'm okay doing so.

gbonaspetti: don't need to, makes sense a pure function :smile:

frontend_app_gallery/src/container/Gallery.jsx
raphj: Ah ah xD
See: https://github.com/tracim/tracim/pull/3146/files/3cbfb880f7389e5bac621879f56e8685013c276a#r436679917
where I have written exactly the same thing as you to @Skylsmoi.

I think we should leave this as is for now and maybe write a ticket to solve this? (this is a backend bug, right, @Skylsmoi?

frontend_app_gallery/src/container/Gallery.jsx
raphj: See https://github.com/tracim/tracim/pull/3153#discussion_r440701992

frontend_app_gallery/src/container/Gallery.jsx
raphj: I'm okay with not using `Number()` for `data.content.workspace_id` if we are sure it's always a number though

gbonaspetti: The state part I understand it might be from an old code, but the data.content we're adding now with TLM, no? In that case we shouldn't be sure of the type?}

Note: I'm ok with leaving the Number, I'm just asking to understand, because I'm not guaranteeing that it's a number in my PRs, I thought that at least for the `data.content` we could be sure.

gbonaspetti:
Feedback from quality tests: 
- the name of the workspace is not changed when one user is in the gallery and another user change it (`workspace.modified` not handled).
- it is not possible to pass from the gallery of one workspace to the gallery of another (switch from galery to gallery by the sidebar)

grignards:
> Feedback from quality tests:
> 
>     * the name of the workspace is not changed when one user is in the gallery and another user change it (`workspace.modified` not handled).
> 
>     * it is not possible to pass from the gallery of one workspace to the gallery of another (switch from galery to gallery by the sidebar)

@gbonaspetti I corrected both points and updated the PR.

frontend_app_gallery/src/container/Gallery.jsx
```diff
@@ -188,6 +196,12 @@ export class Gallery extends React.Component {
     this.setState({ imagePreviewList, displayedPictureIndex })
   }
 
+  updateBreadcrumbsAndTitle = () => {
+    const { state, props } = this
+    this.buildBreadcrumbs(state.workspaceLabel, state.folderDetail, false)
```
Skylsmoi, line 214: There might be an issue in this function.

the setState function of react is asynchronous.
The advantages of that is that React can take upon himself to batch potential multiple setState calls into one to avoid too many calls to `render ()`.

In your case, you are not guaranteed that the state has been updated by `handleWorkspaceModified` when executing `updateBreadcrumbsAndTitle`.
Therefore, the `this.buildBreadcrumbs(state.workspaceLabel, state.folderDetail, false)` might be done with the old workspace label.

The solution to that it to make `updateBreadcrumbsAndTitle` pure and calling it from `handleWorkspaceModified` with the data from the TLM as parameters

inkhey:
@grignards. I've some trouble to distinct correctly broken case, but when i have 2 browser opened with same account on a shared space gallery with only this file (uncompress it to test):
[Tower_1.zip](https://github.com/tracim/tracim/files/4804563/Tower_1.zip)
I got into trouble with only one browser updated correctly.

grignards:
> @grignards. I've some trouble to distinct correctly broken case, but when i have 2 browser opened with same account on a shared space gallery with only this file (uncompress it to test):
> [Tower_1.zip](https://github.com/tracim/tracim/files/4804563/Tower_1.zip)
> I got into trouble with only one browser updated correctly when i delete the file from gallery view.

I saw your problem and made a small change that seems to improve this as I didn't reproduce it after (tried for at least 50 cycles of delete/undelete). I'm just not really confident that it isn't only a coincidence :confused:.

inkhey:
> > @grignards. I've some trouble to distinct correctly broken case, but when i have 2 browser opened with same account on a shared space gallery with only this file (uncompress it to test):
> > [Tower_1.zip](https://github.com/tracim/tracim/files/4804563/Tower_1.zip)
> > I got into trouble with only one browser updated correctly when i delete the file from gallery view.
> 
> I saw your problem and made a small change that seems to improve this as I didn't reproduce it after (tried for at least 50 cycles of delete/undelete). I'm just not really confident that it isn't only a coincidence confused.

I do reproduce the issue.

grignards:
> I do reproduce the issue.

I tried again and cannot reproduce it, I created [an issue](https://github.com/tracim/tracim/issues/3218) to follow this bug and will merge this PR as this bug is not a blocking one for the usage of the gallery

PR #3156, opened 2020-06-10
tracim client token

grignards:
Add client token identifying the browser's tab/window in frontend request headers.

Refs #3129

*Please indicate here a summary of the pull request content. If a related issue exists, please indicate issue reference here*

## Checkpoints

**For code reviewers**

**For developers**

**For the quality team**


frontend_lib/test/helper.spec.js
```diff
@@ -139,4 +144,24 @@ describe('helper.js', () => {
...
   })
+
+  describe('FETCH_CONFIG object', () => {
+    it('should include tracim client token header', () => {
+      expect('X-Tracim-ClientToken' in FETCH_CONFIG.headers).to.eq(true)
+      expect(FETCH_CONFIG.headers['X-Tracim-ClientToken']).to.be.a('string')
+    })
+
+    it('should store the client token in window session', () => {
+      expect(window.sessionStorage.getItem('tracimClientToken')).to.eq(FETCH_CONFIG.headers['X-Tracim-ClientToken'])
```
Skylsmoi, line 155: to be sure, here, sessionStorage exists because, in the previous test, you called FETCH_CONFIG which called COMMON_REQUEST_HEADERS which called getOrCreateSessionClientToken which put it in the sessionStorage.
right ?
If I'm right, I think it makes the tests dependent on each other and it would be preferable not to

grignards: The code putting the client token in the session storage is executed during the import of `helper.js` so I don't think we can say that the two tests are interdependent. I added this one because I didn't find a way to test that, even if you import `helper.js` several times the client token is the same.

Skylsmoi: okok good
you can resolve after read

grignards:
No quality test as this is not a user-visible feature, only a way to correct some buggy behaviors.

Issue #3091, opened 2020-05-21
Bug: Files created in session folders are never deleted


PhilippeAccorsi:
## Bug description & symptom

Tracim creates files in the `sessions_data/` and `sessions_lock` folders each time the Tracim login page is opened.

These files are never deleted, which clutters these folders. On one Tracim instance we have a total of 10M files just in these two folders. This uselessly increases the time to backup data and the space used by Tracim.

## Expected

- [x] explain how the mechanism work
- [x] We need a solution to clean existing installations without breaking them. We need to know quickly what we can done (before we make changes in the Tracim code)
- [x] explore @lebouquetin's idea to use Redis to store `sessions_data` and `sessions_lock` with Docker. -> Added support for redis storage.

## How to reproduce

- delete all files/folders in `sessions_data/` and `sessions_lock`
- go to the login page
- look at files and folders created in the `sessions_data/` and `sessions_lock` folders

## Implemented Solution:

See https://github.com/tracim/tracim/pull/3161
- add Documentation how to upkeep file backend session 
- adapt code to support any kind of session supported by beaker, document redis based session
- solve datetime issue related to use of timezoned date instead of utc one.
- remove reissue time and the concept of reuissing new session after some time, this is not required as beaker does update timeout at each access and this create lot of unneeded session.
- fix the issue in the code to avoid the creation of sessions for user that are not logged in.

inkhey:
These dirs can be safely deleted. The only think can may potentially cause trouble is in the case a user session is deleted at the same time user want to connect, but it's very unlikely and i will break anything.
This will break user active session, so, it's more something to do when we have to stop tracim instance.

To be more restrictive, we can do this:
`
find . -type f -mtime +10 -print -exec rm {} \;
`
in both session_lock and session_data dir. this will delete file older than 10 days (default session timeout in tracim is 604800 session = 7 day)

lebouquetin:
@inkhey dropping old sessions is not something to do when service is stopped:

- first we should be able to upgrade without stopping the instance (this is not yet possible according to the way we deploy, but technically, there are not restriction about this)
- second: even if tracim is stopped it does not mean we accept to drop all opened sessions.
- third, we must have a way to drop sessions during live execution (the `find` command seems to do so)

The issue is not about dropping session data but about "how we should/must manage it".

I don't know if you are still working on this issue, but the two checkbox must be checked soon (your current comment can be considered as a hotfix)

inkhey:
[...]

PR #3161, opened 2020-06-11
Feature/3091  allow redis backend storage for session without reissue time

inkhey:
related to #3091 

## Checkpoints

**For code reviewers**

**For developers**

**For the quality team**


README.md
```diff
@@ -173,6 +173,22 @@ You can now head to (if pushpin is correctly configured and use default port 799
...
+
+You should use this command in both session data and session lock dirs.
+
+Another way to solve this issue is to set session to use another backend(we do support all beaker backend).
+We suggest to use redis backend, which will not created file to disk:
+
+    session.type = ext:redis
+    session.url = redis://localhost:6379/0
+
+read [beaker documentation](https://beaker.readthedocs.io/en/latest/configuration.html) for more information about sessions.
```
raphj, line 191:
```suggestion
If you use the default "file" storage for sessions, new session files will be created but not removed, which leads to a lot of unneeded files.
To mitigate this, you may remove files older than 10 days with:

    find . -type f -mtime +10 -print -exec rm {} \;

You should use this command in both the session data and the session lock directories.

Another way to solve this issue is to use another storage backend (we do support any of the Beaker backend).
We suggest to use the redis backend, which will not create any file on the disk:

    session.type = ext:redis
    session.url = redis://localhost:6379/0

See the [Beaker documentation](https://beaker.readthedocs.io/en/latest/configuration.html) for more information about sessions.
```

backend/tracim_backend/applications/share/lib.py
```diff
@@ -60,7 +60,7 @@ def share_content(
         do_notify: bool = False,
     ) -> typing.List[ContentShare]:
         content_shares = []
-        created = datetime.now()
+        created = datetime.utcnow()
```
lebouquetin, line 63: why do you use `utcnow()` now? What if the server is not configured in UTC?

lebouquetin:
I do not see a mechanism or CLI command to remove expired sessions (user can have never logged out and the session expired -> how to remove these sessions?)

How are managed old sessions in databases like mongo/redis ? They are automatically deleted?

The documentation does not explain what storage backends are supported (it looks like redis is supported but there is no information: "session management supported backends". This is especially strange because the source code has some code about mongo and other databases but nothing  in the documentation.

So, reading tracim documentation is not enough to understand tracim -> the documentation is not enough

backend/development.ini.sample
```diff
@@ -187,10 +187,13 @@ preview_cache_dir = %(basic_setup.preview_cache_dir)s
...
 ### pyramid_beaker parameters ###
+
+# "File" storage config :
 session.type = file
-; session.data_dir = %(here)s/sessions_data
 session.data_dir = %(basic_setup.sessions_data_root_dir)s/sessions_data
-; session.lock_dir = %(here)s/sessions_data
+## "Redis" storage config sample:
+# session.type = ext:redis
+# session.url = redis://localhost:6379/0
```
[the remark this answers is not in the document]
lebouquetin: is there any way to give a name to redis databases? If not, I suggest we have a data related documentation which explain where are stored information (storage backend, location, etc).

In the other case, it is difficult to understand how the software works

backend/tracim_backend/applications/share/lib.py
inkhey: I do use utcnow() to have constant time mecanism. When checking for session, i do see a bug related to delta time calculation between utc based time and timezoned time (for reissue time which is deleted now), we do use utcnow() almost everywhere in code, and to prevent these issue, i do think we should always use utctime.

inkhey:
> I do not see a mechanism or CLI command to remove expired sessions (user can have never logged out and the session expired -> how to remove these sessions?)
> 

Mecanism to remove old session is different depending on backend. The documentation explain how to delete them for _file_ backend. There is no code in beaker to do this, so making a cli command to do this is possible but is completely tracim specific code.

> How are managed old sessions in databases like mongo/redis ? They are automatically deleted?

For mongo, i don't know. 
For redis, by checking database, i can tel that old managed session are deleted after timeout limit (the config parameter) + 2 minutes, when we reuse tracim with same session the timeout

> The documentation does not explain what storage backends are supported (it looks like redis is supported but there is no information: "session management supported backends". This is especially strange because the source code has some code about mongo and other databases but nothing in the documentation.

It depend on beaker as explain on documentation, it can be more clear, but i don't know which backend we should support officially, So i made possible to use any beaker backend we want but document but only file and redis which is known to be well supported.

> 
> So, reading tracim documentation is not enough to understand tracim -> the documentation is not enough

I can try to make thing more clear.

backend/tracim_backend/applications/share/lib.py
lebouquetin: so, what about deploying a tracim server on a non UTC configured server?

> When checking for session, i do see a bug related to delta time calculation between utc based time and timezoned time

What bug? If there is a bug in session mechanism, I'm not sure this is in the share library that we should fix it.

I do not see relationship between the description of bug #3091 and the bug you are explaining here. You should not fix bugs in other ones.

I feel like there are missing explanations for this MR to be accepted

README.md
```diff
@@ -173,6 +173,22 @@ You can now head to (if pushpin is correctly configured and use default port 799
 
 :warning: Using tracim without pushpin will mean no live message leading to unrefreshed frontend information.
 
+### Upkeep
```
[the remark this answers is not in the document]
inkhey: https://www.linguee.fr/anglais-francais/traduction/upkeep.html

As we do need to remove session periodically if we do use file storage for session

backend/development.ini.sample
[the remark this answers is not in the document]
inkhey: It's not a problem to use same database as both async (rq) and beaker don't use same "dir"/"table" in the database, so we can distinct data from each without any problem.

backend/tracim_backend/applications/share/lib.py
inkhey: >  so, what about deploying a tracim server on a non UTC configured server? 

Don't sure to understand. utcnow give a constant value accross server whereas now is timezone dependent. Using utcnow make thing constant, so it should have a constant result now.

> What bug? If there is a bug in session mechanism.
It renew session each time because the delta time difference was always 2h.

>  I'm not sure this is in the share library that we should fix it.

I'm not sure but i do prefer to be constant in all tracim code about this to prevent issue related to this. Because i do see this issue about now() and utcnow(), i do check usage of now() in all tracim code to verify if we could have similar issue in another place.

backend/development.ini.sample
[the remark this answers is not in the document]
grignards: Ok for keeping `0` as long as we are sure that this isn't a problem.

@lebouquetin I didn't see a way to have string database names with redis so yes we could add a line in the backend's README.

backend/doc/setting.md
```diff
@@ -199,6 +200,56 @@ The authentication is done with a login, which is either the user's email addres
...
+
+- the id of the user
+- the session creation datetime
+- the last-access to session datetime.
+
+Sessions are implemented with [Beaker](https://beaker.readthedocs.io/en/latest/configuration.html) and can be stored in several back-ends: files (the default), redis, mongodb, memcached, sql databases…
+Tracim is actively used and tested with 2 session back-ends: files and redis.
+
+We suggest to use redis backend for production as this is a tested backend which will not created file to disk and store session with a time limit,
+which mean no manual upkeep needed to remove expired session.
```
grignards, line 215:
```suggestion
Our recommended session back-end for production is redis as it avoids having to manage deletion of expired session files.```

File: backend/doc/setting.md [hunk 6]
```diff
@@ -199,6 +200,56 @@ The authentication is done with a login, which is either the user's email addres
...
+    session.key = session_key
+    session.secret = %(basic_setup.session_secret)s
+    session.save_accessed_time = True
+    session.cookie_expires = 604800
+    session.timeout = 604800
+    session.cookie_on_exception = True
+
+for other beaker backend, read [beaker documentation](https://beaker.readthedocs.io/en/latest/configuration.html) for more information.
+
+#### Upkeep file back-end
```
grignards, line 243:
```suggestion
#### File back-end upkeep
```

backend/doc/setting.md
```diff
@@ -199,6 +200,56 @@ The authentication is done with a login, which is either the user's email addres
...
+    # note: basic_setup.sessions_data_root_dir parameter should exist and be a real path
+    session.lock_dir = %(basic_setup.sessions_data_root_dir)s/sessions_lock
+    session.key = session_key
+    session.secret = %(basic_setup.session_secret)s
+    session.save_accessed_time = True
+    session.cookie_expires = 604800
+    session.timeout = 604800
+    session.cookie_on_exception = True
+
+for other beaker backend, read [beaker documentation](https://beaker.readthedocs.io/en/latest/configuration.html) for more information.
```
grignards, line 241:
```suggestion
for other beaker backends, read [beaker documentation](https://beaker.readthedocs.io/en/latest/configuration.html) for more information.
```

README.md
```diff
@@ -173,6 +173,17 @@ You can now head to (if pushpin is correctly configured and use default port 799
...
+
+If you use default "file" storage for session, new session file will be created but not removed, which mean lot of unneeded file.
+To mitigate this, you could remove file older than 10 day with :
+
+    find . -type f -mtime +10 -print -exec rm {} \;
+
+You should use this command in both session data and session lock dirs.
+
+Another way to solve this issue is to set session to use another backend(we do support all beaker backend),
+see [session](/backend/doc/setting.md) documentation for more information.
```
PhilippeAccorsi, line 186: For me, its better to indicate that its necessary to match day value by lifetime token. (and set value to default value of lifetime token. Example: if lifetime is 7days, day value can be 8 days).

backend/development.ini.sample
```diff
@@ -187,10 +187,14 @@ preview_cache_dir = %(basic_setup.preview_cache_dir)s
...
+
+# "File" storage config :
 session.type = file
-; session.data_dir = %(here)s/sessions_data
 session.data_dir = %(basic_setup.sessions_data_root_dir)s/sessions_data
-; session.lock_dir = %(here)s/sessions_data
+## "Redis" storage config sample:
+# session.type = ext:redis
+# note: redis db name are integer (by default 16 database are available from 0 to 15 db)
+# session.url = redis://localhost:6379/0
```
PhilippeAccorsi, line 197: We have make a big job on development.ini file Tracim 2.7, for standardize parameter. See top of development.ini file.

README.md
inkhey: changed to 8 days.

backend/development.ini.sample
inkhey: i just readd deleted default value, i do think it's now ok (unless we decided to change the default session method).

README.md
PhilippeAccorsi: thanks

backend/development.ini.sample
PhilippeAccorsi: Ok, thanks.

backend/tracim_backend/applications/share/lib.py
lebouquetin: @inkhey, @grignards explained me what you talked  about this and what you decided. Your decision is ok for me.

inkhey:
wait for clarification about docker ( see docker change here : https://github.com/tracim/tracim/pull/3227)

inkhey:
To test this:
- set storage to redis one, check is no file is created in session data and session lock dir as expected
- Build docker, check if redis storage is used by default (do not use existing development.ini to do this !), to do this, you can check if session data and lock dir remain empty whereas session work correctly.

If you want to make a more complete verification, you can use redis commande line or a redis gui like redis-desktop https://redisdesktop.com/ in order to see what's happening.

PR #3166, opened 2020-06-12
Feature/2850 handle TLM at Home.jsx

gbonaspetti:
#2850

## Checkpoints

**For code reviewers**

**For developers**

**For the quality team**


frontend/test/src/container/Home.spec.js
```diff
@@ -18,29 +38,51 @@ describe('<Home />', () => {
...
 
   describe('static design', () => {
     it('should render the root div', () =>
-      expect(wrapper.find('div.tracim__content').length).equal(1)
+      expect(homeWrapper.find('div.tracim__content').length).equal(1)
     )
 
     it('should not render if workspaceList is not loaded', () => {
-      wrapper.setProps({ system: { ...props.system, workspaceListLoaded: false } })
+      wrapper.setProps({ sys: { ...props.system, workspaceListLoaded: false } })
```
Skylsmoi, line 77: I don't think this prop exists for component Home, so I don't think this test is doing what it's supposed to

gbonaspetti: [It's not possible use setProps to a children when mounting a element](https://github.com/enzymejs/enzyme/issues/1925), so I based on this idea https://github.com/enzymejs/enzyme/issues/1925#issuecomment-490637648 and what I do is I create a prop sys in the component and I pass this prop to the children's prop `system`.
I can use the name `system` for this "fake" prop too if you find it more clear.

Skylsmoi: mmh ok but, your component is in `wrapper` and it never do anything related to a prop named `sys` so I don't understand what you are testing here

gbonaspetti: sys is passed to component HomeWithHOC2 which passes the sys value to the Home component's prop system.
I just update the test _should not render if workspaceList is not loaded_ to works with a `mount` element.

Edit: I changed the names in my last commit to be more understandable.

frontend/src/container/Home.jsx
```diff
@@ -53,7 +55,9 @@ export class Home extends React.Component {
     ])
 
     props.registerLiveMessageHandlerList([
-      { entityType: TLM_ET.USER, coreEntityType: TLM_CET.MODIFIED, handler: this.handleUserModified }
+      { entityType: TLM_ET.USER, coreEntityType: TLM_CET.MODIFIED, handler: this.handleUserModified },
+      { entityType: TLM_ET.SHAREDSPACE_MEMBER, coreEntityType: TLM_CET.CREATED, handler: this.handleWorkspaceMemberCreated },
+      { entityType: TLM_ET.SHAREDSPACE_MEMBER, coreEntityType: TLM_CET.DELETED, handler: this.handleWorkspaceMemberDeleted }
```
Skylsmoi, line 60: I don't think these handler will affect the Home.jsx component.
They will affect the sidebar though
In that case, I think you should let the Sidebar.jsx component handle them

frontend/src/reducer/workspaceList.js
```diff
@@ -1,34 +1,41 @@
...
-        agendaEnabled: ws.agenda_enabled,
-        uploadEnabled: ws.public_upload_enabled,
-        downloadEnabled: ws.public_download_enabled,
+        ...serialize(ws, serializeWorkspaceListProps),
+        sidebarEntryList: ws.sidebar_entries.map(sbe => serialize(sbe, serializeSidebarEntryProps)),
         memberList: []
       }))
 
+    case `${REMOVE}/${WORKSPACE_LIST}`:
+      return action.workspaceList.filter(ws => ws.id !== action.workspaceId)
```
Skylsmoi, line 37: a reducer should almost always return a value from it's previous one (which is the state).
Unless you explicitly wants to override it (like in RESET or SET actions for example).

Your action must not give action.workspaceList and should take it from it's state.

Skylsmoi: but then again, I think this should be done in the Sidebar.jsx refactor

frontend/src/container/Home.jsx
gbonaspetti: When the user has no workspace and someone adds it to some workspace, the Home screen changes and the opposite is also valid. If he only has one workspace and someone deletes it he should render HomeNoWorkspace instead of HomeHasWorkspace.

frontend/test/src/reducer/user.spec.js
```diff
@@ -26,22 +24,15 @@ describe('user reducer', () => {
 
   it(`should handle ${UPDATE}/${USER}`, () => {
     expect(
-      userReducer(defaultUser, updateUser({ ...globalManagerFromApi, public_name: 'newPublicName' }))
+      userReducer(
+        { ...defaultUser, userId: globalManagerFromApi.user_id },
+        updateUser({ ...globalManagerFromApi, public_name: 'newPublicName' }))
```
Skylsmoi, line 29: one closing parenthesis should be on a new line

gbonaspetti:
To quality team:
- Test user changes (especially changing his public name to see if it changes correctly in the homepage).

PR #3173, opened 2020-06-12
Feature/2850 handle TLM at SearchResult.jsx

gbonaspetti:
#2850 

## Checkpoints

**For code reviewers**

**For developers**

**For the quality team**


frontend/src/reducer/searchResult.js
```diff
@@ -16,16 +45,31 @@ const defaultResult = {
...
+  return uniqBy(allResults, 'contentId')
 }
 
 export default function searchResult (state = defaultResult, action) {
   switch (action.type) {
     case `${SET}/${SEARCH_RESULTS_LIST}`:
-      return { ...state, resultsList: uniqueResults(action.newSearchResultsList) }
+      return {
+        ...state,
+        resultsList: uniqueResults((action.newSearchResultsList).map(item => serialize(item, serializeSearchItemProps)))
```
Skylsmoi, line 56: we usually don't put parenthesis around the array we are using .map() on.
There isn't any rules though

frontend/src/reducer/searchResult.js
```diff
@@ -16,16 +45,31 @@ const defaultResult = {
 }
 
 function uniqueResults (allResults) {
-  return uniqBy(allResults, 'content_id')
+  return uniqBy(allResults, 'contentId')
```
Skylsmoi, line 48: I know this function isn't new in this PR but it isn't named well.
This name is too generic, it feels that we can call it on any array and it'll guaranty unicity on it. But internally, it only filters on "content_id"

You have to name utility function from what they are doing and not from their context of use.
Here, it will filter the search results so the name uniqueResults make sense but if we look at it's behavior, all it does is filtering an array on the property contentId.
So I would call it uniqueArrayByContentId, or uniqueByContentId (listToFilter)

Skylsmoi: alternatively, while, in the reducers, we usually do everything in the return statement, it is only because it usually is easy enough to do so.
But you can write multiple statements to have something like this:
``` javascript
const newResultList =  action.newSearchResultsList.map(item => serialize(item, serializeSearchItemProps))
const uniqueResultList = uniqBy(newResultList, 'contentId')
return { ...state, resultsList: uniqueResultList }
```
instead of:
``` javascript
return {
  ...state,
  resultsList: uniqueResults((action.newSearchResultsList).map(item => serialize(item, serializeSearchItemProps)))
}
```
In this particular case, it is even one line shorter

frontend/src/reducer/searchResult.js
```diff
@@ -16,16 +45,31 @@ const defaultResult = {
...
+          ? { ...item, ...serialize(action.searchResultItem, serializeSearchItemProps) }
+          : item
+        ))
+      }
 
     case `${APPEND}/${SEARCH_RESULTS_LIST}`:
-      return { ...state, resultsList: uniqueResults([...state.resultsList, ...action.appendSearchResultsList]) }
+      return {
+        ...state,
+        resultsList: uniqueResults([...state.resultsList, ...((action.newSearchResultsList).map(item => serialize(item, serializeSearchItemProps)))])
```
Skylsmoi, line 71: This line is starting to be a bit too complicated.
The thing that tells me that is the `)))])` at the end of it

frontend/test/src/container/SearchResult.spec.js
```diff
@@ -0,0 +1,130 @@
...
+      it('should return the content path', () => {
+        const parentList = [
+          {
+            label: 'first'
+          },{
+            label: 'second'
+          }
+        ]
+        expect(searchResultInstance.getPath(parentList)).to.equal(
+          `${parentList[1].label} / ${parentList[0].label} / `
```
Skylsmoi, line 110: if this test pass, I think there might be something wrong since an url cannot have spaces in it.
This indicates a potential bug in getPath or `.equal()` isn't the proper function to use

gbonaspetti: This path is not the url, is the path inside Tracim (folder, subfolder,...) which is shown in the search

Skylsmoi: oh I see

frontend/src/container/ReduxTlmDispatcher.jsx
```diff
@@ -88,10 +89,12 @@ export class ReduxTlmDispatcher extends React.Component {
 
   handleContentModified = data => {
     this.props.dispatch(updateWorkspaceContentList([data.content], data.workspace.workspace_id))
+    this.props.dispatch(updateSearchResultList(data.content))
```
Skylsmoi, line 92: couldn't the search result reducer be updated by the action `updateWorkspaceContentList` ?

That would mean the `reducer/searchResult.js` would handle the action `${UPDATE}/${WORKSPACE_CONTENT}`

frontend/src/container/ReduxTlmDispatcher.jsx
```diff
@@ -88,10 +89,12 @@ export class ReduxTlmDispatcher extends React.Component {
...
   }
 
   handleContentDeleted = data => {
     this.props.dispatch(deleteWorkspaceContentList([data.content], data.workspace.workspace_id))
+    this.props.dispatch(updateSearchResultList(data.content))
```
Skylsmoi, line 97: same comment as `handleContentModified`

frontend/src/container/ReduxTlmDispatcher.jsx
gbonaspetti: Functionally, there are no restrictions. The only thing "wrong" would be the naming, because the search result has no connection with the workspace.

Skylsmoi: that's true.
But I don't think it is an issue.
The word "dispatch" in `props.dispatch()` is meaningful, it broadcast the action and every reducers that are interested in implements it

I think it's better than having multiple actions for the same thing

gbonaspetti: Ok, works for me. I'm gonna change. :smile:

Issue #3106, opened 2020-05-29
Feat: Handle TLM for app folder advanced


Skylsmoi:
## Feature description

Refactor the app folder advanced with TLM usage

## Context

The v3.0 is the live update version. Apps must trigger update through this process

## Impacted components

App Folder advanced

## Implementation

- Folder title is updated silently when the folder is changed by another user/browser
- The list of checked content types is updated silently when it is changed by another user/browser
- The "this folder is deleted" warning is displayed if the folder is deleted by another user/browser

PhilippeAccorsi:
When I tested this issue, I have seen something broken, explained here #3217 
Need more test when database migration is fixed.

gbonaspetti:
Verified on develop.

PR #3195, opened 2020-06-18
Feature/3106  tlm in app folder advanced

grignards:
See #3106.

## Checkpoints

**For code reviewers**

**For developers**

- [X] Manual tests done to ensure stability on whole application layers, issue expectation follow
- [X] The original issue has been updated with a short summary of the implemented solution

**For the quality team**

### DIFF
--- a/frontend_app_folder_advanced/src/container/FolderAdvanced.jsx
+++ b/frontend_app_folder_advanced/src/container/FolderAdvanced.jsx
@@ -28,8 +28,10 @@ import {
 } from '../action.async.js'
 
 const filterSubContentTypes = (list) => {
-  // comments cannot be made on a folder, so remove them
-  // from the possible content types
+  /* INFO - SG - 2020-06-18
+   * Comments cannot be made on a folder, so remove them
+   * from the possible content types
+   */
   return list.filter(ct => ct.slug !== 'comment')
 }
 
--- a/frontend_app_folder_advanced/src/container/FolderAdvanced.jsx
+++ b/frontend_app_folder_advanced/src/container/FolderAdvanced.jsx
@@ -27,6 +27,12 @@ import {
   putFolder
 } from '../action.async.js'
 
+const filterSubContentTypes = (list) => {
+  // comments cannot be made on a folder, so remove them
+  // from the possible content types
+  return list.filter(ct => ct.slug !== 'comment')
+}
+
 export class FolderAdvanced extends React.Component {
   constructor (props) {
     super(props)
@@ -64,10 +70,6 @@ export class FolderAdvanced extends React.Component {
       { entityType: TLM_ET.CONTENT, coreEntityType: TLM_CET.DELETED, optionalSubType: TLM_ST.FOLDER, handler: this.handleFolderChanged },
       { entityType: TLM_ET.CONTENT, coreEntityType: TLM_CET.UNDELETED, optionalSubType: TLM_ST.FOLDER, handler: this.handleFolderChanged }
     ])
-  }
-
-  static filterSubContentTypes (list) {
-    return list.filter(ct => ct.slug !== 'comment')
   }
 
   handleAllAppChangeLanguage = async data => {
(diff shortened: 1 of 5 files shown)


frontend_app_folder_advanced/src/container/FolderAdvanced.jsx
```diff
@@ -47,39 +51,62 @@ class FolderAdvanced extends React.Component {
...
+
+    props.registerLiveMessageHandlerList([
+      { entityType: TLM_ET.CONTENT, coreEntityType: TLM_CET.MODIFIED, optionalSubType: TLM_ST.FOLDER, handler: this.handleFolderChanged },
+      { entityType: TLM_ET.CONTENT, coreEntityType: TLM_CET.DELETED, optionalSubType: TLM_ST.FOLDER, handler: this.handleFolderChanged },
+      { entityType: TLM_ET.CONTENT, coreEntityType: TLM_CET.UNDELETED, optionalSubType: TLM_ST.FOLDER, handler: this.handleFolderChanged }
+    ])
   }
 
-  customEventReducer = ({ detail: { type, data } }) => {
+  static filterSubContentTypes (list) {
```
Skylsmoi, line 69: do you really need it to be static ? 
since you only use it inside the class, you could use `this.` instead of `FolderAdvanced.`

frontend_app_folder_advanced/src/container/FolderAdvanced.jsx
```diff
@@ -136,7 +157,7 @@ class FolderAdvanced extends React.Component {
     }
 
     switch (fetchContentTypeList.apiResponse.status) {
-      case 200: this.setState({ tracimContentTypeList: fetchContentTypeList.body.filter(ct => ct.slug !== 'comment') }); break
+      case 200: this.setState({ tracimContentTypeList: FolderAdvanced.filterSubContentTypes(fetchContentTypeList.body) }); break
```
Skylsmoi, line 160: see comment about `static`

frontend_app_folder_advanced/test/container/FolderAdvanced.spec.js
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        const wrapper = shallow(<FolderAdvanced {...props} />)
+
+        before(() => {
+          dispatchEventSpy.resetHistory()
+          wrapper.instance().handleFolderChanged(tlmData)
+        })
+
+        it('should update the component\'s folder', () => {
```
Skylsmoi, line 65: standardjs linting rules prefer using double quotes string for string containing a quote instead of backslashing them

frontend_app_folder_advanced/test/container/FolderAdvanced.spec.js
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        const wrapper = shallow(<FolderAdvanced {...props} />)
+
+        before(() => {
+          dispatchEventSpy.resetHistory()
+          wrapper.instance().handleFolderChanged(tlmData)
+        })
+
+        it('should NOT update the component\'s folder', () => {
```
Skylsmoi, line 89: standardjs linting rules prefer using double quotes string for string containing a quote instead of backslashing them

frontend_app_folder_advanced/src/container/FolderAdvanced.jsx
grignards: The advantage I see to use a static method is that it makes clear that it does not modify something on `this`… at the price of of less "elegant" way of calling it (which I suppose is what you point out). I could also define it as a function outside the class, what would be your opinion on this way?

Skylsmoi: I asked myself this kind of question several times but couldn't find an answer that satisfy me completely.

I do like the fact, as you said, that is makes clear that it doesn't modify `this`
I don't mind the less elegant way of calling it

But it does bother me a little bit to use a specific keyword for something that it isn't intended to.

So I guess I would opt for the outside of class function

frontend_app_folder_advanced/src/container/FolderAdvanced.jsx
```diff
@@ -27,6 +27,12 @@ import {
   putFolder
 } from '../action.async.js'
 
+const filterSubContentTypes = (list) => {
+  // comments cannot be made on a folder, so remove them
```
Skylsmoi, line 31: don't forget the comment rules
`// INFO - SG - 2020-06-18 - comment...`

inkhey:
Verified on branch
Just a remark about current implementation: 
If 2 person are both editing the title of the folder, it mean that the title will be updated lively for the second person to apply the change. This is weird.

grignards:
> Just a remark about current implementation:
> If 2 person are both editing the title of the folder, it mean that the title will be updated lively for the second person to apply the change. This is weird.

I agree this is disturbing, I created an issue to handle this behavior (which is shared by all similar apps): #3199.

Issue #3201, opened 2020-06-19
Feat: Fix rights access on TLM


grignards:
## Feature description

As currently implemented the content TLMs are sent to all administrators, this should be changed to only send those to the content's workspace members.

Another issue is the fact that non-administrator and non author of user TLM do not receive TLM of user known (user in same workspace as the user), which mean user list cannot been updated properly, lively.

## Context

Administrators do not have direct access to all workspace contents, even if they can make themselves member of any workspace.

## Prerequisites

None

## Impacted components

- *backend*

## Extra information

The code to change is in `backend/tracim_backend/lib/core/event.BaseLiveMessageBuilder._publish_messages_for_event()`.

Implemented solution:
- Allow only workspace members to get content event.
- Allow initiator, administrator and user who share workspace with the TLM initiator to see user TLM.
- Small refactoring in UserLIb in order to avoid code duplication.

inkhey:
I do find another issue about TLM access : other Non-administrator doesn't get access to user tlm of "known users" (users which have common workspace with you).

As this bug is found when doing tests about right access for this one and fix is not complicated, i do created a fix in the PR : 
https://github.com/tracim/tracim/pull/3202

PhilippeAccorsi:
Tested on branch. wait merge to close.

gbonaspetti:
Verified on develop.

PR #3202, opened 2020-06-19
limit content event to workspace members only

inkhey:
related to #3201 

## Checkpoints

**For code reviewers**

**For developers**

**For the quality team**

### DIFF
--- a/backend/tracim_backend/lib/core/user.py
+++ b/backend/tracim_backend/lib/core/user.py
@@ -226,25 +226,35 @@ class UserApi(object):
         # should show only user in same workspace as user
         assert not (filter_results and not self._user)
         if filter_results and self._user and self._user.profile.id <= Profile.USER.id:
-            user_workspaces_id_query = (
-                self._session.query(UserRoleInWorkspace.workspace_id)
-                .distinct(UserRoleInWorkspace.workspace_id)
-                .filter(UserRoleInWorkspace.user_id == self._user.user_id)
-            )
-            users_in_workspaces = (
-                self._session.query(UserRoleInWorkspace.user_id)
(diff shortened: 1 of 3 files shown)


backend/tracim_backend/lib/core/user.py
```diff
@@ -226,25 +226,35 @@ def get_known_user(
...
         if exclude_user_ids:
             query = query.filter(~User.user_id.in_(exclude_user_ids))
         query = query.limit(nb_elem)
         return query.all()
 
-    # Check methods
+    def _get_user_ids_in_same_workspace(self, user_id: int):
+        user_workspaces_id_query = (
+            self._session.query(UserRoleInWorkspace.workspace_id)
+            .distinct(UserRoleInWorkspace.workspace_id)
```
grignards, line 239: Is `distinct` necessary here ? (which would imply that a user can have several roles in a workspace).

inkhey: This code is not really new, i just moved it in order to reuse it, so i didn't check this.

UserRoleInWorkspace Table contain both user_id and workspace id as primary key, so we can have multiple item with different workspace id. It's right that normally, if we do select for user, there is no reason to get multiple item for same workspace id (it will brake primary key unicity).

inkhey: removed.

inkhey:
To test this: Check TLM received (from backend) and verify if:
- Simple user (not administrator) can receive user event from other user which belong to the same workspace
- Administrator do not receive content TLM for content in workspace administator is not a member

PhilippeAccorsi:
Tested on branch.

Issue #3238, opened 2020-06-29
Bug: Standardize and factorize the behavior of messages related to TLM


gbonaspetti:
## Bug description & symptom

Need to standardize and factorize the behavior of messages intended for the user related to TLM.

## Expected

- [x] for **frontend**, **frontend_app_admin_workspace_user**, **frontend_app_gallery**, **frontend_app_workspace_advanced** and **timeline at appFeature** - all updates are made automatically,  i.e. without a warning message, more details for each action:

- workspace.created
    - update sidebar automatically
    - update workspaces list at admin account automatically
- workspace.modified
    - update sidebar automatically
    - update dashboard/gallery/workspace_advanced title and description automatically
    - update breadcrumbs automatically
    - update shared spaces and notifications list at user's account automatically
    - update workspaces list at admin account automatically
    - update optional functionalities at workspace_advanced automatically
- workspace.deleted
    - update sidebar automatically
    - update workspaces list at admin account automatically
- member.created
    - update members list at dashboard/workspace_advanced automatically
    - update shared spaces and notifications list at user's account automatically
    - update workspaces list at admin account automatically
- member.modified
    - update members list at dashboard/workspace_advanced automatically
    - update shared spaces and notifications list at user's account automatically
- member.deleted
    - update members list at dashboard/workspace_advanced automatically
    - update shared spaces and notifications list at user's account automatically
    - update workspaces list at admin account automatically
- content.created
    - update workspace content list automatically
    - update recent activities list at dashboard automatically
    - update share folder content list automatically if needed
    - update gallery automatically if needed
- content.comment.created
    - update recent activities list at dashboard automatically
    - add a comment on the timeline at appFeature automatically
- content.modified
    - update workspace content list automatically
    - update content at search result automatically
    - update recent activities list at dashboard automatically
    - create a revision on the timeline at appFeature automatically
    - update gallery automatically if needed
- content.deleted
    - update content's name at search result automatically
    - add a comment on the timeline at appFeature automatically
    - update gallery automatically if needed
- content.undeleted
    - update recent activities list at dashboard automatically
    - add a comment on the timeline at appFeature automatically
    - update gallery automatically if needed
- user.created
    - update users list at admin account automatically
- user.modified
    - update members list and user information at dashboard automatically
    - update accounts info at account and at adminAccount automatically
    - update dashboard and homepage automatically if public name is changed
    - update users list at admin account automatically
- user.deleted
    - update users list at admin account automatically

[...]

PR #3252, opened 2020-06-30
Fix/3238  standardize and factorize refresh messages

gbonaspetti:
#3238 

## Checkpoints

**For code reviewers**

**For developers**

**For the quality team**


frontend_app_folder_advanced/src/container/FolderAdvanced.jsx
```diff
@@ -106,11 +109,17 @@ export class FolderAdvanced extends React.Component {
...
-    if (message.content.content_id !== state.content.content_id) return
-    this.setState({ content: message.content })
-    this.setHeadTitle(message.content.label)
+    if (data.content.content_id !== state.content.content_id) return
+
+    this.setState(prev => ({
+      content: prev.loggedUser.userId === data.author.user_id ? { ...prev.content, ...data.content } : prev.content,
+      newContent: { ...prev.content, ...data.content },
+      editionAuthor: data.author.public_name,
+      hasUpdated: prev.loggedUser.userId !== data.author.user_id
```
Skylsmoi, line 120: it seems that the name `hasUpdated` means that the logged user did the update we just received.
I would like a more explicit property name then

frontend_app_folder_advanced/test/container/FolderAdvanced.spec.js
```diff
@@ -45,25 +45,32 @@ describe('<FolderAdvanced />', () => {
...
+            user_id: 1
+          },
           content: {
             ...folder,
             label: 'Hello, world'
           }
         }
 
-        const wrapper = shallow(<FolderAdvanced {...props} />)
+        // const wrapper = shallow(<FolderAdvanced {...props} />)
```
Skylsmoi, line 65: if you don't use this line anymore, feel free to remove it

frontend_app_folder_advanced/test/container/FolderAdvanced.spec.js
```diff
@@ -73,26 +80,46 @@ describe('<FolderAdvanced />', () => {
...
+            user_id: 1
+          },
           content: {
             ...folder,
             content_id: 2
           }
         }
 
-        const wrapper = shallow(<FolderAdvanced {...props} />)
+        // const wrapper = shallow(<FolderAdvanced {...props} />)
```
Skylsmoi, line 94: if you don't use this line anymore, feel free to remove it

frontend_app_folder_advanced/src/container/FolderAdvanced.jsx
Skylsmoi: EDIT:
I'm now seeing that this property is only an addition in this app and that it already is named `hasUpdated` in others.
I'd still like you to rename them all but if it is too complicated, you can keep it that way.
I should have seen it in the first PR adding it. It's not a big issue anyway

frontend_lib/src/component/RefreshWarningMessage/RefreshWarningMessage.styl
```diff
@@ -2,15 +2,24 @@
...
-  margin 10px 30px 0px
-  white-space nowrap
+  cursor pointer
   .fa
-    margin auto 5px
-  &__button
-    border none
-    background-color transparent
-    color darkGrey
+    margin auto
```
Skylsmoi, line 13: do notice that `margin-top: auto` and `margin-bottom: auto` doesn't work, it set margin-top and bottom to 0.
It probably isn't an issue in your case though

frontend_app_folder_advanced/src/container/FolderAdvanced.jsx
gbonaspetti: Pretty easy to change, a little less easy to think of a good name. I'll change it to `receivedUpdate`, but if you have an idea of a better name, you can say I don't mind changing it again.

frontend_app_folder_advanced/test/container/FolderAdvanced.spec.js
gbonaspetti: Changed

frontend_app_folder_advanced/test/container/FolderAdvanced.spec.js
gbonaspetti: Changed

frontend_lib/src/component/RefreshWarningMessage/RefreshWarningMessage.styl
gbonaspetti: The value `auto`, set the margin at an "suitable" value, then use `margin-top: auto` and `margin-bottom: auto` is a way to center an element vertically. Using this PR example, here are the screenshots without `margin: auto` and with.
Screenshot from 2020-06-30 19-34-44 Screenshot from 2020-06-30 19-34-49
(Maybe it's not super clear in the screenshots, but if you look at the icon compared to the top of the letter R you'll see that one is centralized and the other is not)

frontend_app_folder_advanced/src/container/FolderAdvanced.jsx
Skylsmoi: mmh since it is related to the logged user, I'd like to reflect that info.
`isUpdateFromLoggedUser`, `loggedUserHasUpdated`, isAuthorLoggedUser` 
Don't be afraid of long variable names, especially when they are "kinda" global (it's not global here by it is global to the app)

Skylsmoi: I see you already did the refactor, sorry to answer that late

gbonaspetti: In fact, it depends, for this component it is related to the loggedUser, but for the agenda for example it is related to which page the user is on. Is it better to have more specific but different names between the components or the same name, since the variable has the same function, but more generic?

frontend_lib/src/component/RefreshWarningMessage/RefreshWarningMessage.styl
Skylsmoi: I have tested it and it indeed seems to do what is expected

Skylsmoi: keep in mind that using margin auto to center vertically might not be the most reliable way. I'm going to look a bit into why it works because I'm a bit confused here ^^

frontend_app_folder_advanced/src/container/FolderAdvanced.jsx
Skylsmoi: tough question.
I think since the apps are stand alone React apps we can allow different variable names of the same job.
Ideally, the name is different enough to clearly reflect its usage in each apps and similar enough to understand that these variable are equivalents in each apps.
Good luck with that : )

frontend_lib/src/component/RefreshWarningMessage/RefreshWarningMessage.styl
Skylsmoi: I confirm that margin-top auto and margin-bottom auto are interpreted as 0 (see https://developer.mozilla.org/en-US/docs/Web/CSS/margin part "more example")

But I don't see how it works on your branch

Skylsmoi: I just realized the parent of the icon (`.refreshWarningMessage`) is `display: flex`. I'd advise you to give it the property `align-items: center;` and remove margin-top auto and margin-bottom auto

It'll be cleaner

gbonaspetti: I looked at it too and you're right, but the behavior of `margin: auto` changes when the parent element has `display: flex`. 
See: https://css-tricks.com/the-peculiar-magic-of-flexbox-and-auto-margins/

Skylsmoi: mmh
nice catch, today I learned

frontend_app_folder_advanced/src/container/FolderAdvanced.jsx
gbonaspetti: and if I put `showRefreshWarning`, do you like it? It seems to me more descriptive about the variable's purpose and works for all components.

Skylsmoi: I think it's great. As long as it is actually only used for that purpose

gbonaspetti: so we have a winner! I will update the code

frontend_lib/src/component/RefreshWarningMessage/RefreshWarningMessage.styl
gbonaspetti: Cool :smile:. Anyway, I decided to switch to something less "magical".

inkhey:
Just tested, this PR missing correct support for deletion/restore case.

frontend_app_agenda/src/css/index.styl
```diff
@@ -9,3 +9,15 @@
...
 .pageContentGeneric.agendaPage
   height calc(100% - 135px)
+
+.agendaPage
+  &__warningMessage
+    position absolute
+    right 25px
+    top 110px
+
+// MEDIA QUERIES
```
raphj, line 19: You can safely leave out this comment I think :-)

gbonaspetti: I put because almost all components have, but I agree with you

raphj: Ah, ok :-)

raphj:
I tested the branches. I saw two issues:

- the text in the refresh button is not vertically aligned
- the tooltip does not seem to show up in the html app.

Issue #3242, opened 2020-06-29
Bug: TimeLine doesn't need to auto-scroll when the timline scroll is not at the bottom


metzgegu:
## Bug description & symptom

For now when a new timeline item is created, the timeline will auto scroll to the bottom in order to see the new item
Now with TLM, The autoscroll is active even if the item was not created by the current user, so if the current user is reading some old comments it will scroll to bottom and this is not user-friendly

## Expected/Implementation

When the timeline is modified, it must check if the timeline scroll is already at the bottom (old bottom) to scroll to the new comment, and it need to scroll to the bottom when the new Timeline item added was created by the current user & current session (with Token_id)

gbonaspetti:
The solution works well for large screens, but on smaller screens, since we are at the bottom of the document, it does not remain there, we need to scroll again each time. Check gif:
![scroll](https://user-images.githubusercontent.com/35422692/86231855-740fca80-bb93-11ea-9a57-1b7f13606bdd.gif)

metzgegu:
Autoscroll is disabled when display `width < 1200px ` for now, but we could scroll to the bottom when a new item is added in the timeline and the scroll is already at the bottom when `width < 1200px`

metzgegu:
The current implementation had an issue when the content was changed without close the app, 
I fixed it in the last commit
And I added the auto-scroll in small screen when the current User/TokenID added a comment (and only a comment) in order to have a more friendly behavior 
see :

![TimelineDemo2](https://user-images.githubusercontent.com/31687635/86344059-185c4480-bc5a-11ea-8da0-058f1be2cab9.gif)

gbonaspetti:
Verified on develop.

PR #3253, opened 2020-06-30
Fix/3242 Timeline fix auto-scroll when new items arrives from TLM

metzgegu:
#3242 

## Checkpoints

**For code reviewers**

**For developers**

**For the quality team**

### DIFF
--- a/frontend_app_file/src/container/File.jsx
+++ b/frontend_app_file/src/container/File.jsx
@@ -309,7 +309,8 @@ export class File extends React.Component {
               `${state.config.apiUrl}/workspaces/${state.content.workspace_id}/files/${state.content.content_id}/revisions/${response.body.current_revision_id}/preview/jpg/1920x1080/${filenameNoExtension + '.jpg'}?page=${i + 1}`
             )
           },
-          mode: APP_FEATURE_MODE.VIEW
+          mode: APP_FEATURE_MODE.VIEW,
+          isLastTimelineItemCurrentToken: false
         })
         this.setHeadTitle(filenameNoExtension)
         break
--- a/frontend_app_file/src/container/File.jsx
+++ b/frontend_app_file/src/container/File.jsx
@@ -38,7 +38,8 @@ import {
   FILE_PREVIEW_STATE,
   sortTimelineByDate,
   addRevisionFromTLM,
-  setupCommonRequestHeaders
+  setupCommonRequestHeaders,
+  getOrCreateSessionClientToken
 } from 'tracim_frontend_lib'
 import { PAGE, isVideoMimeTypeAndIsAllowed, DISALLOWED_VIDEO_MIME_TYPE_LIST } from '../helper.js'
 import { debug } from '../debug.js'
@@ -91,7 +92,8 @@ export class File extends React.Component {
       shareLinkList: [],
       previewVideo: false,
       hasUpdated: false,
-      editionAuthor: ''
+      editionAuthor: '',
+      isLastTimelineItemCurrentToken: false
     }
     this.refContentLeftTop = React.createRef()
 
@@ -167,7 +169,8 @@ export class File extends React.Component {
       newContent: newContentObject,
       editionAuthor: data.author.public_name,
       hasUpdated: prev.loggedUser.userId !== data.author.user_id,
-      timeline: addRevisionFromTLM(data, prev.timeline, prev.loggedUser.lang)
+      timeline: addRevisionFromTLM(data, prev.timeline, prev.loggedUser.lang),
+      isLastTimelineItemCurrentToken: data.client_token === getOrCreateSessionClientToken()
     }))
   }
 
@@ -182,7 +185,10 @@ export class File extends React.Component {
           timelineType: data.content.content_type
         }
       ])
-      this.setState({ timeline: sortedNewTimeLine })
+      this.setState({
+        timeline: sortedNewTimeLine,
+        isLastTimelineItemCurrentToken: data.client_token === getOrCreateSessionClientToken()
+      })
     }
   }
 
@@ -199,7 +205,8 @@ export class File extends React.Component {
           ...data.content
         },
         mode: APP_FEATURE_MODE.VIEW,
-        timeline: addRevisionFromTLM(data, prev.timeline, prev.loggedUser.lang)
+        timeline: addRevisionFromTLM(data, prev.timeline, prev.loggedUser.lang),
+        isLastTimelineItemCurrentToken: data.client_token === getOrCreateSessionClientToken()
       })
     )
   }
@@ -215,7 +222,8 @@ export class File extends React.Component {
         ...prev.content,
         ...data.content
       },
-      timeline: addRevisionFromTLM(data, prev.timeline, prev.loggedUser.lang)
+      timeline: addRevisionFromTLM(data, prev.timeline, prev.loggedUser.lang),
+      isLastTimelineItemCurrentToken: data.client_token === getOrCreateSessionClientToken()
     }))
   }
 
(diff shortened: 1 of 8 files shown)


frontend_lib/src/component/Timeline/Timeline.jsx
```diff
@@ -9,40 +9,50 @@ import i18n from '../../i18n.js'
...
-class Timeline extends React.Component {
+export class Timeline extends React.Component {
   constructor (props) {
     super(props)
-    document.addEventListener(CUSTOM_EVENT.APP_CUSTOM_EVENT_LISTENER, this.customEventReducer)
+    props.registerCustomEventHandlerList([
+      { name: CUSTOM_EVENT.ALL_APP_CHANGE_LANGUAGE, handler: this.handleAllAppChangeLanguage }
+    ])
+
+    this.timelineContainerScrollHeight = 0
```
gbonaspetti, line 24: Is there a reason this variable can't be a state?

metzgegu: Yes, it isn't in the state because it is the result of a sum from 2 `ref` property value and it doesn't affect the render of the component so if this value has changed the component doesn't need to be re-rendered.

frontend_app_file/src/container/File.jsx
```diff
@@ -167,7 +169,8 @@ export class File extends React.Component {
       newContent: newContentObject,
       editionAuthor: data.author.public_name,
       hasUpdated: prev.loggedUser.userId !== data.author.user_id,
-      timeline: addRevisionFromTLM(data, prev.timeline, prev.loggedUser.lang)
+      timeline: addRevisionFromTLM(data, prev.timeline, prev.loggedUser.lang),
+      isLastTimelineItemCurrentToken: data.client_token === getOrCreateSessionClientToken()
```
Skylsmoi, line 173: instead of calling the function every time, you can have a variable that store it in the component.
It wont change. I think a class property can be a good place

frontend_app_html-document/src/container/HtmlDocument.jsx
```diff
@@ -102,7 +104,8 @@ export class HtmlDocument extends React.Component {
       editionAuthor: data.author.public_name,
       hasUpdated: prev.loggedUser.userId !== data.author.user_id,
       rawContentBeforeEdit: data.content.raw_content,
-      timeline: addRevisionFromTLM(data, prev.timeline, prev.loggedUser.lang)
+      timeline: addRevisionFromTLM(data, prev.timeline, prev.loggedUser.lang),
+      isLastTimelineItemCurrentToken: data.client_token === getOrCreateSessionClientToken()
```
Skylsmoi, line 108: same comment as File.jsx

frontend_app_thread/src/container/Thread.jsx
```diff
@@ -120,7 +122,8 @@ export class Thread extends React.Component {
       newContent: { ...prev.content, ...data.content },
       editionAuthor: data.author.public_name,
       hasUpdated: prev.loggedUser.userId !== data.author.user_id,
-      timeline: addRevisionFromTLM(data, prev.timeline, this.state.loggedUser.lang)
+      timeline: addRevisionFromTLM(data, prev.timeline, this.state.loggedUser.lang),
+      isLastTimelineItemCurrentToken: data.client_token === getOrCreateSessionClientToken()
```
Skylsmoi, line 126: same comment as File and HtmlDoc

frontend_lib/src/component/Timeline/Timeline.jsx
```diff
@@ -31,28 +31,40 @@ export class Timeline extends React.Component {
...
-      this.timelineContainer.scrollTop + this.timelineContainer.clientHeight === this.timelineContainerScrollHeight ||
-      (props.isLastTimelineItemCurrentToken && props.newComment === '')
+      (window.innerWidth >= 1200 && (isNewContent || isScrollAtTheBottom || (props.isLastTimelineItemCurrentToken && props.newComment === ''))) ||
+      (window.innerWidth < 1200 && props.isLastTimelineItemCurrentToken && props.newComment === '' && props.timelineData[props.timelineData.length - 1].content_type === CONTENT_TYPE.COMMENT)
     ) {
-      this.timelineBottom.scrollIntoView({ behavior: 'instant' })
+      const behavior = isScrollAtTheBottom && props.isLastTimelineItemCurrentToken ? 'smooth' : 'instant'
+      this.timelineBottom.scrollIntoView({ behavior })
     }
   }
```
Skylsmoi: Your lines are too long.
Github displays around 150 columns which is already pretty big but here, your code can't even fit on github pages.
Usually, in javascript, IDE advice lines length at 120 columns. We don't have a strict rule on that but some of your lines are near 200 columns.
For example, the PEP8 line length coding rule in python is 80 columns.

It's always painful to have to scroll horizontally, especially on desktop (without trackpad).

Skylsmoi: Also your last test is still a bit too complicated.
You did well with the variables isNewContent and isScrollAtBottom, try to rename some more conditions

It'll also shorten the lines

Skylsmoi: for example, I see that `props.isLastTimelineItemCurrentToken && props.newComment === '')` is on both innerWidth >= 1200 and < 1200.
Couldn't you have an early return if this condition isn't met ?

frontend_lib/test/component/Timeline.spec.js
```diff
@@ -63,108 +68,195 @@ describe('<Timeline />', () => {
...
+        })
+        describe('a timeline item is added when the timeline scroll is not at the bottom', () => {
+          before(() => {
+            scrollIntoViewCallBack.resetHistory()
+            wrapper.instance().timelineContainer = {
+              scrollTop: scrollTop,
+              clientHeight: clientHeight
+            }
+            wrapper.instance().timelineContainerScrollHeight = scrollTop
+            wrapper.instance().scrollToBottom(props.timelineData)
```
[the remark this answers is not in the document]
metzgegu: The `before()` set up the right context and launch the test for the function `.scrollToBottom()`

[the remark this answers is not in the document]
Skylsmoi: still not clear
why does your context needs to call .scrollToBottom if your test is testing something when the timeline isn't supposed to be at the bottom

frontend_lib/src/component/Timeline/Timeline.jsx
metzgegu: We couldn't have an early return if this condition isn't met because with `innerWidth >= 120` it is in a OR condition

Skylsmoi: indeed

frontend_lib/test/component/Timeline.spec.js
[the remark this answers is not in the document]
metzgegu: Before calling `.scrollToBottom` the scroll isn't at the bottom, and after `.scrollToBottom()` was called we need to test if the auto-scroll was enabled because `.scrollToBottom()` has conditions in it (we are testing these conditions in fact).
But maybe I can put `wrapper.instance().scrollToBottom(props.timelineData)` in the `it()` to be more clear

gbonaspetti:
Tested on branch

frontend_lib/test/component/Timeline.spec.js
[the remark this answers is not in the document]
Skylsmoi: Ok that's because scrollToBottom is called by componentDidUpdate.

that's fine then, a bit unclear but consistent

Issue #3203, opened 2020-06-19
Bug: TLM is broken when network connection is lost or when we stop/start backend server


PhilippeAccorsi:
## Bug description & symptom

If connection is lost with backend and it found again, user not see anything in frontend but when it add new comment in app file for example, nothing append in timeline. Message is not visible (but POST request in devtools show that request is send correctly).

Another user with new connection can seen this comment directly (because his connection is open after connection restablish). 

## Diagnostic

*"unknown".*

## Expected

- [x] immediately show user message about connection loss if no TLM connection during page load
- [x] show the message when a connection loss for more than 4 seconds is detected after page load
- [x] reconnect automatically in the background

## Context and version information

- Tracim 3.0.0 beta

## How to reproduce

You need an instance with two users, this two users are connected in Tracim and look same file. Each user can add comment in timeline, and the TLM show message correctly for this two users.

The administrator of Tracim server restart Tracim.

The 2 users think that tracim work correctly (no message about this connection break).
At this point, TLM are broken.

What is going on now:
- one user add a new comment
- his comment is not visible in his timeline but also not visible in timeline of the second user
- the second user open a new tab to see this document. He can see the comment of the first user (if this first user add a new comment, the TLM show this new comment in this new tab. First user have write two comment and not see anything in his timeline).
- If the first user reload his tab, all comment are visible (new one also) and at this time, each new comment can be visible correctly.

## Impacted components

*"unknown"*

## Extra information

This bug impact all TLM, so everywhere tlm are used, Tracim not show information correctly before reload page.

### For backend:
see https://github.com/tracim/tracim/pull/3234
- Add a "stream-opened" event at stream opening instead of returning empty.
- Add a "keepalive" event that keepalive connection by sending message each 30 sec.

grignards:
The standard way of implementing reliability in this situation is to use a heartbeat: a special SSE is regularly sent by the backend and the frontend displays a message "Connection lost" when no SSE is received for a set duration (that we have to discuss).

Pushpin has a feature to implement such a heartbeat: https://pushpin.org/docs/advanced/#keep-alives.

PhilippeAccorsi:
@celiamgtt we need to show a message about connection lost (or something like that), design system exist for this kind of message ?  (I think we can create label `design system` now).

inkhey:
[...]

PR #3277, opened 2020-07-03
feat(frontend): handle TLM reconnection in case of error and not received keep-alives

grignards:
See #3203.

Display a message to the user if connection is lost for more than 4
seconds

*Please indicate here a summary of the pull request content. If a related issue exists, please indicate issue reference here*

## Checkpoints

**For code reviewers**

**For developers**

**For the quality team**


frontend/test/src/LiveMessageManager.spec.js
```diff
@@ -0,0 +1,73 @@
...
+    it('should set manager status to CLOSED', () => {
+      const { manager } = openedManager()
+      managers.push(manager)
+      manager.closeLiveMessageConnection()
+      expect(manager.status).to.be.equal(LIVE_MESSAGE_STATUS.CLOSED)
+    })
+  })
+
+  describe('the heartbeat timer', () => {
+    it('should restart connection when the heartbeat event is not received in time', async () => {
```
inkhey, line 59: Just a question, Do you expect heartbeat event at a regular time ?
If so, you should probably modify a bit https://github.com/tracim/tracim/blob/1e9d9c30fd4e7e2de62888b0b81d83de358cc005/backend/tracim_backend/views/core_api/user_controller.py#L626
with "interval"
see https://pushpin.org/docs/advanced/#keep-alives

grignards: Ah, yes you're right I saw this… I can also reset the frontend timer when a tlm is received :thinking:

frontend/src/component/FlashMessage.styl
```diff
@@ -38,6 +38,9 @@
         &__paragraph
           padding-right 20px
 
+#connection_error
```
Skylsmoi, line 41: avoid using html ids for styling, prefer class usage
The idea behind it is to keep specificity low and manageable

frontend/src/container/Tracim.jsx
```diff
@@ -247,6 +282,18 @@ export class Tracim extends React.Component {
...
+            id='connection_error'
+            flashMessage={
+              [{
+                message: props.t('Tracim has a connection problem, please wait or reload the page if the problem persists'),
+                type: 'danger'
+              }]
+            }
+            showCloseButton={false}
+            t={props.t}
+          />}
```
Skylsmoi, line 296: to have the closing bracket at the same column as the opening one while having standardjs ok with it is by using parenthesis like so
```suggestion
        {state.displayConnectionError && (
          <FlashMessage
            id='connection_error'
            flashMessage={
              [{
                message: props.t('Tracim has a connection problem, please wait or reload the page if the problem persists'),
                type: 'danger'
              }]
            }
            showCloseButton={false}
            t={props.t}
          />
        )}
```

frontend/src/reducer/tlmManager.js
```diff
@@ -1,9 +1,11 @@
-import { SET, TLM_MANAGER } from '../action-creator.sync.js'
+import { SET, TLM_MANAGER, TLM_MANAGER_STATUS } from '../action-creator.sync.js'
 
-export default function tlmManager (state = null, action) {
+export default function tlm (state = {}, action) {
```
Skylsmoi, line 3: If you rename the reducer, also rename its filename please

Skylsmoi: Also, now you have multiple properties, its better to add them to its default value
```jsx
const defaultTlm = {
  manager: null,
  status: LIVE_MESSAGE_STATUS.CLOSED
}
export default function tlm (state = defaultTlm, action) {
```

Issue #3222, opened 2020-06-23
Bug: apache config and uwsgi need improvement to make Tracim work correctly with Docker (with pushpin and multiple users)


PhilippeAccorsi:
## Bug description & symptom
Pushpin not working if ` On` is in Apache configuration.
Also when many users used Tracim, uWSGI bottleneck exist.

## Diagnostic

- This directive  `ProxyPreserveHost On` is just necessary for webdav. Just need to add this directive in webdav context
- uWSGI is configured like this by default in `tracim_web.ini`:
~~~
#workers = 1
#threads = 8
~~~

## Expected

- [x] Pushpin work correctly with docker image
- [x] improve uWSGI configuration in `tracim_web.ini` : use 4 workers and 4 threads by default
- [x] add ProxyPreserveHost for webdav location only
- [x] add Destination header fix to docker image with comment.
- [x] document apache and uwsgi config file when needed
- [x] don't change tracim_webdav and tracim_caldav uwsgi config.
- [x] update documentation according to change made (Apache, uWSGI)
- ~~[ ]  add fix for pushpin with antivirus ?~~ => most probably unneeded, if modifications are done they will be in #3235.

## Context and version information

- Pushpin come in v3.0.0
- Apache configuration is from oldest Tracim version

## How to reproduce

Pushpin:
- Try to use pushpin available in Tracim unstable image
uWSGI:
- see what is going on when many user try to send file in same Tracim instance

## Impacted components

- Docker images

PhilippeAccorsi:
After discuss with @lebouquetin , we can change uWSGI default configuration from 
~~~
#workers = 1
#threads = 8
~~~
to
~~~
workers = 8
threads = 8
~~~

PhilippeAccorsi:
Need also modification in documentation, exemple here about apache https://github.com/tracim/tracim/blob/develop/backend/doc/apache.md

inkhey:
After some research, we found that removing `ProxyPreserveHost On` broke webdav support for copy/move actions, in a similar but slighty different way than : https://github.com/tracim/tracim/issues/1765

```
curl --user 'username:password' -X COPY --header 'Destination:https//domainame/webdav/contentmanager' 'https://domainame/webdav/reader'
<!DOCTYPE HTML PUBLIC '-//W3C//DTD HTML 4.01//EN' 'http://www.w3.org/TR/html4/strict.dtd'>
<html><head>
[...]

PR #3285, opened 2020-07-03
adapt uwsgi and apache config

inkhey:
related to #3222 

## Checkpoints

**For code reviewers**

**For developers**

**For the quality team**

### DIFF
--- a/backend/doc/apache.md
+++ b/backend/doc/apache.md
@@ -30,23 +30,36 @@ Create a file named `/etc/apache2/sites-available/tracim.conf` containing:
             Dav On
         </Directory>
         SetEnv proxy-sendcl
-        ProxyPreserveHost On
 
-        ProxyPass /webdav http://127.0.0.1:3030/webdav
-        ProxyPassReverse /webdav http://127.0.0.1:3030/webdav
+        # Setting Destination header from https to http in proxy
+        # is needed for working move/copy in webdav
+        RequestHeader edit Destination ^https http early
 
+        # Proxying Webdav
+        <Location /webdav>
+            # Preserving host is needed for working move/copy in webdav
+            ProxyPreserveHost On
+            ProxyPass http://127.0.0.1:3030/webdav
+            ProxyPassReverse http://127.0.0.1:3030/webdav
+        </Location>
+
+        # Proxying Caldav
         ProxyPass /agenda http://127.0.0.1:5232/agenda
         ProxyPassReverse /agenda http://127.0.0.1:5232/agenda
 
+        # Proxying Frontend
         ProxyPass /ui http://127.0.0.1:6543/ui
         ProxyPassReverse /ui http://127.0.0.1:6543/ui
 
+        # Proxying Tracim Live Message to Pushpin
         ProxyPassMatch /api/users/(.*/live_messages)$ http://127.0.0.1:7999/api/users/$1
         ProxyPassReverse /api/users/(.*/live_messages)$ http://127.0.0.1:7999/api/users/$1
 
+        # Proxying Backend API
         ProxyPass /api http://127.0.0.1:6543/api
         ProxyPassReverse /api http://127.0.0.1:6543/api
 
+        # Proxying Plugin assets
         ProxyPass /custom_toolbox-assets http://127.0.0.1:6543/custom_toolbox-assets
         ProxyPassReverse /custom_toolbox-assets http://127.0.0.1:6543/custom_toolbox-assets
 
@@ -84,8 +97,8 @@ Create the file named `/etc/uwsgi/apps-available/tracim.ini` containing:
     virtualenv = [TRACIM_PATH]/backend/env
     http-socket = :6543
     socket-timeout = 360
-    #workers = 1
-    #threads = 8
+    #workers = 4
+    #threads = 4
     logto = /var/log/uwsgi/uwsgi_tracim.log
 
 Replace `[TRACIM_PATH]` by the path of your Tracim configuration file.
--- a/backend/doc/apache.md
+++ b/backend/doc/apache.md
@@ -31,12 +31,11 @@ Create a file named `/etc/apache2/sites-available/tracim.conf` containing:
         </Directory>
         SetEnv proxy-sendcl
 
-        # Setting Destination header from https to http in proxy
-        # is needed for working move/copy in webdav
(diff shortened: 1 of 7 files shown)


backend/doc/apache.md
```diff
@@ -30,23 +30,36 @@ Create a file named `/etc/apache2/sites-available/tracim.conf` containing:
             Dav On
         </Directory>
         SetEnv proxy-sendcl
```
inkhey, line 32: @PhilippeAccorsi @grignards A comment here may be nice, did you know why do we need this
https://httpd.apache.org/docs/2.4/en/mod/mod_proxy_http.html said  that this can be inefficient:

>  However, it may require the entire request to be buffered by the proxy, so it becomes very inefficient for large requests.

backend/doc/apache.md
```diff
@@ -30,23 +30,36 @@ Create a file named `/etc/apache2/sites-available/tracim.conf` containing:
             Dav On
```
inkhey, line 30: Comment may be nice here. https://httpd.apache.org/docs/2.4/en/mod/mod_dav.html. 
I do not understand why we do need mod dav and what it is doing.

Anyway if this is useful only for webdav, it may make sense to put this only in webdav location.

backend/doc/apache.md
```diff
@@ -30,23 +30,36 @@ Create a file named `/etc/apache2/sites-available/tracim.conf` containing:
...
         </Directory>
         SetEnv proxy-sendcl
-        ProxyPreserveHost On
 
-        ProxyPass /webdav http://127.0.0.1:3030/webdav
-        ProxyPassReverse /webdav http://127.0.0.1:3030/webdav
+        # Setting Destination header from https to http in proxy
+        # is needed for working move/copy in webdav
+        RequestHeader edit Destination ^https http early
```
inkhey, line 36: We should consider move this to /webdav location as this is the only known place where we do need this.

tools_docker/Debian_Uwsgi/apache2.conf.sample
```diff
@@ -26,20 +26,32 @@
...
-    #ProxyPassReverse /webdav http://127.0.0.1:3030/webdav
+    # Setting Destination header from https to http in proxy
+    # is needed for working move/copy in webdav
+    RequestHeader edit Destination ^https http early
 
+    # Proxying Webdav
+    <Location /webdav>
+        # Preserving host is needed for working move/copy in webdav
+        ProxyPreserveHost On
+        #ProxyPass http://127.0.0.1:3030/webdav
```
inkhey, line 38: Not sure about this strategy to enable/disable webdav using sed. May be better to remove all location but more complicated.

backend/doc/apache.md
```diff
@@ -84,8 +97,8 @@ Create the file named `/etc/uwsgi/apps-available/tracim.ini` containing:
     virtualenv = [TRACIM_PATH]/backend/env
     http-socket = :6543
     socket-timeout = 360
-    #workers = 1
-    #threads = 8
+    #workers = 4
+    #threads = 4
```
inkhey, line 101: These commented value are not uncommented with keeping the value, is it normal ?

backend/doc/apache.md
PhilippeAccorsi: I thinks its for Webdav with macos or something like that. I can search in the history of this file to find commit.

backend/doc/apache.md
PhilippeAccorsi: Apache.md documentation is come from tracim v1 I think. so some information need to be updated.

tools_docker/Debian_Uwsgi/apache2.conf.sample
PhilippeAccorsi: I have seen docker removing configuration on another project (jitsi meet) and its not a good idea for administration. You not understand what is available for the app.

If all parameter are visible but commented, you can understand what is possible to do. Maybe understand what is missing.

If we remove config if webdav is not started, if you activate webdav but is not working (because this line is not added in the file) you need to find what you need to do to use webdav with Tracim.

If config is visible, you can try to uncomment manually, or you can understand what is missing for Tracim.

backend/doc/apache.md
PhilippeAccorsi: I not understand what you says.

inkhey: This part of code is either : 
- untouched: stay uncommented
- modified with sed like  : ` sed -i "s|^#threads = .*|threads = 4|g" /etc/tracim/tracim_web.ini` 

which mean the default commented value is either overrided either commented.

tools_docker/Debian_Uwsgi/apache2.conf.sample
grignards: Another possibility would be to use additional config files for webdav/... and enable them through `a2enconf`. This has the advantage of being more robust than a sed based solution (as the script calling sed has to be updated whenever a change is made in the configuration it edits). But you end up with several configuration files to read (which is a matter of habits I think).

PhilippeAccorsi: I'm not sure is more simple (for basic user) to use multiple configuration files. Its better than use `sed` (like you explain) but not sure is more simple for support.

In the futur (we have discuss about that with @lebouquetin, but not validate choice), we probably won't continue to leave all configuration file (apache, uwsgi, development.ini) directly in docker volume. So when we do this modification we can improve usage of configuration files (like you purpose for apache).

backend/doc/apache.md
grignards: I opened an [issue](https://github.com/tracim/tracim/issues/3320) for this subject, we'll keep this directive until it is addressed.

backend/doc/apache.md
grignards: I removed this line and didn't see any effect on webdav operations (tried creating/moving/deleting files, folders and workspaces).
As gives unauthenticated access to the frontend's code (luckily read-only as the directory/files are RO for the apache's user) I would recommend to remove it.

backend/doc/apache.md
grignards: Done.

grignards:
Tested by building an image with this new configuration and enabled/disabled webdav/caldav. I didn't see any problem. Also tested webdav functionality.

Issue #3057, opened 2020-05-14
Feat: add pagination and event_type filter to user message endpoint


inkhey:
## Feature description and goals

For performance concerns, we need a pagination mechanism like in other endpoints. Currently (see PR https://github.com/tracim/tracim/pull/3022), we only allow filtering by read status (The notification pane can display either only mentions or all messages).

On endpoint `/users/{user_id}/messages`:
- [X] add pagination with `count` and `page_token` parameters
  The pagination will return a maximum of `count` messages, the page token will allow to go to a specific "page", when using the endpoint, `previous_page_token` and `next_page_token` will be returned, allowing frontend to go easily to next/previous page.
- [X] add `event_types` filter.
  This filter will take a comma separated list of event types and only return messages matching one of the given types. It is optional.
- [X] make sure if we always use pagination somewhere in Tracim, to find a way to have similar mechanism everywhere if possible.

## ~~Needs discussion before development~~

~~It is unclear how pagination should be done offset-free or not, and if offset-free, it needs clarification on what should be the rule for all offset-free paginations for later endpoints: https://github.com/tracim/tracim/issues/3057#issuecomment-660875567~~

## Implemented solution

- add `event_types` filter to `/messages` endpoint
- reverse order `/messages `endpoint result to get last message first
- use library `sqlakeyset` for pagination in `/messages` endpoint : put messages in `items` field and add field `next_page_token`, `has_next`, `has_previous`, `next_page_token`, `per_page`, `previous_page_token`. frontend should set page_token query parameter to value obtain either by `next_page_token` or `previous_page_token`

effective endpoint specification:
```

GET /api/users/1/messages?read_status=read&event_types=user.created%2Cuser.modified&count=1&page_token=%3Ei%3A2 HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Authorization: Basic YWRtaW5AYWRtaW4uYWRtaW46YWRtaW5AYWRtaW4uYWRtaW4=
Connection: keep-alive
Host: tracim.test
User-Agent: HTTPie/0.9.8

HTTP/1.1 200 OK
Connection: Keep-Alive
Content-Type: application/json
Date: Thu, 23 Jul 2020 13:38:04 GMT
Keep-Alive: timeout=5, max=100
Server: Apache
Transfer-Encoding: chunked

{
    "has_next": false,
    "has_previous": true,
    "items": [
        {
            "created": "2020-07-23T11:24:17Z",
            "event_id": 1,
[...]

Issue #2835, opened 2020-04-07
 Add endpoints to set messages (TLMs) as read/not read


Skylsmoi:
## Feature description

We need to be able to mark messages as (un)read when clicking on the "mark all as read" button (#3322) on the notification pane or on a notification.

Add these endpoints:
- [x] PUT `/api/users/<user_id>/messages/<event_id>/read` that will set the message's `read` field to the current UTC datetime
- [x] PUT `/api/users/<user_id>/messages/<event_id>/unread` that will set the message's `read` field to `None`/`null`
- [x] PUT `/api/users/<user_id>/messages/read` that will set all messages `read` field to the current UTC datetime if they not already have a read date.

## Implemented solution
add endpoints as expected:
- PUT `/api/users/<user_id>/messages/<event_id>/read`, 
- PUT `/api/users/<user_id>/messages/<event_id>/unread` 
- PUT `/api/users/<user_id>/messages/read` 

All theses endpoint  returning , returning 204 NO CONTENT if ok.
- When tuple user_id,event_id don't match with a valid, should return error 400, errorcode  1009 (MESSAGE_NOT_FOUND)
- Reread a content will reset the read datetime.
- Unread a readed content will delete the read datetime.
- Unread an unread content change nothing
---
_13-07-2020 - Description edition: Added task list and unread endpoint._
_17-07-2020 - Description edition: Added specification at message_id and event_id._
_20-07-2020 - Update with state of the PR: https://github.com/tracim/tracim/pull/3362 and use `event_id` instead of composite key `message_id` ._

lebouquetin:
> NB: the message_id is in fact event_id in database as the messages have a composite primary key (user_id, event_id).

I'm not very confident with this design. The API design let us understand that `message_id` is the primary key. If writing issues need this kind of explanation, what about a developer who wants to use our API without knowing how Tracim works?

inkhey:
> > NB: the message_id is in fact event_id in database as the messages have a composite primary key (user_id, event_id).
> 
> I'm not very confident with this design. The API design let us understand that `message_id` is the primary key. If writing issues need this kind of explanation, what about a developer who wants to use our API without knowing how Tracim works?

As there is always one message per user per event, i don't see any reason to name it message_id. Using event_id seems to me better.

PhilippeAccorsi:
api visible here http://localhost:7999/api/doc/

PR #3351, opened 2020-07-17
Feat/3057  pagination message endpoint

inkhey:
related to #3057 
require https://github.com/tracim/tracim/issues/2835
## Checkpoints

**For developers**

**For code reviewers**

**For testers**


backend/tracim_backend/lib/core/event.py
```diff
@@ -93,6 +100,21 @@ def _base_query(
...
+                    Event.entity_type == event_type.entity,
+                    Event.operation == event_type.operation,
+                    Event.entity_subtype == event_type.subtype,
+                )
+                event_type_filters.append(event_type_filter)
+
+            if len(event_type_filters) > 1:
+                query = query.filter(or_(*event_type_filters))
+            else:
+                query = query.filter(event_type_filters[0])
```
raphj, line 117: Would this work, to avoid a special case? Or does it add an overhead to call `or_` "uselessly"? 
```suggestion
            query = query.filter(or_(*event_type_filters))
```

I'm okay either way, though.

backend/tracim_backend/lib/utils/utils.py
```diff
@@ -492,3 +495,13 @@ def get_build_version(path: str) -> str:
     except StopIteration:
         # INFO - G.M - 2020-01-13 - return the 10 first letter of current commit hash
         return repo.head.object.hexsha[:10]
+
+
+def validate_page_token(page_token: str):
+    # INFO - G.M - 2020-07-23 - Explicitly catch error for unvalid bookmark
+    # as unserialize_bookmark method of sqlakeyset doesn't return explicit exception
+    # see https://github.com/djrobstep/sqlakeyset/issues/34
```
raphj, line 503:
```suggestion
    # INFO - G.M - 2020-07-23 - Because they lack an explicit message, we catch exceptions from
    # the unserialize_bookmark method of sqlakeyset and re-raise them with an explicit message.
    # See https://github.com/djrobstep/sqlakeyset/issues/34
```

backend/tracim_backend/models/event.py
```diff
@@ -19,6 +20,32 @@
...
+class EventType(object):
+    """
+    Helper to convert event_type string as real used parameters in
+    database
+    """
+
+    def __init__(self, name: str):
+        event_type_data = name.split(".")
+        if not len(event_type_data) in [2, 3]:
+            raise ValidationError("event_type should have 2 to 3 part")
```
raphj, line 32:
```suggestion
            raise ValidationError("event_type should have 2 to 3 parts")
```

backend/tracim_backend/models/context_models.py
```diff
@@ -581,6 +584,24 @@ def __init__(self, label: str, raw_content: str) -> None:
         self.raw_content = raw_content
 
 
+class LiveMessageQuery(object):
+    """
+    Folder Content update model
```
grignards, line 589: This docstring seems to be a copy/paste?

backend/tracim_backend/models/context_models.py
```diff
@@ -1526,3 +1547,13 @@ def filename(self) -> str:
         :return: complete filename with both label and file extension part
         """
         return core_convert_file_name_to_display(self.revision.file_name)
+
+
+class PaginatedObject(object):
+    def __init__(self, page: Page):
```
grignards, line 1553: Don't forget to add the typing for the returned value `-> None`.

backend/tracim_backend/models/event.py
```diff
@@ -19,6 +20,32 @@
 from tracim_backend.models.meta import DeclarativeBase
 
 
+class EventType(object):
+    """
+    Helper to convert event_type string as real used parameters in
+    database
+    """
+
+    def __init__(self, name: str):
```
grignards, line 29: Same here.

backend/tracim_backend/views/core_api/schemas.py
```diff
@@ -1433,12 +1449,35 @@ class LiveMessageSchema(marshmallow.Schema):
...
+    )
+    page_token = marshmallow.fields.String(
+        description="token of the page wanted, if not provided get first" "elements",
+        validate=page_token_validator,
     )
+    read_status = StrippedString(missing=ReadStatus.ALL.value, validator=OneOf(ReadStatus.values()))
+    event_types = EventTypeListField()
+
+    @post_load
+    def live_message_query(self, data: typing.Dict[str, typing.Any]) -> object:
```
grignards, line 1479: The type should be `LiveMessageQuery` isn't it?

backend/tracim_backend/lib/core/event.py
inkhey: Unfortunately this doesn't work.

backend/tracim_backend/models/context_models.py
inkhey: Right, i will solve this.
```suggestion
    Live Message query model
```

backend/tracim_backend/models/context_models.py
inkhey, line 1553:
```suggestion
    def __init__(self, page: Page) -> None:
```

backend/tracim_backend/models/event.py
inkhey, line 29:
```suggestion
    def __init__(self, name: str) -> None:
```

backend/tracim_backend/views/core_api/schemas.py
inkhey: Yes, code is not always as strict as it should for post_load code as this add many redundancy between object and  schema related. 

```suggestion
    def live_message_query(self, data: typing.Dict[str, typing.Any]) -> LiveMessageQuery:
```

backend/tracim_backend/models/context_models.py
inkhey: done.

backend/tracim_backend/models/event.py
inkhey: done.

backend/tracim_backend/models/event.py
```diff
@@ -19,6 +20,32 @@
...
+        event_type_data = name.split(".")
+        if not len(event_type_data) in [2, 3]:
+            raise ValidationError("event_type should have 2 to 3 parts")
+        if len(event_type_data) == 2:
+            event_type_data.append(None)
+        entity = event_type_data[0]
+        operation = event_type_data[1]
+        subtype = event_type_data[2]
+
+        if entity not in EntityType.values():
```
grignards, line 39: Only trying to create an EntityType with the value and catch/reraise the exception that is raised if the string does not represent an entity type would be simpler (and more pythonic :smiley: ).

backend/tracim_backend/models/event.py
```diff
@@ -19,6 +20,32 @@
 from tracim_backend.models.meta import DeclarativeBase
 
 
+class EventType(object):
```
grignards, line 23: The naming is confusing: this is not an event type but a utility class.
I would also rather have a static method to convert from event_type -> database parameters:
```python
class EventTypeDatabaseParameters:
    def __init__(self, entity: EntityType, operation: OperationType, subtype: typing.Optional[str]) -> None:
        self.entity = entity
        self.operation = operation
        self.subtype = subtype
   
    @staticmethod
    def from_event_type(event_type: str) -> "EventTypeDatabaseParameters":
        ...
        return EventTypeDatabaseParameters(entity, operation, subtype)
```
This makes easier for a reader to see which attributes are set on an instance as the only code in `__init__` is setting them.

backend/tracim_backend/models/event.py
inkhey: done.

backend/tracim_backend/models/event.py
inkhey: done.

Issue #3343, opened 2020-07-16
Bug: postgres database migration from 2.7.6 to 3.0.0_rc1 not working


PhilippeAccorsi:
## Bug description & symptom

I have tried to install tracim 3.0.0_rci on existing database (tracim 2.7.6, but its a tracimv1 database from beggining)
I use docker and I seen this error

~~~
Starting 0011_algoo
Attaching to 0011_algoo
0011_algoo                   | Traceback (most recent call last):
0011_algoo                   |   File "/usr/local/lib/python3.5/dist-packages/sqlalchemy/engine/base.py", line 1244, in _execute_context
0011_algoo                   |     cursor, statement, parameters, context
0011_algoo                   |   File "/usr/local/lib/python3.5/dist-packages/sqlalchemy/engine/default.py", line 552, in do_execute
0011_algoo                   |     cursor.execute(statement, parameters)
0011_algoo                   | psycopg2.errors.UndefinedObject: constraint "fk_content_revisions_content_id_content" of relation "content_revisions" does not exist
0011_algoo                   | 
0011_algoo                   | 
0011_algoo                   | The above exception was the direct cause of the following exception:
0011_algoo                   | 
0011_algoo                   | Traceback (most recent call last):
0011_algoo                   |   File "/usr/local/bin/alembic", line 8, in <module>
0011_algoo                   |     sys.exit(main())
0011_algoo                   |   File "/usr/local/lib/python3.5/dist-packages/alembic/config.py", line 503, in main
0011_algoo                   |     CommandLine(prog=prog).main(argv=argv)
0011_algoo                   |   File "/usr/local/lib/python3.5/dist-packages/alembic/config.py", line 497, in main
0011_algoo                   |     self.run_cmd(cfg, options)
0011_algoo                   |   File "/usr/local/lib/python3.5/dist-packages/alembic/config.py", line 480, in run_cmd
0011_algoo                   |     **dict((k, getattr(options, k, None)) for k in kwarg)
0011_algoo                   |   File "/usr/local/lib/python3.5/dist-packages/alembic/command.py", line 254, in upgrade
0011_algoo                   |     script.run_env()
0011_algoo                   |   File "/usr/local/lib/python3.5/dist-packages/alembic/script/base.py", line 427, in run_env
0011_algoo                   |     util.load_python_file(self.dir, 'env.py')
0011_algoo                   |   File "/usr/local/lib/python3.5/dist-packages/alembic/util/pyfiles.py", line 81, in load_python_file
[...]

PR #3352, opened 2020-07-17
feat(backend): add db update-naming-conventions command to tracimcli

grignards:
See #3343.

## Checkpoints

**For developers**

- [X] Automated tests are irrelevant as this is a command for selected databases coming from v1 version.

**For code reviewers**

**For testers**

### DIFF
--- a/backend/tracim_backend/command/database.py
+++ b/backend/tracim_backend/command/database.py
@@ -156,7 +156,6 @@ class UpdateNamingConventionsV1ToV2Command(AppContextCommand):
         app_config.configure_filedepot()
         engine = get_engine(app_config)
         inspector = reflection.Inspector.from_engine(engine)
-        v1_index_convention = re.compile(r"idx__(\w+)")
         v1_unique_convention = re.compile(r"uk__(\w+)__(\w+)")
         v1_foreign_key_convention = re.compile(r"fk__(\w+)__(\w+)__(\w+)")
         v1_primary_key_convention = re.compile(r"pk__(\w+)")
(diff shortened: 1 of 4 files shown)


backend/tracim_backend/command/database.py
```diff
@@ -137,3 +139,75 @@ def take_action(self, parsed_args: argparse.Namespace) -> None:
...
+        "pk": "pk__%(table_name)s",  # Primary keys
+    }
+
+    def get_description(self) -> str:
+        return "Update database naming conventions from V1 database to V2"
+
+    def take_action(self, parsed_args: argparse.Namespace) -> None:
+        super(UpdateNamingConventionsV1ToV2Command, self).take_action(parsed_args)
+        config_uri = parsed_args.config_file
+        # setup_logging(config_uri)
```
inkhey, line 160: remove this commented line.

lebouquetin:
@grignards @inkhey one small remark about the naming convention you use in the database model.

- Historically, there were names like `fk__content_revisions__owner_id"`
- now there are names like `fk_revision_read_status_revision_id_content_revisions`

It's almost unreadable with no explicit separator, except if you already know all tables and columns. I'd prefer you to use naming convention with explicit separators between field names.

Do you agree?

grignards:
> Do you agree?

In principle yes, I would just be curious to know why those conventions changed as there could be a valid reason.
Also there is a limitation in sqlalchemy when referring multiple columns (with `%(column_0_N_name)s`: either you can have no separator or a `_`, but you cannot tell it to put `__` to separate column names.

inkhey:
> @grignards @inkhey one small remark about the naming convention you use in the database model.
> 
>     * Historically, there were names like `fk__content_revisions__owner_id"`
> 
>     * now there are names like `fk_revision_read_status_revision_id_content_revisions`
> 
> 
> It's almost unreadable with no explicit separator, except if you already know all tables and columns. I'd prefer you to use naming convention with explicit separators between field names.
> 
> Do you agree?

I agree the previous convention is more readable. I do think we should find a way to make thing clear later, but first we should have fixed same rule everywhere. It will make migration way more easier (it may be tricky for sqlite and mariadb though)

inkhey:
> > Do you agree?
> 
> In principle yes, I would just be curious to know why those conventions changed as there could be a valid reason.
> Also there is a limitation in sqlalchemy when referring multiple columns (with `%(column_0_N_name)s`: either you can have no separator or a `_`, but you cannot tell it to put `__` to separate column names.

I don't think there is a good reason. I do not take much attention to database convention when beginning work on tracim2, and i do just don't reuse old tracim v1 metadata config.

inkhey:
verified on branch with old database dump.

PR #3362, opened 2020-07-20
add endpoints to set message as read/unread

inkhey:
related to #2835 

## Checkpoints

**For developers**

**For code reviewers**

**For testers**


backend/tracim_backend/lib/core/event.py
```diff
@@ -71,20 +74,61 @@ def __init__(
...
+        try:
+            return self._base_query(user_id=user_id, event_id=event_id).one()
+        except NoResultFound as exc:
+            raise MessageDoesNotExist(
+                'Message for user {} with event id "{}" not found in database'.format(
+                    user_id, event_id
+                )
+            ) from exc
+
+    def read_message_of_user(self, user_id, event_id) -> Message:
```
grignards, line 108: With the method starting by `read`, I understand it reads the message with a "file read" semantic. Wouldn't `mark_message_of_user_as_read` be more explicit?
And given the order message_of_user, I would expect `(event_id, user_id)` for the order of arguments (so that it is the same as in the name)… But you have a `read_all(user_id)` method… Then use `mark_user_message_as_read(user_id, event_id)`?

backend/tracim_backend/lib/core/event.py
```diff
@@ -71,20 +74,61 @@ def __init__(
...
+        message = self.get_one_message(user_id, event_id)
+        message.read = datetime.utcnow()
+        self._session.add(message)
+        self._session.flush()
+        return message
+
+    def unread_message_of_user(self, user_id, event_id) -> Message:
```
grignards, line 115: Same comment as the last one.

backend/tracim_backend/tests/functional/test_message.py
```diff
@@ -20,9 +20,10 @@ def create_events_and_messages(session) -> tracim_event.Event:
             fields={"example": "hello"},
         )
         session.add(event)
-        messages.append(
-            tracim_event.Message(event=event, receiver_id=1, read=datetime.datetime.utcnow())
-        )
+        read_datetime = datetime.datetime.utcnow()
+        if unread:
+            read_datetime = None
```
grignards, line 25: Not really relevant here as it is a test method, but in general it is easier and more powerful to give `read_datetime` as argument to the method instead of only a boolean.

backend/tracim_backend/views/core_api/user_controller.py
```diff
@@ -603,6 +605,48 @@ def get_user_messages(
...
+    def read_all_messages(self, context, request: TracimRequest, hapic_data: HapicData) -> None:
+        """
+        Read all unread message for user
+        """
+        app_config = request.registry.settings["CFG"]  # type: CFG
+        event_api = EventApi(request.current_user, request.dbsession, app_config)
+        event_api.read_all_messages_of_user(request.candidate_user.user_id)
+
+    @hapic.with_api_doc(tags=[SWAGGER_TAG__USER_EVENT_ENDPOINTS])
+    @hapic.handle_exception(MessageDoesNotExist, http_code=HTTPStatus.BAD_REQUEST)
```
grignards, line 621: Is this case really a bad request? or only a not found?

backend/tracim_backend/views/core_api/user_controller.py
```diff
@@ -603,6 +605,48 @@ def get_user_messages(
...
+        Read one message
+        """
+        app_config = request.registry.settings["CFG"]  # type: CFG
+        event_api = EventApi(request.current_user, request.dbsession, app_config)
+        event_api.read_message_of_user(
+            user_id=request.candidate_user.user_id, event_id=hapic_data.path.event_id,
+        )
+
+    @hapic.with_api_doc(tags=[SWAGGER_TAG__USER_EVENT_ENDPOINTS])
+    @hapic.handle_exception(MessageDoesNotExist, http_code=HTTPStatus.BAD_REQUEST)
```
grignards, line 636: Same comment as last one.

backend/tracim_backend/views/core_api/user_controller.py
inkhey: According to current usage, yes, we do use bad request for other case where parameter given doesn't match with existing elements. The error_code explain the exact reason
404 not found is used to mean "the endpoint doesn't exist".

backend/tracim_backend/lib/core/event.py
inkhey: Fixed, better ?
I keep a same order everywhere with event_id, user_id which seems to me more clear.

backend/tracim_backend/lib/core/event.py
inkhey: fixed

inkhey:
api change only and no special stuff, not need for quality test here.

Issue #3349, opened 2020-07-17
Feat: Produce TLMs when a mention is created


raphj:
## Feature description and goals

This issue is a sub-task of #2838. `mention.created` events are produced when new mentions are found (implemented in #3347 and #3348).

TLMs must be produced for these events. Recipients of these TLMs depend on the mentioned name:
  - `@username` : send a TLM to the user matching the given username
  - `@all`: send TLMS to all workspace users

## Prerequisites

- probably #3347 (feel free to remove this if it's not the case)

gbonaspetti:
Tested on branch.

PhilippeAccorsi:
Verified.

Issue #3347, opened 2020-07-17
Feat: Generate events for mentions in HTML documents and comments


raphj:
## Feature description and goals

In order to generate TLMs for mentions (meta: #2838), we need to be able to generate events for mentions in:
 - [x] HTML documents
 - [x] comments

Those events, of type `mention.created`, will have a `mention` field (with the found mention and its id in the content), `workspace`, `content` and `author` fields:

```json
"mention": {
  "recipient": "all",
  "id": "<uuid>"
}
```

Those events will be created for new mentions when creating and modifying the content. This will be done by parsing the `description` field of contents as it is used to store the html documents and comments.

- [x] new `EntityType.MENTION`, create the related migration script.
- [x] add `parent_content_type` in `CommentSchema` and ensure it is present. The frontend needs it to link back to the content in the notification panel.

##  Discussed

- How to do the parsing of HTML: we'll use [BeautifulSoup](https://www.crummy.com/software/BeautifulSoup/) which is already a backend dependency.
- Agree to do it in the backend (see discussion in the meta-issue): The frontend will do a first parsing adding html tags and the backend will search this tags.
- Be sure to address this comment: https://github.com/tracim/tracim/issues/3347#issuecomment-660313273 (and edit the issue accordingly)

## Extra information

- Keep #3348 in mind when working on this issue.
- We plan to handle the translation of the group mentions: #3350

----
_20-07-2020 - Description edition: linked this issue to #3347 discussion_

lebouquetin:
the mentions are not related to html content or comment but about content description. It should be generic otherwise all new contents will require to upgrade the mention process.

Other point: I think if a user mention another one in a comment, I'm not sure there should be two TLM... otherwise when showing the full history, there will be two events:

- one being the comment `Hi @damien, could you read the new revision of the documentation?` linking to the comment?
- one being the mention `Hi @damien, could you read the new revision of the documentation?` linking to the comment?

I would add a mention property to existing TLM (property could be structured), no? No backend developer talked to you about this issue?

grignards:
> the mentions are not related to html content or comment but about content description. It should be generic otherwise all new contents will require to upgrade the mention process.
> 
> Other point: I think if a user mention another one in a comment, I'm not sure there should be two TLM... otherwise when showing the full history, there will be two events:
> 
>     * one being the comment `Hi @damien, could you read the new revision of the documentation?` linking to the comment?
> 
[...]

PR #3400, opened 2020-07-31
Feature/3347  generate mentions

grignards:
See #3347.

## Checkpoints

**For developers**

**For code reviewers**

**For testers**

- [ ] Manual, quality tests have been done: no quality test for this PR as it only implement part of a user functionality. Will be tested with the notification panel PR #3376.

**Manual test method**

Manual tests of this PR must be done by viewing the `events` table contents as generating the messages is part of #3349. Inserting a mention can be done with:

```html
<span id="mention-12345">@TheAdmin</span>
```

Viewing table contents with `sqlite3`:

```sql
select * from events where entity_type="MENTION"
```

### DIFF
--- a/backend/tracim_backend/lib/core/mention.py
+++ b/backend/tracim_backend/lib/core/mention.py
@@ -1,3 +1,4 @@
+import abc
 import typing
 
 from bs4 import BeautifulSoup
@@ -5,13 +6,16 @@ from bs4 import Tag
 from pluggy import PluginManager
 
 from tracim_backend.config import CFG
+from tracim_backend.app_models.contents import COMMENT_TYPE
 from tracim_backend.lib.core.content import ContentApi
 from tracim_backend.lib.core.event import BaseLiveMessageBuilder
 from tracim_backend.lib.core.event import EventApi
 from tracim_backend.lib.core.plugins import hookimpl
 from tracim_backend.lib.core.workspace import WorkspaceApi
+from tracim_backend.lib.utils.logger import logger
 from tracim_backend.lib.utils.request import TracimContext
 from tracim_backend.models.data import Content
+from tracim_backend.models.data import ContentRevisionRO
 from tracim_backend.models.event import EntityType
 from tracim_backend.models.event import OperationType
 from tracim_backend.models.event import Event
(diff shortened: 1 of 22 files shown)


backend/tracim_backend/lib/core/mention.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def __init__(self, recipient: str, id_: str) -> None:
+        self.recipient = recipient
+        self.id = id_
+
+    def __eq__(self, other) -> bool:
+        return (
+            isinstance(other, Mention) and other.recipient == self.recipient and other.id == self.id
+        )
+
```
raphj, line 32: It seems this could be simplified by using named tuples, or dataclasses

- https://docs.python.org/3.8/library/typing.html#typing.NamedTuple
- https://docs.python.org/3.8/library/collections.html#collections.namedtuple
- https://docs.python.org/3/library/dataclasses.html

With these, you won't have to write these `init` and `_eq__` method

(Compatibility with Python 3.5 or 3.6 might be an issue though)

backend/tracim_backend/lib/core/mention.py
```diff
@@ -0,0 +1,136 @@
...
+    def __repr__(self) -> str:
+        return "<Mention(recipient={}, id={})>".format(self.recipient, self.id)
+
+    def __hash__(self) -> int:
+        return hash(self.id)
+
+
+class MentionBuilder:
```
lebouquetin, line 40: We will have in a near future mentions extracted from other content than HTML. If the builder is dedicated to HTML, I suggest to name it like this.

By the way, I suggest to prepare the design for supporting several builders (at least: markdown and html): 

- you know exactly how the backend is design for extracting mentions
- you know that other builders will be required (I think you took into account this evolution)
- you are the best player for designing a "markdown-ready" design.

Otherwise, it will require refactoring in a near future, which does not seem to be a good idea.

backend/tracim_backend/lib/core/mention.py
grignards: Yes, dataclasses are not possible with Python 3.5 (there is a backport for 3.6). Could use named tuple it just doesn't play well with typing (I do not think it is possible to type the arguments to namedtuple, at least for 3.5: https://docs.python.org/3/library/typing.html#typing.NamedTuple).

backend/tracim_backend/lib/core/mention.py
grignards: In my mind `MentionBuilder` is aimed to be generic, with specific formats handled in separate classes. I hesitated to implement this directly as we currently only support html, but I can do it now.

backend/tracim_backend/lib/core/mention.py
raphj: I am also for dropping Python 3.5, I think this is holding us back in this kind of situation. Python 3.7 can still be installed on Stretch using the backports.

This discussion can be resolved.

lebouquetin: > Note: I would be in favor of dropping 3.5 support, but this is a general discussion as this would likely lead to also drop stretch support…

This is a product strategic decision, not a "ease of development decision" (ie to be discussed more largely)

raphj: Sure, I think this is what @grignards meant by saying "general discussion".

backend/tracim_backend/lib/core/mention.py
grignards: Changes done.

Issue #3415, opened 2020-08-06
Bug: TLMs aren't published when modifying a big html-document


grignards:
## Description and expectations

When modifying an html-document containing several images (amounting to around 1.7Mbytes) the corresponding `content.modified` TLMs are not published.

### How to reproduce

Create an html-document with several embedded images. Modify it (for example its status).

### Version information

- Tracim version: 3.0.0rc4
- Web browser: any

## Diagnostic

Publishing to pushpin is done with a HTTP request which has a maximum body size.

## Solution

Publishing through 0mq (PUB socket) doesn't show the problem.

PhilippeAccorsi:
Tested on preprod.

PR #3416, opened 2020-08-06
Fix/3415  publish tlm with 0mq pushpin

grignards:
See #3415.

## Checkpoints

**For developers**

**For code reviewers**

**For testers**

### DIFF
--- a/backend/doc/live_message_setup.md
+++ b/backend/doc/live_message_setup.md
@@ -17,7 +17,7 @@ you need to set tracim config like this to use default pushpin port:
 
 ~~~ini
 basic_setup.website_base_url = http://localhost:7999
-live_messages.control_uri = http://localhost:5561
+live_messages.control_zmq_uri = tcp://localhost:5563
 ~~~
 
 :warning:  Temporary, you also need to ensure "jobs.processing_mode" parameter is unset or set to "sync".
--- a/backend/tracim_backend/tests/functional/test_live_messages.py
+++ b/backend/tracim_backend/tests/functional/test_live_messages.py
@@ -29,9 +29,7 @@ def big_html_document(workspace_api_factory, content_api_factory, session) -> Co
         do_notify=False,
     )
     with new_revision(session=session, tm=transaction.manager, content=html_document):
-        content_api.update_content(
-            html_document, new_content=description, new_label="Big document"
-        )
+        content_api.update_content(html_document, new_content=description, new_label="Big document")
         content_api.save(html_document)
     transaction.commit()
     return html_document
@@ -64,9 +62,7 @@ class TestLivesMessages(object):
             "Basic",
             ("admin@admin.admin", "admin@admin.admin"),
         )
-        res = web_testapp.get(
-            "/api/users/{}/live_messages".format(admin_user.user_id), status=400,
-        )
+        res = web_testapp.get("/api/users/{}/live_messages".format(admin_user.user_id), status=400,)
         assert res.json_body
         assert "code" in res.json_body
         assert res.json_body["code"] == ErrorCode.GENERIC_SCHEMA_VALIDATION_ERROR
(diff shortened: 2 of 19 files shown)


backend/doc/live_message_setup.md
```diff
@@ -17,7 +17,7 @@ you need to set tracim config like this to use default pushpin port:
 
 ~~~ini
 basic_setup.website_base_url = http://localhost:7999
-live_messages.control_uri = http://localhost:5561
+live_messages.control_zmq_uri = tcp://localhost:5563
```
raphj, line 20: Would it be possible to use a UNIX socket (maybe more efficient than local tcp?)
Probably far from being a bottleneck though.

backend/tracim_backend/tests/functional/test_live_messages.py
```diff
@@ -1,20 +1,51 @@
...
 from tracim_backend.tests.fixtures import *  # noqa: F403,F40
 
 
+@pytest.fixture
+def big_html_document(workspace_api_factory, content_api_factory, session) -> Content:
+    workspace_api = workspace_api_factory.get()
+    workspace = workspace_api.create_workspace(label="Foobar")
+    content_api = content_api_factory.get()
+
+    description = "".join(random.choice(string.ascii_letters) for _ in range(2000000))
```
raphj, line 23: I don't think we should have randomness in tests. Is it sufficient to generate a zeroed string, or provide a seed so we are sure the test is always the same / deterministic? Not that it should impact anything here, but still.

raphj:
Don't forget to document the solution in the issue :-)

backend/doc/live_message_setup.md
grignards: maybe, but this involves modifying pushpin's configuration (as it uses tcp by default) so I propose to leave this value for now.

backend/tracim_backend/tests/functional/test_live_messages.py
grignards: You're right, I'll change this to a fixed content.

backend/doc/live_message_setup.md
raphj: :+1:

backend/tracim_backend/tests/functional/test_live_messages.py
lebouquetin: I suggest to also put UFT-8 chars...

PhilippeAccorsi:
A small fix is also added in this MR about missing `apt update` before installing pushpin in docker images.

backend/tracim_backend/tests/functional/test_live_messages.py
raphj: After discussion, we are putting this for future work to ease the release of Tracim 3.0.0. Unexpectedly, putting Japanese characters in there makes the test ruuuuunnnn veeerrrryyy sloooowly.

backend/tracim_backend/tests/functional/test_live_messages.py
```diff
@@ -18,7 +18,12 @@ def big_html_document(workspace_api_factory, content_api_factory, session) -> Co
     workspace = workspace_api.create_workspace(label="Foobar")
     content_api = content_api_factory.get()
 
-    description = 2000000 * "a"
+    # NOTE: MySQL and MariaDB have a maximum of 65536 bytes for description,
+    # so the size is chosen accordingly
```
raphj, line 22: Won't this effectively break Tracim 3.0 on MySQL / MariaDB as soon as someone writes a document bigger than 64 KiB?

Should we prevent people using MySQL from migrating to Tracim 3.0 until we sort this out?

raphj: oh, by the way, you should put `2020-08-07 - SG` - in this comment.

lebouquetin: > Should we prevent people using MySQL from migrating to Tracim 3.0 until we sort this out?

The problem is already there.

Issue #3322, opened 2020-07-10
Feat: Allow marking all notification as read


gbonaspetti:
## Feature description

It should be possible to mark all notifications as read.

- [x] add a button `Mark all as read`
  - [x] on click: all the notifications are updated with the read status and it's design changes
  - [x] on click: the header's button should also be updated

## Translations

| English | French | Portuguese |
|--|--|--|
| Mark all as read | Tout marquer comme lu | Marcar tudo como lido |

## Prerequisites

#2840 
#3323 

## Interface

![allAsRead](https://user-images.githubusercontent.com/35422692/87935620-b55c1180-ca91-11ea-83c4-01b9753d3d78.jpg)

Envelope icon's size: `12px`
Envelope icon: `fa-envelope-open-o`

----
_13-07-2020 - Description edition: Added translations._
_16-07-2020 - Description edition: Added interface and its items._
_20-07-2020 - Description edition: Added sub-items at task list and update interface._

gbonaspetti:
This issue was moved to In Progress because some interface development is in progress, but the label remains because it's yet `blocked by frontend` (see prerequisites). I can start the development, just not finish it.

PhilippeAccorsi:
Verified.

PR #3427, opened 2020-08-11
Feature/3322  add mark all notifications as read

gbonaspetti:
#3322 

## Checkpoints

**For developers**

**For code reviewers**

**For testers**


functionnal_tests/cypress/integration/notification_wall/mark_all_as_read_spec.js
```diff
@@ -0,0 +1,27 @@
...
+describe('Notification Wall', () => {
+  before(function () {
+    cy.resetDB()
+    cy.setupBaseDB()
+  })
+
+  beforeEach(function () {
+    cy.loginAs('users')
+    cy.visitPage({ pageName: PAGES.HOME })
+    this.skip() // FIXME - GB - 2020-08-10 - To be activate when the notification wall (https://github.com/tracim/tracim/issues/2840) and button's header (https://github.com/tracim/tracim/issues/3323) were merge
```
raphj, line 12:
```suggestion
    this.skip() // FIXME - GB - 2020-08-10 - To be enabled when the notification wall (https://github.com/tracim/tracim/issues/2840) and button's header (https://github.com/tracim/tracim/issues/3323) are merged
```

functionnal_tests/cypress/integration/notification_wall/translation_spec.js
```diff
@@ -0,0 +1,21 @@
+import { PAGES } from '../../support/urls_commands'
+
+describe('Notification Wall', () => {
+  before(function () {
+    cy.resetDB()
+    cy.setupBaseDB()
+    cy.loginAs('users')
+    cy.visitPage({ pageName: PAGES.HOME })
+    this.skip() // FIXME - GB - 2020-08-10 - To be activate when the notification wall (https://github.com/tracim/tracim/issues/2840) and button's header (https://github.com/tracim/tracim/issues/3323) were merge
```
raphj, line 9:
```suggestion
    this.skip() // FIXME - GB - 2020-08-10 - To be enabled when the notification wall (https://github.com/tracim/tracim/issues/2840) and button's header (https://github.com/tracim/tracim/issues/3323) are merged
```
Maybe add a checkbox in these issues to enable these tests?

frontend/src/container/NotificationWall.jsx
```diff
@@ -161,6 +162,20 @@ export class NotificationWall extends React.Component {
     return { icon, text, url }
   }
 
+  handleClickMarkAllAsRead = async () => {
+    const { props } = this
+
+    const fetchAllPutNotificationAsRead = await props.dispatch(putAllNotificationAsRead(props.user.userId))
+    switch (fetchAllPutNotificationAsRead.status) {
+      case 204: {
```
raphj, line 170: Braces are unnecessary here, but harmless and I don't mind them.

frontend/src/container/NotificationWall.jsx
```diff
@@ -161,6 +162,20 @@ export class NotificationWall extends React.Component {
...
+        props.notificationPage.list.forEach(notification => props.dispatch(updateNotification({ ...notification, read: true })))
```
raphj, line 171: I'd find this more readable:
```suggestion
        for (const notification of props.notificationPage.list) {
            props.dispatch(updateNotification({ ...notification, read: true }))
        }
```
I think we really should avoid forEach when `for…of` works. I don't see any advantage of using forEach in these cases (but I know this is what we currently do in the frontend). `forEach` is no more functional than `for` and was great when `for..of` was still not supported by browsers.

Invoking @Skylsmoi who may not agree with me here.

functionnal_tests/cypress/integration/notification_wall/translation_spec.js
gbonaspetti: The idea is enable this tests in this PR, I put the comment just to explain to the reviewers. But the skip it temporary, it will not be merge like this.

frontend/src/container/NotificationWall.jsx
gbonaspetti: For me `forEach` is more readable, but I can wait to see @Skylsmoi like a casting vote.

raphj: If you find it more readable, keep it as is ;-)

functionnal_tests/cypress/integration/notification_wall/translation_spec.js
raphj: ah, ok :-)
Let's use this conversation as a guard against a merge without this modification then.

frontend/src/container/NotificationWall.jsx
gbonaspetti: Changed

frontend_lib/src/component/PopinFixed/PopinFixed.styl
```diff
@@ -18,12 +18,14 @@
     align-items center
     padding 15px 10px
     height 62px
+    & > *
```
raphj, line 21: Why a star? Would it be possible to use a specific selector?

gbonaspetti: Because I want that wall elements have this margin as default.

raphj: All right. I'm not sure using a star is the best way to do this, instead of specifying the specific tags that we know will be in the header (which is something we control), but I may be missing something here so feel free to resolve the conversation.

frontend/src/container/NotificationWall.jsx
```diff
@@ -159,6 +160,19 @@ export class NotificationWall extends React.Component {
     return { icon, text, url }
   }
 
+  handleClickMarkAllAsRead = async () => {
+    const { props } = this
+
+    const fetchAllPutNotificationAsRead = await props.dispatch(putAllNotificationAsRead(props.user.userId))
+    switch (fetchAllPutNotificationAsRead.status) {
+      case 204:
+        props.notificationPage.list.forEach(notification => props.dispatch(readNotification(notification.id)))
```
metzgegu, line 169: I think it would be more efficient if we add a `readAllNotification` action in redux which will map the `list` without any condition (`eg .map(no => ({ ...no, read: true }))`) and `notificationNotReadCount` set to 0, it will avoid useless loop

gbonaspetti: Changed

frontend/src/reducer/notificationPage.js
```diff
@@ -74,6 +74,13 @@ export default function notificationPage (state = defaultNotificationsObject, ac
       }
     }
 
+    case `${READ}/${NOTIFICATION_LIST}`: {
+      const notificationList = action.notificationList.map(notification => (
```
metzgegu, line 78: In order to have a more stable action we should use the state here and not action data, because if it use action data (even if it will exactly the same data because we use it correctly) it doesn't ensure that the notificationList is exactly the same with state.notificationList

gbonaspetti: You are right, we don't even need to pass the action data. I'm gonna change it.

Issue #3049, opened 2020-05-14
Feat: Allow more than 6 tabs on HTTP 1.1


grignards:
The SSE implementation for the Tracim Live Messages has the inconvenient to limit the user to use 6 tabs maximum as SSE connections on one domain is limited to 6 in most browsers. 

See the warning in [Mozilla's documentation](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events).

## Implemented Solution

A broadcast channel allows the frontend to share the connection for TLMs across all the pages in the same browser.

## Past discussion and information

Several possibilities to implement this feature have been seen:

- use [HTTP/2](https://http2-explained.haxx.se/) for the transport as it allows multiplexing HTTP request on only one TCP connection: this is currently (2020-08-10) in place for the private algoo's repository but apache has a strange behavior with persistent SSE connections + http/2. See #2650 comments for more details.

- use a [shared worker](https://developer.mozilla.org/en-US/docs/Web/API/SharedWorker) (https://github.com/pubkey/broadcast-channel)) on the frontend to share the SSE connection between tabs. This can be problematic since availability of javascript features in workers may not be homogeneous across browsers, and shared workers are not supported on mobile browsers.  
- use something like [broadcast channel web API](https://developer.mozilla.org/en-US/docs/Web/API/BroadcastChannel) (or [its equivalent for non-supporting browsers]. After discussion, this is the privileged solution to investigate for now.

### HTTP/2 Pros/cons

- *Pro* only Apache configuration
- *Pro* supported by all browsers
- *Pro* potential to help the loading speed of Tracim
- *Con* has to be tested
- *Con* does it work in all network environments (business use-cases with firewall/...)

*NOTE*: an independent issue for HTTP/2 support already exist : #2650.

### SharedWorker Pros/Cons

- *Pro* HTTP/1.1, so works in any network environment
- *Con* probably more work to implement
- *Con* does not work on mobile browsers

# Discussed

 - We are planning to implement this in Tracim 3.0 so we have a solution available for any problem we could encounter with HTTP/2, should we straight up disable HTTP/2.
- We are considering BroadcastChannel

inkhey:
@grignards Just to understand Isn't possible to do both ?
Adapting tracim to be compatible http/2 and use sharedworker when possible in frontend ?

inkhey:
There is already an issue about http/2 potential support: https://github.com/tracim/tracim/issues/2650

grignards:
[...]

Issue #3409, opened 2020-08-04
Feat: Ensure missed TLMs are fetched when recovering connection


grignards:
## Feature description and goals

Starting with 3.0, the Tracim frontend has a persistent connection (through `EventSource`) to the backend. This connection can be interrupted (network failure, computer sleep, …). If TLMs are published during this interruption they are currently not received at all leading to missed user interface updates.

To avoid this we should:

- [x] Keep the event id of the last received message in the frontend
- [x] Add this id to the connection URL when reconnecting: `/api/users/2/live_messages?last_event_id=42`
- [x] When the backend has this option, it responds with all messages starting at id 43 (and still keeps the connection open).

We also:
 - [x] close all the live messages connections in the backend when a user logs out from somewhere.  
 - [x] check that the user is still logged in when re-establishing the connection to the live messages.
    - [x] If not, we redirect the user to the login page.
    - [x] otherwise, we re-establish the connection transparently.
 - [x] disable a few console.log that cause performance issues in the HTML doc: the whole state of the application was dumped on the console and this causes performance issues when there are images / a lot of comments / contents on a document. 

This issue won't address the potential issue of the number of TLMs to send, this will be handled later (see #3413).

## Extra information

We check that the user is still logged in in the frontend by fetching the /auth/whoami endpoint. A nicer solution would have been to check the status of the HTTP request sent by the EventSource object when an error is raised. Unfortunately, EventSource does not expose the HTTP status of requests…

lebouquetin:
I'm not sure about the strategy and the cost in term of developments.

Currently the behavior is `on page load` :
  - get old TLM from REST API
  - connect to SSE/TLM in order to get live updates

What I thoug is
- `on page load` : do nothing
- `on SSE/TLM connect` : 
  - get old TLM from REST API
  - will get live updates from the SSE connection.

I understand using the pushpin feature is the right way to do what we want to, but in terms of refactoring, I'm not sure it's time for doing this. What will be the impact on TLM received from API vs TLM received from SSE? This is something which was clear in the original design (with some not-working cases) but this is not clear enoguh with the current issue

grignards:
> Currently the behavior is `on page load` :
> 
>     * get old TLM from REST API
> 
>     * connect to SSE/TLM in order to get live updates

Not exactly: currently only the connection to SSE/TLM is done. The frontend will get the messages through the REST API only when opening the notification panel.

> What I thought is
> 
>     * `on page load` : do nothing
> 
[...]

PR #3432, opened 2020-08-12
Feat/3049  share TLM connection between tabs using a broadcast channel

raphj:
#3049 

## Checkpoints

**For developers**

- [X] Automated tests covering the feature have **NOT** been written: this seems hard to test automatically, this would require open multiple table in cypress or something like this

**For code reviewers**

**For testers**


frontend/src/util/LiveMessageManager.js
```diff
@@ -9,19 +10,24 @@ export const LIVE_MESSAGE_STATUS = {
   HEARTBEAT_FAILED: 'heartbeat_failed'
 }
 
+// increment this number each time the channel protocol is changed in an incompatible way
```
gbonaspetti, line 13: This comment does not follow the pattern:
// INFO - RJ - 2020-08-12  - increment this number each time the channel protocol is changed in an incompatible way

frontend/src/util/LiveMessageManager.js
```diff
@@ -126,10 +160,19 @@ export class LiveMessageManager {
...
   handleHeartbeatFailure () {
-    this.setStatus(LIVE_MESSAGE_STATUS.HEARTBEAT_FAILED)
+    this.broadcastAndSetStatus(LIVE_MESSAGE_STATUS.HEARTBEAT_FAILED)
     this.restartLiveMessageConnection()
   }
 
+  broadcastAndSetStatus (status) {
+    this.setStatus(status)
+    this.broadcastStatus(status)
+  }
```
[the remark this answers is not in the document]
raphj: it won't, in fact.

frontend/src/container/Header.jsx
```diff
@@ -77,7 +77,7 @@ export class Header extends React.Component {
 
     const fetchPostUserLogout = await props.dispatch(postUserLogout())
     if (fetchPostUserLogout.status === 204) {
-      props.tlm.manager.closeLiveMessageConnection()
+      LiveMessageManager.getInstance().closeLiveMessageConnection()
```
grignards, line 80: I'm not sure that introducing a singleton is a really good idea as it makes the code more difficult to test and restrict the used class to have only one instance (which is currently what it is, but we cannot totally be sure it will always be the case). See https://www.michaelsafyan.com/tech/design/patterns/singleton for an argumentation against singletons.

Did you d this to get rid of un-necessary refreshes of the application? If yes, couldn't we find a way to keep the object passed as props without those refreshes?

frontend/src/util/LiveMessageManager.js
```diff
@@ -70,7 +73,11 @@ export class LiveMessageManager {
     this.closeEventSourceConnection()
 
     this.eventSource = new EventSource(
-      `${url}/users/${this.userId}/live_messages`,
+      `${url}/users/${this.userId}/live_messages` + (
+        this.lastEventId
+          ? `?last_event_id=${this.lastEventId}`
+          : ''
+      ),
```
grignards, line 80: I think those changes are not part of this issue…

frontend/src/container/Header.jsx
lebouquetin: Singleton is a bad design. The right way to implement this behavior is:
- implement a normal class
- implement a static global shared instance (can be implemented through a factory that manage a unique instance globally)

I'm pretty sure this pattern is used in backend code for an example

raphj: @lebouquetin a factory has pretty much the same drawbacks as a singleton, of which I am aware (except that it forbids creating a new instance by construction). I had my reasons to go this way, you can't tell me this is bad design without asking why I did this.

You are both right though, it is best to avoid these patterns and after discussion with @grignards we found an elegant solution that solves the problem I am trying to  solve (avoid a bad refresh of the whole app when the connection goes live, which is completely unnecessary and breaks frontend tests), removes the need for a factory or a singleton, and makes the code cleaner and easier to work with: the best of every worlds.

frontend/src/util/LiveMessageManager.js
raphj: right, I reverted this.

frontend/src/container/Header.jsx
```diff
@@ -40,7 +40,6 @@ const qs = require('query-string')
 
 export class Header extends React.Component {
   componentDidMount () {
-    this.props.dispatchCustomEvent('TRACIM_HEADER_MOUNTED', {})
```
Skylsmoi, line 43: why are you removing this ? I haven't seen you adding it back somewhere else

frontend/src/container/Home.jsx
```diff
@@ -80,14 +81,14 @@ export class Home extends React.Component {
...
-    if (!(Cookies.get(COOKIE_FRONTEND.HIDE_USERNAME_POPUP) || this.props.user.username)) {
+  maybeShowUsernamePopup () {
+    if (!(this.props.user.username || Cookies.get(COOKIE_FRONTEND.HIDE_USERNAME_POPUP))) {
       this.setState({ usernamePopup: true })
     }
   }
 
   handleClickCloseUsernamePopup = () => {
-    this.setState(prevState => ({ usernamePopup: !prevState.usernamePopup }))
+    this.setState({ usernamePopup: this.state.usernamePopup })
```
Skylsmoi: This doesn't seem to do anything. What was wrong with the previous version ?

frontend/src/container/Home.jsx
```diff
@@ -133,17 +134,17 @@ export class Home extends React.Component {
   }
 
   handleClickCheckbox = () => {
-    this.setState(prevState => ({ hidePopupCheckbox: !prevState.hidePopupCheckbox }))
+    this.setState({ hidePopupCheckbox: !this.state.hidePopupCheckbox })
```
Skylsmoi: why this change ? This is React bad practice

frontend/src/container/Home.jsx
```diff
@@ -65,7 +66,7 @@ export class Home extends React.Component {
   }
 
   componentDidMount () {
-    this.checkUsername()
+    this.maybeShowUsernamePopup()
```
gbonaspetti, line 69: I don't love that a function starts with maybe, some other possibilities: `setDisplayUsernamePopup` or `setUsernamePopupVisibility`
This comment is just a suggestion, if you prefer your name yet, you can just resolve the conversation.

frontend/src/container/Home.jsx
gbonaspetti, line 137: same comment about setState

frontend/test/src/LiveMessageManager.spec.js
```diff
@@ -4,25 +4,45 @@ import {
...
+
+const setStatusOrig = LiveMessageManager.prototype.setStatus
+LiveMessageManager.prototype.setStatus = function (status) {
+  setStatusOrig.call(this, status)
+
+  const resolve = (this.statuscallbacks && this.statuscallbacks.shift())
+  resolve && resolve()
+}
+
+LiveMessageManager.prototype.onstatuschange = function (callback) {
```
Skylsmoi, line 27: shouldn't you name this function in camel case ?

frontend/test/src/LiveMessageManager.spec.js
```diff
@@ -4,25 +4,45 @@ import {
...
-const waitForStatus = async (manager, status, timeout = 25) => {
-  // NOTE SG 2020-07-03 - have to wait a small amount as manager status changes
-  // can be done via a setTimeout
-  var retries = 10
-  while (manager.status !== status && retries > 0) {
-    await new Promise(resolve => setTimeout(resolve, timeout / 10))
-    retries--
-  }
+const waitForStatus = async (manager) => {
+  await new Promise(resolve => { manager.onstatuschange(resolve) })
```
Skylsmoi, line 45: I'm not sure I understand the purpose awaiting a new Promise just to give the callback to onstatuschange.
Can you explain the behavior of this please ?

frontend/src/container/Home.jsx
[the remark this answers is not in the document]
raphj: Interesting, thanks. I am going to revert it.

frontend/src/container/Home.jsx
raphj: Resolving this, @gbonaspetti already raised this issue.

frontend/test/src/LiveMessageManager.spec.js
raphj: I figured I would follow the case for the event handlers used everywhere on the DOM. Are you fine with this or should I change this?

(onmousemove, onreadystatechange, etc)

frontend/src/container/Header.jsx
raphj: It's not used anywhere as far as I could see. It's therefore unnecessary and untested code, and I was trying to change unneeded state changes in this PR.

Should I revert this anyway?

frontend/test/src/LiveMessageManager.spec.js
Skylsmoi: mmh it makes sense but it also surprised me to see all your new functions in camel case but this one.
I'd say to put it in camel case to match all the others functions name of the file

frontend/test/src/LiveMessageManager.spec.js
raphj: I am waiting for the Promise to resolve. The code continues executing as soon as the promise resolves.

frontend/src/container/Header.jsx
Skylsmoi: It is useful, keep it.
The purpose of this is to allow external apps to do stuffs based on Tracim's event

raphj: all right! I'm going to revert this.

I have a question however: isn't this prematurilly guessing what will be needed in an uncertain future on a architecture that may change sooner or later anyway?

frontend/test/src/LiveMessageManager.spec.js
Skylsmoi: shouldn't .onstatuschange() be asynchronous then ? That way you wouldn't have to set a callback mechanism

raphj: Mhm, yes it is, and I did. For `onstatuschange`, resolve is just a regular callback here. I'm using it in a Promise, but it does not know about this.

frontend/test/src/LiveMessageManager.spec.js
raphj: ok :-)

frontend/test/src/LiveMessageManager.spec.js
raphj: I turned this function into an asynchronous function and used camel case. Does it looks good to you @Skylsmoi?

frontend/src/util/LiveMessageManager.js
```diff
@@ -9,62 +10,99 @@ export const LIVE_MESSAGE_STATUS = {
...
-      `${url}/users/${userId}/live_messages`,
+    this.setStatus(LIVE_MESSAGE_STATUS.PENDING)
+
+    if (!this.broadcastChannel) {
+      this.broadcastChannel = new BroadcastChannel(BROADCAST_CHANNEL_NAME)
+      this.broadcastChannel.addEventListener('message', this.broadcastChannelMessageReceived.bind(this))
+      this.broadcastChannel.postMessage({ canIHaveStatus: true })
+      this.electLeader()
+    }
+  }
```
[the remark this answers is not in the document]
raphj: it is done in `closeLiveMessageConnection` @grignards, or do you speak about something else?

gbonaspetti (approved):

frontend/test/src/LiveMessageManager.spec.js
```diff
@@ -21,12 +21,12 @@ LiveMessageManager.prototype.setStatus = function (status) {
   setStatusOrig.call(this, status)
 
   const resolve = (this.statuscallbacks && this.statuscallbacks.shift())
-  resolve && resolve()
+  resolve && resolve(this.status)
 }
 
-LiveMessageManager.prototype.onstatuschange = function (callback) {
+LiveMessageManager.prototype.onStatusChange = async function () {
```
Skylsmoi, line 27: I'm confused because you only `await` for this function in unit tests, I'm not sure to understand

raphj: It is only present in unit tests.

Skylsmoi: oh, the file is in tests folder ok.

Skylsmoi (approved):

raphj:
NOT tested on branch, we are going to test this in branch `maintenance/3.0-more_reliable_tlms` with the work on #3409.

PR #3448, opened 2020-08-19
Feat/3409 Clean resume after suspend

raphj:
#3409 

## Checkpoints

**For developers**

**For code reviewers**

**For testers**


backend/tracim_backend/tests/functional/test_live_messages.py
```diff
@@ -38,6 +39,52 @@ def big_html_document(workspace_api_factory, content_api_factory, session) -> Co
...
+
+def put_document(html_document):
+    update_user_request = requests.put(
+        "http://localhost:7999/api/workspaces/{}/html-documents/{}/status".format(
+            html_document.workspace_id, html_document.content_id
+        ),
+        auth=("admin@admin.admin", "admin@admin.admin"),
+        json={"status": "closed-validated"},
+    )
+    assert update_user_request.status_code == 204
```
grignards, line 85: You should put your asserts in the test method because asserts in other code won't be beautified by pytest. Here you could return the `status_code` then assert in the test method.
By the way if this code is not what you are testing, you shouldn't assert and it should be in a fixture.

backend/tracim_backend/views/core_api/user_controller.py
```diff
@@ -615,6 +616,20 @@ def open_message_stream(self, context, request: TracimRequest, hapic_data=None)
         stream_opened_event = ":Tracim Live Messages for user {}\n\nevent: stream-open\ndata:\n\n".format(
             str(request.candidate_user.user_id)
         )
+
+        if "after_event_id" in hapic_data.query:
```
grignards, line 620: You don't need the tests here, provided you have a schema-defined default value for after_event_id.

backend/tracim_backend/views/core_api/schemas.py
```diff
@@ -1424,6 +1424,9 @@ class TracimLiveEventHeaderSchema(marshmallow.Schema):
     # TODO - G.M - 2020-05-14 - Add Filtering for text/event-stream mimetype with accept header,
     #  see: https://github.com/tracim/tracim/issues/3042
     accept = marshmallow.fields.String(required=True, load_from="Accept", dump_to="Accept")
+    after_event_id = marshmallow.fields.Int(
+        required=False, example=42, validate=strictly_positive_int_validator
```
grignards, line 1428: You should but a `default=0` (or `missing`, I don't remember) as your query method declares to only handle `int`, not `None`.

frontend/src/util/LiveMessageManager.js
```diff
@@ -75,7 +76,11 @@ export class LiveMessageManager {
     this.closeEventSourceConnection()
 
     this.eventSource = new EventSource(
-      `${url}/users/${this.userId}/live_messages`,
+      `${url}/users/${this.userId}/live_messages` + (
+        this.lastEventId
+          ? `?after_event_id=${this.lastEventId}`
+          : ''
+        ),
```
grignards: Wouldn't it be more readable by writing like this?

```javascript
let afterEventFilter = ''
if this.lastEventId:
  afterEventFilter = '?after_event_id="${this.lastEventId}`
this.eventSource = new EventSource(
   `${url}/users/${this.userId}/live_messages${afterEventFilter}`
...
```

frontend/test/src/LiveMessageManager.spec.js
```diff
@@ -47,64 +51,137 @@ describe('LiveMessageManager class', () => {
...
       manager.closeLiveMessageConnection()
       expect(manager.status).to.be.equal(LIVE_MESSAGE_STATUS.CLOSED)
+      manager.closeLiveMessageConnection()
     })
   })
 
   describe('the heartbeat timer', async () => {
-    it('should restart connection when the heartbeat event is not received in time', async () => {
-      const { manager } = openedManager(2, 10)
+    it('should restart connection when the heartbeat event is not received on time', async () => {
```
grignards, line 111: I beg to differ, `in` seems ok to me as we do not compare the exact time of arrival of heartbeats, only that they are not received too late.

grignards (changes requested):

backend/tracim_backend/tests/functional/test_live_messages.py
raphj: Ok, I have removed the assertion from the fixture, added it back to the test regarding the big document, and the status is not checked anymore for the other tests.

frontend/src/util/LiveMessageManager.js
raphj: I don't have a strong opinion here, I am going to go for your version.

frontend/test/src/LiveMessageManager.spec.js
raphj: I implore your pardon sir, you are very right according to https://www.anglaisfacile.com/exercices/exercice-anglais-2/exercice-anglais-16425.php

I'm going to restore your perfectly fine "in".

backend/tracim_backend/views/core_api/user_controller.py
```diff
@@ -607,14 +609,31 @@ def get_user_messages(
...
+        if after_event_id:
+            app_config = request.registry.settings["CFG"]  # type: CFG
+            event_api = EventApi(request.current_user, request.dbsession, app_config)
+            messages = event_api.get_messages_for_user(
+                request.candidate_user.user_id, after_event_id=after_event_id
+            )  # type: typing.List[Message]
+
+            stream_opened_event += "".join(
+                [
+                    "data:" + json.dumps(LiveMessageSchema().as_dict(message)) + "\n\n"
```
grignards, line 632: Avoid re-creating a schema for each message, you can use one instance only. It could even be declared as a class attribute.

grignards (changes requested):

backend/tracim_backend/views/core_api/user_controller.py
raphj: Would you put this in the UserController class? This is also used for producing live message, would we store a class instance that is accessible to both the code producing the live messages and here?
Does it makes sense to actually have a ready to use class instance static `as_dict` method in LiveMessageSchema that uses a "cached" instance as a private attribute?

raphj: Done, in LiveMessagesLib.

grignards (approved):

Issue #2849, opened 2020-04-08
Display a prompt to notify members of the workspace when editing a file


Skylsmoi:
## Feature description and context

After saving a new version of a content, add a prompt in the app with a button to notify members of the new version. This prompt is only shown if the last modification of the content was made by the logged user. As long as the user does not make a decision, every time they enter the content the message will be there.
This is a way to notify members about a specific new version that might be important.

- [x] add the prompt if the user made the last change
- [x] if the user clicks on `X`, the prompt closes
- [x] if the user clicks to notify, a comment is automatically sent with *`@all` please notice that I did an important update on this content.* (this comment will generate a notification)

## Translations

| English | French | Portuguese |
|--|--|--|
| `@all` please notice that I did an important update on this content. | `@tous` veuillez noter que j'ai fait une mise à jour importante sur ce contenu. | `@todos` percebam que fiz uma actualização importante neste conteúdo. |
| To notify all members of the shared space of your modification, click here! | Pour notifier tous les membres de l'espace partagé de votre modification, cliquez ici ! | Para notificar a todos os membros do espaço partilhado da sua modificação, clique aqui! |

## Prerequisites

#3411  

## Interface

![notifications_notifier_message](https://user-images.githubusercontent.com/35422692/89786494-4e73ca80-db1c-11ea-88bb-4d4687f44114.jpg)

This prompt will be inside of the content part (left part) and will have the same color and specifications that the other prompts in the same place.
_Other prompts: deprecated and deleted content._

## Discussion

- _As long as the user does not make a decision, every time he enters the content the message will be there._
    - We'll have a backend variable or just a cookie? If we chose the cookie and the user decides not to notify, the next day he enters at this content, the message will appear again or we will have a timeout of the type "the message only appears if the user made the last change AND this change was less than X hours/days ago". We can also use the generic enpoint (https://github.com/tracim/tracim/issues/3380) to set user decision.

  - *2020-08-24: we'll use the generic config endpoint*. The stored key will be `content.<content_id>.notify_all_members_message`, its value is a bool.

----
_10-07-2020 - Description edition: Put context at the beginning and added task list, impacted components and backend variable vs cookie discussion._
_13-07-2020 - Description edition: Added translations._
_20-07-2020 - Description and label edition: Change impacted components to labels and comment text._
_21-07-2020 - Description and label edition: message → prompt to clarify._
_10-08-2020 - Add interface._
_24-08-2020 - Resolve discussions._

[...]

Issue #2843, opened 2020-04-07
Add mention process for all contents


Skylsmoi:
## Feature description

Mentioning someone will create a notification to the related users. The mentions can be done in a comment and in HTML documents.

- [x] add autocomplete list that it is displayed in front of the current interface after writing `@` and one following letter at the beginning of a sentence or after a blank space` ` (see list interface bellow)
  - auto complete popup must list the matching know members in the shared space (username, full name or email)
  - related backend issue: https://github.com/tracim/tracim/issues/2830
- [x] Pressing the _Enter_ key should complete with the user being hovered
     - there should always be a  "selected" user. By default, it is the first of the list
- [x] add group mentions on autocomplete lists (just `@all`)
    - these mentions should be translated according to the current language 
- ~~the comment parsing will be the addition of a html tag with at least a id unique~~
  - this task has been moved to its specific issue here: https://github.com/tracim/tracim/issues/3411

## Prerequisites

- backend for autocomplete https://github.com/tracim/tracim/issues/2830
- backend for TLM mentions https://github.com/tracim/tracim/issues/2838 

## Interface

[ Need @lebouquetin validation ]

![notifs_mention_popup_in_app](https://user-images.githubusercontent.com/50453784/88678678-00f46800-d0ef-11ea-890e-a6e611cdcc12.png)

![notifs_mention_popup_comment_advanced](https://user-images.githubusercontent.com/50453784/88678682-02be2b80-d0ef-11ea-8b85-388fc23befb2.png)

![notifs_mention_popup_comment](https://user-images.githubusercontent.com/50453784/88678695-05208580-d0ef-11ea-83b5-d99f39d4569f.png)

### TextArea

In order to show the autocomplete pop-up when typing a new comment, the textarea is parsed to find a mention candidate before the current cursor position (e.g. `@word|`), if a mention candidate is found the autocmplete popup will show up and it will fetch users according to the word typed. The user can validate his choice by pressing enter or click the mention item, it replace the mention candidate by the correct mention.

## TinyMCE

TinyMCE autompletion use the same popup as the texterea. but the trigger to open the popup is different. Indeed tinymce edit directly the html and getting the current selection is different. The autocomplete popup will be triggered if the user type `@` and will stay activated while the user type a correct mention candidate (e.g. `@word|` with `|` the current cursor position), when the user will validate the mention from the popup it will use html `<span>` to select and replace the mention in the text.

# Backend

[...]

PR #3475, opened 2020-08-31
Feature/2849  display prompt to notify all members

gbonaspetti:
#2849 

## Checkpoints

**For developers**

**For code reviewers**

**For testers**


frontend_lib/i18next.scanner/fr/translation.json
```diff
@@ -1,4 +1,5 @@
 {
+  "@all please notice that I did an important update on this content.": "@tous veuillez noter que j'ai fait une mise à jour importante sur ce contenu.",
```
raphj, line 2: So we settled for translated mentions?

raphj (approved):

frontend_lib/i18next.scanner/fr/translation.json
gbonaspetti: yes, https://github.com/tracim/tracim/issues/2843 (third check point) we have just `@all` but it's translated

raphj (approved):

raphj:
tested on branch and works correctly. However, I have one problem and one question with the current implementation:
 - the click here button should either be a link ("a" tag) with an outliine, or remain a button but have an outline when it is focused. The outline should in this case probably look like the outline of the buttons at the top of the app.
 - the language of the message that is left when clicking the button is in the language  of the Tracim UI. Which is fine for now, but should we have a concept of "the language of a workspace" that would be used for this message? I can very well see a workspace used by people speaking different languages, where the working language is common but users can have their preferences regarding which language they use for the UI personally and this shouldn't impact anything other than the UI.

Issue #3523, opened 2020-09-09
Bug: mentions should appear in the notification wall but modifications of the logged user shoudn't


raphj:
## Feature description and goals

Notifications shown in the wall are going to be filtered. 
- Mentions should appear in the wall with the text and the icon showed here https://github.com/tracim/tracim/issues/2840.
- Updates made by the logged user shouldn't make a notification for itself 

### Expected

- [x] (front) Handle mention TLM in ReduxTlmDispatcher.jsx
- [x] (front) Filter TLM from the logged user in ReduxTlmDispatcher.jsx
- [x]  (front) Check if the notifications gotten from the API in Notification.jsx has a filter about the logged user.
- [x] (back) implement parameter `exclude_author_ids` for the `/messages` endpoint
- [x] (front) use parameter `exclude_author_ids`

## Diagnostic

- The mentions are already handled at NotificationWall.jsx component, but not its TLM in ReduxTlmDispatcher.jsx.

## Extra information

:warning: This fix bumps the documented minimal version of PostgreSQL from 9.3 to 9.6. It relies on a fix likely introduced in [version 9.4](https://www.postgresql.org/docs/9.4/release-9-4.html):

>    Corner cases in the JSON field/element/path extraction operators now return NULL rather than raising an error (Tom Lane)
 
We needed to extract a value from a JSON object which is sometimes NULL. We decided to match the PostgreSQL version that we battle-test Tracim against, which is also the one that is shipped in Debian Stretch, but Tracim likely still works on PostgreSQL 9.4+. Note however that PostgreSQL 9.4 and 9.5 have been unsupported by the PostgreSQL project for some time now.
 ________

2020-09-10 - exclude_author_id → exclude_author_ids
2020-09-14 - added the extra information section

raphj:
tested on branch by @inkhey.

PhilippeAccorsi:
Verified.

PR #3535, opened 2020-09-09
Fix/3523 mentions and only other users notifications in wall

raphj:
#3523

## Checkpoints

**For developers**

Automated tests are not there yet. It's coming!

**For code reviewers**

**For testers**


backend/tracim_backend/lib/core/event.py
```diff
@@ -120,8 +121,15 @@ def _base_query(
             else:
                 query = query.filter(event_type_filters[0])
 
+        if exclude_author_id:
+            print("OK")
```
grignards, line 125: Do not forget to remove the `print`

backend/tracim_backend/lib/core/event.py
```diff
@@ -163,13 +171,17 @@ def get_messages_for_user(self, user_id: int, after_event_id: int = 0) -> typing
     def get_paginated_messages_for_user(
         self,
         user_id: int,
+        exclude_author_id: int,
```
grignards, line 174: This parameter can be `None` I think.

backend/tracim_backend/lib/core/event.py
```diff
@@ -190,7 +206,7 @@ def create_event(
         additional_fields: typing.Dict[str, JsonDict],
         context: TracimContext,
         entity_subtype: typing.Optional[str] = None,
-    ) -> Event:
+    ) -> None:
```
grignards, line 209: I would rather add a `return event` in the method as it seems natural to be able to get it (but you're right that it was forgotten).

backend/tracim_backend/models/context_models.py
```diff
@@ -600,20 +600,27 @@ def __init__(
         count: int,
         event_types: Optional[List[EventTypeDatabaseParameters]] = None,
         page_token: Optional[str] = None,
+        exclude_author_id: Optional[int] = 0,
```
grignards, line 603: Should the default value be `None` instead (it gives the same result of course, as long as we do not have user with id equal to 0)?

backend/tracim_backend/models/context_models.py
```diff
@@ -600,20 +600,27 @@ def __init__(
...
 class UserMessagesSummaryQuery(object):
     """
     Message summary query model
     """
 
-    def __init__(self, event_types: Optional[List[EventTypeDatabaseParameters]] = None,) -> None:
+    def __init__(
+        self,
+        event_types: Optional[List[EventTypeDatabaseParameters]] = None,
+        exclude_author_id: int = 0,
```
grignards, line 620: Same comment as above.

backend/tracim_backend/views/core_api/schemas.py
```diff
@@ -109,6 +109,11 @@ def _deserialize(self, value, attr, data, **kwargs):
         return None
 
 
+ExcludeAuthorIdField = marshmallow.fields.Int(
+    example=10, validate=positive_int_validator, missing=0, allow_none=False,
```
grignards, line 113: Same comment here.

grignards: BTW the naming is too restrictive as this field type is not reserved to author ids. Maybe something like `PositiveInteger`?

frontend/src/container/ReduxTlmDispatcher.jsx
```diff
@@ -68,41 +71,58 @@ export class ReduxTlmDispatcher extends React.Component {
...
+    this.handleNotification(data)
   }
 
   handleMemberDeleted = data => {
-    this.props.dispatch(removeWorkspaceMember(data.fields.user.user_id, data.fields.workspace.workspace_id))
-    if (this.props.user.userId === data.fields.user.user_id) this.props.dispatch(removeWorkspace(data.fields.workspace))
-    this.props.dispatch(addNotification(data))
+    const { props } = this
+    props.dispatch(removeWorkspaceMember(data.fields.user.user_id, data.fields.workspace.workspace_id))
+    if (props.user.userId === data.fields.user.user_id) props.dispatch(removeWorkspace(data.fields.workspace))
```
grignards, line 107: I know you didn't write this line, but isn't the condition strange here? It makes me think that other tabs of a user who removes a workspace won't be updated...

backend/tracim_backend/lib/core/event.py
gbonaspetti, line 125: forgotten print?

frontend/src/container/ReduxTlmDispatcher.jsx
```diff
@@ -68,41 +71,58 @@ export class ReduxTlmDispatcher extends React.Component {
...
-    this.props.dispatch(addNotification(data))
+    const { props } = this
+    props.dispatch(addWorkspaceContentList([data.fields.content], data.fields.workspace.workspace_id))
+    this.handleNotification(data)
   }
 
   handleContentCommentCreated = async data => {
-    if (data.fields.author.user_id === this.props.user.userId) return
+    const { props } = this
+    if (data.fields.author.user_id === props.user.userId) return
```
grignards, line 123: Same comment as above, shouldn't the condition be `the tlm comes from this tab/window`? That we can test with `data.fields.clientTokenId`?

functionnal_tests/cypress/integration/login/username_popup_spec.js
```diff
@@ -149,6 +149,7 @@ describe('Login', function () {
         .click()
     })
     it('should not open the set username popup', function () {
+      this.skip() // FIXME - GB - 2020-09-09 - this tests is unstable and it will be fixed at https://github.com/tracim/tracim/issues/3483
```
grignards, line 152:
```suggestion
      this.skip() // FIXME - GB - 2020-09-09 - this test is unstable and it will be fixed at https://github.com/tracim/tracim/issues/3483
```

frontend/src/container/ReduxTlmDispatcher.jsx
```diff
@@ -68,41 +71,58 @@ export class ReduxTlmDispatcher extends React.Component {
     ])
   }
 
+  handleNotification (data) {
```
grignards, line 74: I know there isn't a clear rule for method definition, but here all other methods are defined with `= data => {`, so I would rather keep this definition flavor.

grignards (changes requested):

backend/tracim_backend/models/context_models.py
raphj: I hesitated, I'm fine with None. I'll change this.

backend/tracim_backend/views/core_api/schemas.py
raphj: Indeed. Anyway, this is going to be changed to a list of positive integer. Are you fine with PosivieIntegerListField?

frontend/src/container/ReduxTlmDispatcher.jsx
raphj: @grignards We need to talk about this. They are not the same.

frontend/src/container/ReduxTlmDispatcher.jsx
raphj: @gbonaspetti git says that you wrote this line, can you shed some light please?

frontend/src/container/ReduxTlmDispatcher.jsx
raphj: Ping @gbonaspetti

gbonaspetti: Use the clientTokenId will make the changes inside this if only for this tab, but we want to set the read state for all tabs that the user is connected, right?

frontend/src/container/ReduxTlmDispatcher.jsx
raphj: Re-reading this line, it seems correct. @grignards are you confusing `data.fields.user` with `data.fields.author`?

Here, `data.fields.user` is removed from `data.fields.workspace`. If we are user `data.fields.user`, we should remove the workspace from the UI.

frontend/src/container/ReduxTlmDispatcher.jsx
gbonaspetti: @Skylsmoi has more knowledge than I do to talk about it, but I'd like to add that regular functions are often linked to the context that they are invoked (this), and if I'm not wrong arrow functions are pure in that sense. Here the function doesn't seems to me that it depends of the context. Can you explain to me why in this case you this a regular function it's better ?

raphj: Because it belongs to the prototype of the class, instead of being a field of the specific instance.

Here's the difference:

```js
class C {
    method() {
        console.log(this)
    }
}
```
is translated to:

```js
function C () {}
C.prototype.method =  {
    console.log(this)
}
````

Whereas

```js
class C {
    method = () => {
        console.log(this)
    }
}
```

is translated to:

```js
function C() {
    this.method = () => {
        console.log(this)
    }
}
```
Equivalent to:

```
function C() {
    this.method = (function () {
        console.log(this)
    }).bind(this)
}
```

You can try this by running `(new C()).hasOwnProperty('method')`. In the first case, you get false, because `method` is in the prototype of the class, while in the second case, you get `true`, because `method` is not a method: it is a field of the instance that happens to be a function. 

You are right with respect to the context `this`. And our way of doing is useful for event handlers that need `this`.

I know in the Tracim frontend we widely use `method =  () => {}` but this breaks the prototypical class model of Javascript, makes proper inheritance more difficult and is probably less memory efficient. I can revert to  `method =  () => {}` but I think we really should progressively tend towards using proper Javascript methods in Tracim. I think we should limit `method =  () => {}` to event handlers that use `this`.

gbonaspetti: Don't worry, I'm not telling you how to do it, I'm just trying to understand why, in general I know some advantages in using arrow functions, but I don't know the advantages of the regular ones. In this case, you are using `this.props` within the function, wouldn't you need `bind`?

raphj: I understood that you asked out of curiosity (and I answered in this spirit), and I'll probably revert anyway for consistency reasons, but to answer your question: no, no need to use bind here.

```js
class C {
    f() {
        console.log(this.answer, this instanceof C)
    }
    g() {
        this.f() // is this bound here?
    }
}

const c = new C()
c.answer = 42
c.g() // prints 42 true
```

When using object.method() (or this.method()), method is bound to this object at call time.

The story is (unfortunately) different if you use a callback:

```js
function call(f) {
    f()
}

call(c.g)  // this is not bound - TypeError: Cannot read property 'f' of undefined at g, at call
```

You'd need to bind here:

```js
call(c.g.bind(c))  // prints 42 true
```

raphj: As for an advantage of using regular methods instead of the arrow functions:
 - first, it's more concise
 - second: inheritance works:

```js
class C {
  f() {
    console.log(42)
  }
}

class D extends C {
  f () {
    super.f()
    console.log('my extension')
  }
}

d = new D()
d.f()

```
This prints:

```
42
my extension
```

Compared to:

```js
class C {
  f = () => {
    console.log(42)
  }
}

class D extends C {
  f = () => {
    super.f()
    console.log('my extension')
  }
}

const d = new D()
d.f()
```

This gives:

```
TypeError: (intermediate value).f is not a function
    at D.f (/home/raphael/t.js:9:11)
    at Object.<anonymous> (/home/raphael/t.js:15:3)
    at Module._compile (internal/modules/cjs/loader.js:955:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:991:10)
    at Module.load (internal/modules/cjs/loader.js:811:32)
    at Function.Module._load (internal/modules/cjs/loader.js:723:14)
    at Function.Module.runMain (internal/modules/cjs/loader.js:1043:10)
    at internal/main/run_main_module.js:17:11
```

Classes written as we write them in Tracim cannot really be extended.

Skylsmoi: If I may, while I agree with all the above, to me, they are not good enough reason to state a clear rule.

Here's why:

About `this` binding, there is an eslint plugin that already does it for every functions of the components.

About arrow function makings it harder for inheritance, it's true but we don't do inheritance it is complicated for components because they already all extends from React.Component. And React guidelines are to avoid inheritance.

About arrow function having poorer performances, it's also true but the loss can't make any difference on our app since we don't instantiate the same components that often on pages. While this is only true for modern browsers, we already have poor support for old ones and we barely test on them (like IE).
___
Originally, when I was the only frontend dev on Tracim and when I didn't know all the differences between the 2 function declarations, I used the regular function declaration for inherited functions from React.Component (componentDidUpdate, render ...) and everything else was arrow functions. I'm not saying it's a good rule.

I'm not against stating a clear rule but I would prefer it to be for a good reason.
Remember that, if you state a strict rule, it comes with a choice: refactor all our code base or agree that our own code base doesn't respect the rule.
Obviously, refactoring is better, but it comes with a cost.

This conversation is a good technical dept subject that can be put in an issue with the according tag. Things we, Raphael, maybe should have already done since we already discussed about it. It's just that I don't think now a good time to decide and apply said rule.

For now, the most important thing to me is knowing that all developers working on tracim know the differences between regular functions and arrow functions.

gbonaspetti (approved):

frontend/src/container/ReduxTlmDispatcher.jsx
raphj: Thanks for your comment @Skylsmoi. In any case, I don't think that we should change everything in the code base should we adopt a new rule (for anything, not necessarily on this specific point) (which is not today anyway). This is costly and error prone. But we should allow ourselves to adopt new rules for new code and gradually moving.

> For now, the most important thing to me is knowing that all developers working on tracim know the differences between regular functions and arrow functions.

Me too :-)
and class methods vs regular functions / arrow functions assigned to each instance.

backend/tracim_backend/views/core_api/schemas.py
```diff
@@ -109,6 +109,22 @@ def _deserialize(self, value, attr, data, **kwargs):
         return None
 
 
+class PositiveIntegerListField(StrippedString):
+    def _deserialize(self, value, attr, data, **kwargs):
+        value = super()._deserialize(value, attr, data, **kwargs)
+        if value:
+            return [int(item) for item in value.split(",")]
+        return None
```
inkhey, line 117: You should add a specific validator for this one to avoid weird None case, see `regex_string_as_list_of_int`

Also, we should probably avoid code replication here, we currently do something very similar to this with `string_to_list` method

inkhey (changes requested):
About backend, take a look at : 
- validator
- existing code doing same stuff to adapt code in a way we do not have 2 way to do same thing

backend/tracim_backend/views/core_api/schemas.py
raphj: done

frontend/src/container/ReduxTlmDispatcher.jsx
grignards: Yes, I was confusing author and user. This is OK.

frontend/src/container/ReduxTlmDispatcher.jsx
grignards: Yes, you're right.

backend/tracim_backend/lib/core/event.py
```diff
@@ -90,6 +90,7 @@ def _base_query(
         event_id: typing.Optional[int] = None,
         user_id: typing.Optional[int] = None,
         event_types: typing.List[EventTypeDatabaseParameters] = None,
+        exclude_author_ids: typing.List[int] = None,
```
grignards, line 93: This is an `Optional[List...]`:

```suggestion
        exclude_author_ids: typing.Optional[typing.List[int]] = None,
```

grignards (changes requested):

grignards (approved):

lebouquetin:
@raphj why do you upgrade postgresql minimum version in this issue? This is not the place, and if we do support postgresql 9.3 we must have a minimum version to 9.3

inkhey:
Just tested found some issue:
- for some reasons, it seems that icons are uncorrect according to the previous spec, need to be clarified.
- event workspace.created is seen by user itself.
- translation issue here in french:
Capture du 2020-09-11 10-29-35
(NOUVEAU is a space).

grignards (approved):

raphj:
The failing test is the unstable one. This has been tested by @inkhey.

raphj:
For posterity, here is the error message we got on PostgreSQL 9.3:

```
_ TestMessages.test_api__get_messages__ok_200__exclude_author_ids_filter[config_section0] _

self = <backend.tracim_backend.tests.functional.test_message.TestMessages object at 0x7fb11cd9c0b8>

session = <sqlalchemy.orm.session.TracimSession object at 0x7fb0f466ba90>

web_testapp = <webtest.app.TestApp object at 0x7fb0f2c39b00>

    def test_api__get_messages__ok_200__exclude_author_ids_filter(

        self, session, web_testapp

    ) -> None:

        """

        Get messages through the classic HTTP endpoint. Filter by event_type

        """

        messages = create_events_and_messages(session)

    

        web_testapp.authorization = ("Basic", ("admin@admin.admin", "admin@admin.admin"))

    

        result = web_testapp.get("/api/users/1/messages?exclude_author_ids=", status=200,).json_body

        message_dicts = result.get("items")

        assert len(message_dicts) == len(messages)

    

        result = web_testapp.get(

>           "/api/users/1/messages?exclude_author_ids=1", status=200,

        ).json_body

/home/travis/build/tracim/tracim/backend/tracim_backend/tests/functional/test_message.py:94: 

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

/home/travis/virtualenv/python3.6.10/lib/python3.6/site-packages/webtest/app.py:328: in get

    expect_errors=expect_errors)

/home/travis/virtualenv/python3.6.10/lib/python3.6/site-packages/webtest/app.py:649: in do_request

    self._check_status(status, res)

_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <webtest.app.TestApp object at 0x7fb0f2c39b00>, status = 200

res = <500 Internal Server Error application/json body=b'{"messag...ull}'/729>

    def _check_status(self, status, res):

        if status == '*':

            return

        res_status = res.status

        if (isinstance(status, string_types) and '*' in status):

            if re.match(fnmatch.translate(status), res_status, re.I):

                return

        if isinstance(status, string_types):

            if status == res_status:

                return

        if isinstance(status, (list, tuple)):

            if res.status_int not in status:

                raise AppError(

                    "Bad response: %s (not one of %s for %s)\n%s",

                    res_status, ', '.join(map(str, status)),

                    res.request.url, res)

            return

        if status is None:

            if res.status_int >= 200 and res.status_int < 400:

                return

            raise AppError(

                "Bad response: %s (not 200 OK or 3xx redirect for %s)\n%s",

                res_status, res.request.url,

                res)

        if status != res.status_int:

            raise AppError(

>               "Bad response: %s (not %s)\n%s", res_status, status, res)

E           webtest.app.AppError: Bad response: 500 Internal Server Error (not 200)

E           b'{"message": "(psycopg2.errors.InvalidParameterValue) cannot extract element from a scalar\\n\\n[SQL: SELECT messages.receiver_id AS messages_receiver_id, messages.event_id AS messages_event_id, messages.sent AS messages_sent, messages.read AS messages_read \\nFROM messages JOIN events ON events.event_id = messages.event_id \\nWHERE messages.receiver_id = %(receiver_id_1)s AND CAST(((events.fields -> %(fields_1)s) ->> %(param_1)s) AS INTEGER) != %(param_2)s ORDER BY messages.event_id DESC \\n LIMIT %(param_3)s]\\n[parameters: {\'receiver_id_1\': 1, \'fields_1\': \'author\', \'param_1\': \'user_id\', \'param_2\': 1, \'param_3\': 11}]\\n(Background on this error at: http://sqlalche.me/e/13/9h9h)", "details": {"error_detail": {}}, "code": null}'

/home/travis/virtualenv/python3.6.10/lib/python3.6/site-packages/webtest/app.py:684: AppError
```

[PostgreSQL 9.4](https://www.postgresql.org/docs/9.4/release-9-4.html) is probably the release that fixed this issue:

> Corner cases in the JSON field/element/path extraction operators now return NULL rather than raising an error (Tom Lane)

Issue #3536, opened 2020-09-10
Feat: Add an access type parameter to spaces


grignards:
## Feature description and goals

Currently, the only way for a user to be a member of a Tracim space is for a space manager to add them to the space.
It can be useful to have other access methods for selected spaces in order to add "discoverability" to Tracim instances.

- [x] Add an `access_type` parameter to space APIs.
- [x] The access type parameter can be:
  - `CONFIDENTIAL`: the current space access type
  - `ON_REQUEST`: those spaces are visible to any user (but their contents are only accessible by their members). A user will be able to request access to those spaces. Access implementation is detailed in https://github.com/tracim/tracim/issues/3539 and is not part of this issue
  - `OPEN`: those spaces are accessible by any user of the Tracim instance. Access implementation is detailed in #3538 and is not part of this issue
- [x] Add a database migration script that will add the `access_type` to the `workspaces` table and set it to `CONFIDENTIAL` for all existing rows
- [x] The `access_type` must be given when creating the space through the API
- [x] We won't initially allow to modify the `access_type` of an existing space, so restrict it in the corresponding PUT API.

## Discussed before development

- is not allowing the modification of `access_type` after the space creation acceptable?

It is acceptable, see comments below.

## Extra information
Implemented solution here: https://github.com/tracim/tracim/pull/3585
Contain also frontend fix to force usage of access_type.

raphj:
>   is not allowing the modification of access_type acceptable?

I guess it is acceptable, and we can always revert this decision later if needed.

inkhey:
>     * is not allowing the modification of `access_type` acceptable?

What the goal of not allowing it ? We can keep the default value to CONFIDENTIAL and this way we can keep the current behaviour of frontend.

grignards:
> > ```
> > * is not allowing the modification of `access_type` acceptable?
> > ```
> 
> What the goal of not allowing it ? We can keep the default value to CONFIDENTIAL and this way we can keep the current behaviour of frontend.

Changing the access type (OPEN -> (ON_REQUEST/CONFIDENTIAL)) after creation means that we'd potentially have to remove users from the space/do other things that administrators could expect. It also means that the frontend has to provide an interface for this change.
Nothing unfeasible, just takes time…

lebouquetin:
I agree with @grignards : first iteration need to add this property at creation time. Allowing to update it will introduce side effects which are not necessarely usefull.

Example:

1. User XXX requested to join space YYY
2. space YYY become OPEN
[...]

PR #3585, opened 2020-09-15
Feature/3536  add access type parameter to space

inkhey:
related to #3536 

Add access type to both workspace api and workspace database object

## Checkpoints

**For developers**

**For code reviewers**

**For testers**

### DIFF
--- a/backend/tracim_backend/models/context_models.py
+++ b/backend/tracim_backend/models/context_models.py
@@ -32,6 +32,7 @@ from tracim_backend.models.data import ContentNamespaces
 from tracim_backend.models.data import ContentRevisionRO
 from tracim_backend.models.data import UserRoleInWorkspace
 from tracim_backend.models.data import Workspace
+from tracim_backend.models.data import WorkspaceAccessType
 from tracim_backend.models.event import EventTypeDatabaseParameters
 from tracim_backend.models.event import ReadStatus
 from tracim_backend.models.roles import WorkspaceRoles
@@ -539,6 +540,7 @@ class WorkspaceCreate(object):
         self,
         label: str,
         description: str,
+        access_type: str = WorkspaceAccessType.CONFIDENTIAL.value,
         agenda_enabled: bool = True,
         public_upload_enabled: bool = True,
         public_download_enabled: bool = True,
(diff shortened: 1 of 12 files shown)


backend/tracim_backend/models/context_models.py
```diff
@@ -539,6 +540,7 @@ def __init__(
         self,
         label: str,
         description: str,
+        access_type: str = WorkspaceAccessType.CONFIDENTIAL.value,
```
grignards, line 543: This is funny to take a str argument (converted from `WorkspaceAccessType` for the default value) then build a `WorkspaceAccessType` in the `__init__`.

backend/tracim_backend/views/core_api/schemas.py
```diff
@@ -1026,6 +1033,11 @@ class WorkspaceMinimalSchema(marshmallow.Schema):
     workspace_id = marshmallow.fields.Int(example=4, validate=strictly_positive_int_validator)
     slug = StrippedString(example="intranet")
     label = StrippedString(example="Intranet")
+    access_type = StrippedString(
+        example=WorkspaceAccessType.CONFIDENTIAL.value,
+        validate=workspace_access_type_validator,
+        default=WorkspaceAccessType.CONFIDENTIAL.value,
```
grignards, line 1039: I would rather not leave a default here, the access type is as important as the label.

grignards (approved):

backend/tracim_backend/views/core_api/schemas.py
inkhey: i understand and this is not compilant with specification, see https://github.com/tracim/tracim/issues/3536#issuecomment-692596230 but such change break frontend, so if we are strict here, we do need to fix this in this PR or in a very soon PR, because this break frontend.

backend/tracim_backend/models/context_models.py
inkhey: yes, i don't know how to deal better with enum case where we:
- prefer using enum anywhere possible
- get string from the api itself.

Maybe with some marshmallow code ?
The problem is also that we do use custom validator which validate string and not "enum", which is nicely supported by hapic giving a list of valid value in autogenerated doc.

backend/tracim_backend/views/core_api/schemas.py
inkhey: fixed.

grignards:
Tested on branch.

Issue #3587, opened 2020-09-15
Feat: Allow user to join space


gbonaspetti:
## Feature description and goals

Users must have the option to join an **open** space and the option to ask to join an **on request** space. For this they will have a "join space" page that it's accessible by a button at the sidebar (see interface examples).

Some implementation/interface details:
- The style of the list is the same as in "All contents".

### Expected
- [x] All type icons have a tooltip with the type name
- [x] A big description should be chopped with ellipsis (...)

## Interfaces examples
_Obs: these interfaces are just a model. Things like the "Create space" button (that should have the same size of the Join button) will not be implemented in this issue._
- Sidebar button
![Capture d’écran de 2020-09-15 13-39-55](https://user-images.githubusercontent.com/35422692/93206442-b2139800-f759-11ea-9bd0-511191a85dbe.png)

- Join page
![ASF_join_space](https://user-images.githubusercontent.com/35422692/94577781-39dbc500-0277-11eb-8fb5-8cbddff986ed.png)

## Translations

|English|French|Portuguese|
|--|--|--|
|Join a space|Rejoindre un espace|Junte-se a um espaço|
|Title and description|Titre et description|Título e descrição|
|Access request|Demande d'accès|Pedido de acesso|
|Request access|Demander l'accès|Pedir acesso|
|Request sent|Demande envoyée|Pedido enviado|
|Join the space|Rejoindre l'espace|Junte-se ao espaço|
|Filter spaces|Filtrer les espaces|Filtrar espaços|
|You can join an existing space or create your first space|Vous pouvez rejoindre un espace existant ou créer votre premier espace|Você pode juntar-se a um espaço existente ou criar o seu primeiro espaço|
|You can create your first space|Vous pouvez créer votre premier espace|Você pode criar o seu primeiro espaço|
|You can join an existing space|Vous pouvez rejoindre un espace existant|Você pode juntar-se a um espaço já existente|
|Please ask access to your spaces to an administrator or space manager|Veuillez demander accès à vos espaces à un administrateur ou un gestionnaire d'espace|Por favor peça acesso aos seus espaços a um administrador ou gestor de espaço|

## Prerequisites
- https://github.com/tracim/tracim/issues/3538
- https://github.com/tracim/tracim/issues/3539

## Discussed before development 

1. The filter (search) should be done just for name or for the words in the description too?
Answer: The filtering should be on both and implemented in frontend only (https://github.com/tracim/tracim/issues/3587#issuecomment-692671610)

[...]

Issue #3540, opened 2020-09-10
Feat: API for previewing OPEN and ON_REQUEST spaces


grignards:
## Feature description and goals

`OPEN` and `ON_REQUEST` spaces (see #3536 for details) can be previewed by any user so that they can become a member.

- [x] Add an API GET `/api/users/<user_id>/accessible_workspaces`:
  - accessible by the current user and administrators
  - returns the list of accessible spaces: every space the user is not member of and have access type `ON_REQUEST` or `OPEN`, see #3536).
  - the returned space object should only include:
```json
{
  "workspace_id": 42,
  "label": "My workspace",
  "description": "A short description",
  "access_type": "open",
  "slug": "my-workspace"
}
```
NOTE: added `slug` and `sidebar_entries` as they are part of every workspace object in the API. Using `description` instead of `public_description` until we choose whether to do #3551 or not.

## Prerequisites

- #3536
- #3551

## Design alternatives

Another design would be to modify the semantics of `/api/users/<user_id>/workspaces` in order to return every workspace the user is member of/has access to. And add a filter `is_member`.

@grignards I chose not to change the semantics as modifying the existing is generally more difficult than to create an new API. Also the usefulness of changing the semantics wasn't an evidence for me.

inkhey:
Need probably some clarification about what informations we do provide to user about theses space.
Depending on the meaning we do have from "space descriptions", it should or shouldn't be publicly visible.
I do suggest to firstly do not leak description but a new "summary" public field may be a solution.

lebouquetin:
I agree with @inkhey : we should have a short public description which is different from the space description (which may be large and rich text through HTML)

inkhey:
issue created here about the public_description : https://github.com/tracim/tracim/issues/3551

lebouquetin:
@grignards I suggest `s/can be viewed/can be previewed` (or discovered) to be clear that it is only a limited access

raphj:
l like "discovered" (or "listed")?

PhilippeAccorsi:
@grignards title of issue is `Feat: Allow API access to OPEN and ON_REQUEST spaces` but if I understand correctly, `ON_REQUEST``is done in another issue. This is right ?

If yes can you modify title or add comment about that ?

grignards:
I think the `ON_REQUEST` in the title is OK: the goal of this issue was to provide an API for users to preview OPEN and ON_REQUEST spaces so that they can see some information about them before self-subscribing.
[...]

Issue #3601, opened 2020-09-16
Feat: Add a `default_user_role` attribute to spaces


grignards:
## Feature description and goals

The `OPEN` spaces require a default role for users who become members of a space by themselves/are added automatically.
Furthermore having a default role allows the user interface to pre-select this role when a user is added.

- [x] add a `default_user_role` attribute to `Workspace`
- [x] return it in all `workspaces` APIs and make it mandatory
- [x] create a migration script that sets this column to `reader` in existing rows.
- [x] the role will be required at space creation
- [x] adapt frontend to always set default_user_role at workspace creation and update : (PUT
`/api/workspaces/{workspace_id}` and `POST /api/workspaces/)

## Implemented solution:
https://github.com/tracim/tracim/pull/3615
- `default_user_role` field as enum in workspace table, default database value to reader.
- migration script to add default_user_role field in workspace table
- `default_user_role` added to `WorkspaceSchema`
- `default_user_role` available in workspace update endpoint
- `default_user_role` mandatory in workspace creation endpoint
- adaptation of frontend to not break the ui (as `default_user_role` is mandatory)

inkhey:
Is reader as default role for existing spaces OK?
- at database level, i do think yes, as existing workspace will be set to this, this seen to me both easier and logical.
- at api level, i don't know. should default_user_role being mandatory at workspace creation ? what should be the default in frontend ui ?

grignards:
> > Is reader as default role for existing spaces OK?
> 
>     * at database level, i do think yes, as existing workspace will be set to this, this seem to me both easier and logical.
> 
>     * at api level, i don't know. should default_user_role being mandatory at workspace creation ? what should be the default in frontend ui ?

I added the ui point to the corresponding frontend issue #3577.
As for the API… it's a bit like the `access_type`: not having a default is "better" because the backend cannot really guess what is best, but breaks the compatibility.

grignards:
> > > Is reader as default role for existing spaces OK?
> > 
> > 
> > ```
> > * at database level, i do think yes, as existing workspace will be set to this, this seem to me both easier and logical.
> > 
> > * at api level, i don't know. should default_user_role being mandatory at workspace creation ? what should be the default in frontend ui ?
> > ```
> 
> I added the ui point to the corresponding frontend issue #3577.
> As for the API… it's a bit like the `access_type`: not having a default is "better" because the backend cannot really guess what is best, but breaks the compatibility.

[...]

PR #3613, opened 2020-09-18
Feat/3540  accessible workspaces

asf-algoo:
See #3540.

## Checkpoints

**For developers**

**For code reviewers**

**For testers**

- [x] Manual, quality tests have been done: the frontend doesn't make use of this new API so we should better check this with #3587 PR.

Manual tests should aim to check the possible regressions as the workspaces API changed in this PR: try features that need access to workspaces information (dashboard/advanced dashboard/workspace administration/…).


backend/tracim_backend/lib/core/workspace.py
```diff
@@ -257,6 +256,25 @@ def get_all_for_user(
         query = query.order_by(Workspace.label)
         return query.all()
 
+    def get_all_accessible_for_user(self, user: User) -> typing.List[Workspace]:
+        """
+        Return workspaces accessible by user.
```
raphj, line 261: by or for, maybe pick one. I'd say "by"

backend/tracim_backend/tests/functional/test_user.py
```diff
@@ -2726,6 +2728,72 @@ def test_api__get_user_workspaces__err_400__user_does_not_exist(self, web_testap
...
+        user_api_factory,
+        workspace_api_factory,
+        workspace_access_type,
+        accessible_workspaces_count,
+        user_credentials,
+    ):
+        """
+        Check obtain all workspaces reachables for one user who does
+        not exist
+        with a correct user auth.
```
raphj, line 2752:
```suggestion
        Check that we get all workspaces reachable by one user who does
        not exist with a correct user auth.
```

backend/tracim_backend/views/core_api/user_controller.py
```diff
@@ -780,6 +782,27 @@ def set_user_config(self, context, request: TracimRequest, hapic_data: HapicData
...
 
+    @hapic.with_api_doc(tags=[SWAGGER_TAG__USER_CONFIG_ENDPOINTS])
+    @check_right(has_personal_access)
+    @hapic.input_path(UserIdPathSchema())
+    @hapic.output_body(WorkspaceMinimalSchema(many=True))
+    def get_accessible_workspaces(
+        self, context, request: TracimRequest, hapic_data: HapicData
+    ) -> typing.List[WorkspaceInContext]:
+        """
+        Return the list of accessible workspace for the given user id.
```
raphj, line 793:
```suggestion
        Return the list of accessible workspaces by the given user id.
```

raphj (approved):

backend/tracim_backend/lib/core/workspace.py
```diff
@@ -257,6 +256,25 @@ def get_all_for_user(
...
+        Accessible workspaces:
+          - are of type OPEN or ON_REQUEST
+          - do not have user as a member
+        """
+        query = self._base_query_without_roles().filter(
```
inkhey, line 266: Can you add just a docstring to `_base_query_without_roles` method or rename it ? just checked, don't find this really explicit.

backend/tracim_backend/views/core_api/schemas.py
```diff
@@ -1031,6 +1031,7 @@ class Meta:
 
 class WorkspaceMinimalSchema(marshmallow.Schema):
     workspace_id = marshmallow.fields.Int(example=4, validate=strictly_positive_int_validator)
+    description = StrippedString(example="All intranet data.")
```
inkhey, line 1034: what the reason behind this change ? WorkspaceMinimalSchema is not really minimal anymore.

grignards: I checked its usage, it is used only in `users/…/workspaces` and I think both APIs should return the same content. Also we already have 3 different public representations of a workspace (minimal,digest,full) adding one would make the usage of the APIs even more difficult (many `workspaces` endpoint but not two return the same structure).

backend/tracim_backend/lib/core/workspace.py
grignards: Yes that's a good idea, I will do it.

backend/tracim_backend/lib/core/workspace.py
grignards: I agree, but several other methods use `_for` so I preferred consistency over rightness.

backend/tracim_backend/views/core_api/schemas.py
grignards: I also saw that the `/api/workspaces` endoint doesn't use the digest representation which surprised me a bit.

inkhey: https://github.com/tracim/tracim/issues/3601#issuecomment-694872894
Current usage of schema:
```
WorkspaceMinimalSchema:
- GET /workspaces/{workspace_id}/disk_space as subitem
WorkspaceDigestSchema :
- GET /users/{user_id:\d+}/workspaces
- GET /workspaces/{workspace_id}/members as subitem
- GET /workspaces/{workspace_id}/members/{user_id} as subitem
- PUT /workspaces/{workspace_id}/members/{user_id} as subitem
WorkspaceSchema:
- GET /workspaces
- POST /workspaces
- PUT /workspaces/{workspace_id} 
- GET /workspaces/{workspace_id}
- TLM
```

I don't see any advantage to get description in
```
- GET /workspaces/{workspace_id}/disk_space as subitem
- GET /workspaces/{workspace_id}/members as subitem
- GET /workspaces/{workspace_id}/members/{user_id} as subitem
- PUT /workspaces/{workspace_id}/members/{user_id} as subitem
```

Having description in
```
- GET /users/{user_id:\d+}/workspaces
```
may be not a big deal, has user has already access, and make things more consistent with `GET /workspaces`

So, i suggest:
```
WorkspaceMinimalSchema: same as actual
- GET /workspaces/{workspace_id}/disk_space as subitem
WorkspaceDigestSchema : all but sidebars entries
- GET /workspaces/{workspace_id}/members as subitem
- GET /workspaces/{workspace_id}/members/{user_id} as subitem
- PUT /workspaces/{workspace_id}/members/{user_id} as subitem
WorkspaceSchema:
- GET /users/{user_id:\d+}/workspaces
- GET /api/users/<user_id>/accessible_workspaces
- GET /workspaces
- POST /workspaces
- PUT /workspaces/{workspace_id} 
- GET /workspaces/{workspace_id}
- TLM
```

raphj:
tested on branch

backend/tracim_backend/views/core_api/schemas.py
inkhey: @grignards can we clarify this point, you instead move description to DigestSchema, which mean description now added to :
```
- GET /workspaces/{workspace_id}/members as subitem
- GET /workspaces/{workspace_id}/members/{user_id} as subitem
- PUT /workspaces/{workspace_id}/members/{user_id} as subitem
```
which is not needed.

grignards (commented):
I can return `WorkspaceSchema` in `users/<user_id>/(workspaces,accessible_workspaces) but in your proposal you remove `sidebar_entries` from `Digest` schema, are we sure that the frontend doesn't use this attribute from the routes using the `DigestSchema`? We should check what they are using from this route, if the `MinimalSchema` is enough for them we could remove/rename `Digest` to only keep two schemas for workspaces.

backend/tracim_backend/views/core_api/schemas.py
grignards: I can return `WorkspaceSchema` in `users/<user_id>/(workspaces,accessible_workspaces) but in your proposal you remove `sidebar_entries` from `Digest` schema, are we sure that the frontend doesn't use this attribute from the routes using the `DigestSchema`? We should check what they are using from this route, if the `MinimalSchema` is enough for them we could remove/rename `Digest` to only keep two schemas for workspaces.

inkhey:
> _Moved to the right discussion_


> Other possibility: only use the digest schema everywhere. Was there a special requirement to use a different representation in users/<user_id>/workspaces? Or was it only "the frontend does not need all digest information"?

I guess that frontend take advantage of normal WorkspaceSchema when user is admin using this endpoint of `GET /workspaces` and that's probably why there is different schema between this one and user workspace.

In your specific case you need both description which normally does not make lot of sense in digest schema but not all the workspace information, so, in this case, i think we should go to WorkspaceSchema almost everwhere and use digest only to not provide sidebar entries and maybe description to for specific case (subitem).

backend/tracim_backend/views/core_api/schemas.py
inkhey: Yes we are sure because members schema does not contain sidebar_entries:
```
class WorkspaceMemberSchema(WorkspaceMemberDigestSchema):
    user_id = marshmallow.fields.Int(example=3, validate=strictly_positive_int_validator)
    workspace_id = marshmallow.fields.Int(example=4, validate=strictly_positive_int_validator)
    is_active = marshmallow.fields.Bool()
    user = marshmallow.fields.Nested(UserDigestSchema())
    workspace = marshmallow.fields.Nested(WorkspaceDigestSchema(exclude=("sidebar_entries",)))

    class Meta:
        description = "Workspace Member information"
```

Also, my suggestion say Digest should contain all but sidebar entries, but in fact we can just remove sidebar entries from current digest and keep it like this.

But you're right, it may probably more sense to check what is useful for workspace subitem in workspace members endpoint in frontend.

So i suggest check usage of workspace subitem in sidebar entry and merge digest/minimal schema if possible (and do the other change i suggest to keep consistency).

raphj (approved):

raphj:
We don't need a new manual test do we @grignards?

grignards:
> We don't need a new manual test do we @grignards?

I would recommend some generaI tests like I put in the PR's description, following the changes on workspaces schemas.

inkhey:
tested on branch.

Issue #3631, opened 2020-09-23
Bug: escape to close the mention suggestions does not work when writing a non-advanced comment


raphj:
- [x] escape should close the mentions popup

PhilippeAccorsi:
Verified on preprod.

Issue #3630, opened 2020-09-23
Feat: start showing members in the mention suggestions right away


raphj:
When a @ is written, the popup should open not only with the group mentions but also with the list of all space members immediately and not after two characters are typed.

- [x] When autocompleting mentions (that is, listing only members from the current workspace), start sending the list of members right away.
- [x] add a `limit` parameter to the known members endpoint
- [x] Limit the returned list to `limit` elements, if provided

## Discussed before development

- What should be the maximum number of returned members?
- Should this be a parameter the frontend sends? 

(taken from #3544)

___
2020-09-24: solved the discussions

gbonaspetti:
> To be discussed before development
What should be the maximum number of returned members?
Should this be a parameter the frontend sends?

To have the same number of results everywhere I would say to use `NUMBER_RESULTS_BY_PAGE` constant (`frontend/src/util/helper.js`, for now)

raphj:
ok :-) good idea
So since there is already a constant in the frontend, I'd say let's allow the frontend to set the maximum when doing the request to the backend.

grignards:
Just a small remark for `max_count`: check if there is the same behavior in another API. If there is, you should use the same naming.

raphj:
There is indeed a case where this parameter is named limit. I'm changing this.

PhilippeAccorsi:
What is the default limit for `Limit the returned list to limit elements, if provided` ?

raphj:
The limit provided by the frontend is 15

PhilippeAccorsi:
Thanks.
Tested on preprod.

PR #3634, opened 2020-09-24
Feat/3630  immediately show members in the autosuggestion and fix escape key

raphj:
#3630, #3631 

## Checkpoints

**For developers**

- [X] The original issues are up to date w.r.t the latest discussions and contains a short summary of the implemented solution

**For code reviewers**

**For testers**

Manual, quality tests have will be done in maintenance/3.1


backend/tracim_backend/lib/core/user.py
```diff
@@ -201,15 +201,44 @@ def get_known_users(
...
+
+        if include_workspace_ids:
+            user_in_every_included_workspaces = True
+
+            include_user_ids = []
+            for workspace_id in include_workspace_ids:
+                user_ids = self.get_members_of_workspaces([workspace_id])
+                include_user_ids += user_ids
+                if user_in_every_included_workspaces:
+                    user_in_every_included_workspaces = self._user.user_id in user_ids
```
grignards, line 232: I'm afraid this will be a bit slow if many workspace ids are given. Wouldn't a single query on `UserRoleInWorkspace` work?
```python
query(UserRoleInWorkspace).filter(UserRoleInWorkspace.user_id==self._user.user_id).filter(UserRoleInWorkspace.workspace_id.in_(include_workspace_ids)).count()
```
then compare the count with the count of workspaces?

backend/tracim_backend/lib/core/user.py
```diff
@@ -201,15 +201,44 @@ def get_known_users(
...
-        if len(acp) < 2:
+        nb_elems = KNOWN_MEMBERS_ITEMS_LIMIT
+
+        # RJ - 2020-09-14-NOTE
+        # This is a bit convoluted. Keep in mind that we want to allow:
+        #  - listing a number of members greater than KNOWN_MEMBERS_ITEMS_LIMIT
+        #  - autocomplete less than 2 characters
+        # only if the requesting user is in each and every included workspace.
+        # Otherwise, we don't want to allow that.
+        # By default, we return a maximum of KNOWN_MEMBERS_ITEMS_LIMIT members (when limit is not set),
```
grignards, line 219: Even with your note the behavior of this method is rather unclear.
If you have the use cases of this method in mind, could you write an issue to list them so that we can think of a better answer that only "one method to rule them all"? And link the issue here?

backend/tracim_backend/views/core_api/schemas.py
```diff
@@ -682,6 +679,14 @@ class KnownMembersQuerySchema(marshmallow.Schema):
         description="comma separated list of included workspaces: members of this workspace are excluded from the result; cannot be used with exclude_workspace_ids",
     )
 
+    limit = marshmallow.fields.Int(
+        example=15,
+        default=0,
+        allow_none=True,
```
grignards, line 685: Is `allow_none` necessary here?

frontend_lib/src/tinymceAutoCompleteHelper.js
```diff
@@ -2,6 +2,8 @@ const MENTION_AUTOCOMPLETE_REGEX = /(?:^|\s)@([a-zA-Z\-_]*)$/
 
 const USERNAME_ALLOWED_CHARACTERS_REGEX = /[a-zA-Z\-_]/
 
+let lastSelAndOffset = null
```
grignards, line 5: Isn't it `previousSelAndOffset` closer to the usage of this variable?

grignards (changes requested):

backend/tracim_backend/app_models/validator.py
```diff
@@ -87,7 +87,6 @@ def validate_all(self) -> bool:
 # string matching list of string (without',') separated by ','
 regex_string_as_list_of_string = Regexp(regex=(re.compile("^([^,]+(,[^,]+)*)?$")))  # noqa: W605
 
-acp_validator = Length(min=2)
```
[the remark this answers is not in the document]
raphj: yes. Note that I'm not introducing anything new, and this is a line I remove.

backend/tracim_backend/lib/core/user.py
raphj: Good idea

frontend_lib/src/action.async.js
```diff
@@ -76,8 +76,9 @@ export const putContentRestoreDelete = (apiUrl, workspaceId, contentId) => {
   })
 }
 
-export const getMyselfKnownMember = (apiUrl, userNameToSearch, workspaceIdToInclude) => {
-  return fetch(`${apiUrl}/users/me/known_members?acp=${userNameToSearch}&include_workspace_ids=${workspaceIdToInclude}`, {
+export const getMyselfKnownMember = (apiUrl, userNameToSearch, workspaceIdToInclude, limit = 0) => {
+  const limitParam = limit ? '&limit=' + limit : ''
+  return fetch(`${apiUrl}/users/me/known_members?acp=${userNameToSearch}&include_workspace_ids=${workspaceIdToInclude}` + limitParam, {
```
gbonaspetti, line 81:
```suggestion
  return fetch(`${apiUrl}/users/me/known_members?acp=${userNameToSearch}&include_workspace_ids=${workspaceIdToInclude}${limitParam}`, {
```

frontend_lib/src/action.async.js
```diff
@@ -76,8 +76,9 @@ export const putContentRestoreDelete = (apiUrl, workspaceId, contentId) => {
   })
 }
 
-export const getMyselfKnownMember = (apiUrl, userNameToSearch, workspaceIdToInclude) => {
-  return fetch(`${apiUrl}/users/me/known_members?acp=${userNameToSearch}&include_workspace_ids=${workspaceIdToInclude}`, {
+export const getMyselfKnownMember = (apiUrl, userNameToSearch, workspaceIdToInclude, limit = 0) => {
+  const limitParam = limit ? '&limit=' + limit : ''
```
gbonaspetti, line 80:
```suggestion
  const limitParam = limit ? `&limit=${limit}` : ''
```

frontend_lib/src/tinymceAutoCompleteHelper.js
```diff
@@ -10,15 +12,31 @@ const seekUsernameEnd = (text, offset) => {
...
+    offset: sel.anchorOffset
+  }
 }
 
 export const tinymceAutoCompleteHandleInput = (e, setState, fetchMentionList, isAutoCompleteActivated) => {
-  if (MENTION_AUTOCOMPLETE_REGEX.test(getTextOnCursor())) {
+  const selAndOffset = getSelAndOffset()
+
+  if (lastSelAndOffset && lastSelAndOffset.text === selAndOffset.text && lastSelAndOffset.offset === selAndOffset.offset) {
+    // RJ - 2020-09-14 - NOTE: handleInput is called twice after typing a key because the selection also changes.
```
gbonaspetti, line 32:
```suggestion
    // INFO - RJ - 2020-09-14 - handleInput is called twice after typing a key because the selection also changes.
```

gbonaspetti (changes requested):

backend/tracim_backend/lib/core/user.py
raphj: After discussion, we decided not to do this for now: we plan to refactor this method. Moreover, in the current use case, only one space is ever listed in include_workspaces, so this would add a query and make the method slower instead of making it faster.

frontend_lib/src/action.async.js
raphj: Not a big fan of this, but committing your suggestion anyway (no strong opinion).

gbonaspetti (approved):

raphj:
There was two failing tests:
 - one that seems to be a random failure. It consistently works locally. I am Leaving it as is without disabling it and we'll see tomorrow.
 - one related to tinymce and a weird behavior of Cypress. I don't understand what's happening and why this test started to consistently fail, especially that I add the opposite issue when writing it. I had to change `.type(' @johndoe')` to `.type(' ') .type('@johndoe')` - and these was not working before this PR and the other version, yes. But at least, now, the test passes.

I am merging this so the CI can run on maintenance/3.1 with all the planned changes.

Issue #3549, opened 2020-09-11
Feat: Support child spaces in WebDAV view


grignards:
## Feature description and goals

The child spaces (see #3546) introduce new levels of "folders" in WebDAV, we have to take care those work properly:

- [x] the sub-spaces shall be displayed as folders (check if it is possible to distinguish them from folders which are contents)
- [x] check that sub-space move is properly disabled
- [x] sub-spaces shall have a `.space` extension to avoid name collision with folders

## Prerequisites

- #3546
- #3625

### How do we handle the possibility of having a folder+sub-space with the same name ?

Seen 4 possibilities:

- forbid having same name between folder and space.
- ~~add a suffix in case of conflict (for example `~s<workspace_id>` and `~f<content_id>`): a bit like space naming conflict (see #2477)~~ deprecated as we do decide to remove the support for same id space, see #3625 
- add an extension to spaces (`.space`): easy but could disturb users
- re-organize spaces in webdav in order to separate sub-spaces and folders. This would require a reserved space name (eg. `All Contents`).

**Decision**: we'll go with the `.space` extension.

## Implemented solution

https://github.com/tracim/tracim/pull/3687

Features:
- add filemanager_filename property for workspace to access to webdav showed version of workspace and use it in order that all workspace end with this extension in webdav view
- adapt workspace to be able to show both subworkspace and content, workspace have priority so if both workspace and folder exist with same webdav name (if you create dir with .space extension) it means workspace is prioritary and you will not be able to see the folder. It' s a similar issue as same workspace name for one user.
- adapt workspace to show orphan workspace (workspace with parent not accessible) at root.

Refactoring/Tests:
- remove some unused code like get_folder_with_workspace_path_labels or filter_query_for_content_label_as_path methods
- adapt test with the new space extension.
- add WebdavPath object helping process path hierarchy for webdav instead of old static method.
- add few tests to check if children and orphans workspaces are accessible as expected.

- Refactor Webdav ressource in order to avoid confusing inheritance by using composition with WebdavContainer class.

inkhey:
just found 2 things:
- creating a folder in space will create a folder.
- we can have issue with same name for folder and subspace. in fact we do allow same name space as far i know. This can cause issue in some file manager

inkhey:
> just found 2 things:
> 
>     * creating a folder in space will create a folder.
> 
>     * we can have issue with same name for folder and subspace. in fact we do allow same name space as far i know. This can cause issue in some file manager

[...]

Issue #3579, opened 2020-09-14
Feat: Display child spaces and their access types in the sidebar


grignards:
## Feature description and goals

Child spaces (see #3546) must be displayed in the sidebar as a tree:

![sidebar](https://user-images.githubusercontent.com/35422692/94559597-c54a5b80-0261-11eb-9793-21be0e0d6267.png)

- [x] All spaces the user is member of shall be displayed in the sidebar
- [x] All tree items are always unfolded (so that all child spaces are visible)
- [x] Scrollbar handling stays the same as currently
- [x] Clicking on a space item shall open the space's dashboard
- [x] The current space shall be displayed in bold+underline:

![highlight](https://user-images.githubusercontent.com/35422692/94573017-f6328c80-0271-11eb-9bb8-4519d396eb09.png)

- [x] It is possible for a user to have access to a sub-space without having access to its parent. In this case the sub-space shall be treated as a root space for displaying it.
- [x] The "drag-and-drop" icon that was displayed when we moved some content will be displayed at the left corner, this will "move" the space's title, but it's an acceptable temporary solution until we think in a better way of show it.

## Prerequisites

- #3546 (merged)
- https://github.com/tracim/tracim/issues/3584 (merged)

### Finalize the access type icons

_2020-09-29_ **Decision**: We are not going to display the type icons in the sidebar.

### Should the tree structure be built by the frontend or by the backend?

Building the tree structure in the frontend is possible (since the API will provide a `parent_id` filter), but less efficient than a backend approach (the frontend will need *n*+1 queries where *n* is the number of spaces the user is member of).
A backend approach is more efficient (a single SQL query could do the job) but is more "specialized".
The code complexity should be similar for both solutions.

**Decision**: A last approach (and the one we chose) is for the backend to return all spaces a user has access to (with the `parent_id` information in them), then the frontend builds its tree from this list.

raphj:
> the frontend will need n+1 queries
What to you mean? It will need to fetch data about each space via HTTP?

Then I'm for the creation of an endpoint that gives the tree with all the needed data so the frontend can fetch all the data in one request.

But can't the backend just send the list of the spaces of which the user is member, with their parent id, in one go, with the existing endpoint? If so, I think building the tree should be done by the frontend, because showing them as a tree is a UI decision.

inkhey:
> > the frontend will need n+1 queries
> 
> What to you mean? It will need to fetch data about each space via HTTP?
> 
[...]

Issue #2024, opened 2019-06-28
Use last version of Wsgidav


inkhey:
## Bug description & symptom

We are currently using wsgidav 2.X because new wsgidav 3.X break our code

## Expected

Change code in order to rely on wsgidav upstream version.

lebouquetin:
Do you have any idea of what would be broken with wisgidav 3.x?

inkhey:
> Do you have any idea of what would be broken with wisgidav 3.x? @inkhey

I have rapidly tested last year, and many signatures have changed, but that not all. 
Breaking change came from the version 3 : https://github.com/mar10/wsgidav/blob/master/CHANGELOG.md#300--2019-03-04

PR #3676, opened 2020-10-08
Feature/3549  support child space in webdav

inkhey:
related to #3549 

## Checkpoints

**For developers**

**For code reviewers**

**For testers**


inkhey:
I suggest waiting for https://github.com/tracim/tracim/issues/3579 to test this fully, as we can:
- check consistency of workspace view between webdav an frontend.
- easily create hierarchy of workspace without setting value in database to test

All workspaces features should be tested as this PR does big change in the webdav process.
The new things to test:
- test sub-workspace access
- tests move/copy content between hierarchical workspace.
- test orphans workspaces access (workspace with parent where user is not members)
- test that all workspace end with .space
- test workspace over content priority if same name.

backend/tracim_backend/lib/core/workspace.py
```diff
@@ -218,16 +218,40 @@ def get_one(self, workspace_id: int) -> Workspace:
                 "workspace {} does not exist or not visible for user".format(workspace_id)
             ) from exc
 
-    def get_one_by_label(self, label: str) -> Workspace:
+    def get_one_by_filemanager_filename(
+        self, filemanager_filename: str, parent: typing.Optional[Workspace] = None
+    ) -> Workspace:
         """
         get workspace according to label given, if multiple workspace have
         same label, return first one found.
```
grignards, line 226: The docstring would gain to be updated, especially explaining what is a `filemanager_filename`.

backend/tracim_backend/lib/core/workspace.py
```diff
@@ -317,6 +341,19 @@ def get_all_for_user(
         query = query.order_by(Workspace.label)
         return query.all()
 
+    def get_user_orphan_workspaces(self, user: User):
+        """Get all user workspace where user is not member of the parent and parent does exist"""
+        query = self._base_query()
+        workspace_ids = []
+        rapi = RoleApi(session=self._session, current_user=self._user, config=self._config)
+        workspace_ids.extend(
```
grignards, line 349:
```suggestion
        workspace_ids =  rapi.get_user_workspaces_ids(user_id=user.user_id, min_role=UserRoleInWorkspace.READER)
```

backend/tracim_backend/lib/webdav/dav_provider.py
```diff
@@ -1,33 +1,140 @@
...
+
+    def __init__(self, path: str, current_user: User, session: TracimSession, app_config: CFG):
+        self.path = path
+        self.workspace_api = WorkspaceApi(
+            current_user=current_user, session=session, config=app_config
+        )
+        self.content_api = ContentApi(current_user=current_user, session=session, config=app_config)
+
+        self.workspaces = deque()
+        self.contents = deque()
```
grignards, line 52: Why do you use a `deque` here? Wouldn't a list be enough?

backend/tracim_backend/lib/core/workspace.py
```diff
@@ -218,16 +218,40 @@ def get_one(self, workspace_id: int) -> Workspace:
...
-    def get_one_by_label(self, label: str) -> Workspace:
+    def get_one_by_filemanager_filename(
+        self, filemanager_filename: str, parent: typing.Optional[Workspace] = None
+    ) -> Workspace:
         """
         get workspace according to label given, if multiple workspace have
         same label, return first one found.
         """
         # INFO - G.M - 2019-10-10 - result should be ordered same way as get_all() method,
         # to unsure working
```
raphj, line 229: I don't understand this comment

backend/tracim_backend/lib/core/workspace.py
```diff
@@ -317,6 +341,19 @@ def get_all_for_user(
         query = query.order_by(Workspace.label)
         return query.all()
 
+    def get_user_orphan_workspaces(self, user: User):
+        """Get all user workspace where user is not member of the parent and parent does exist"""
```
raphj, line 345:
```suggestion
        """Get all user workspaces where the users is not member of the parent and parent exists"""
```

backend/tracim_backend/lib/webdav/dav_provider.py
```diff
@@ -1,33 +1,140 @@
...
+    WORKSPACE = "workspace"
+    CONTENT = "content"
+
+
+class ProcessedWebdavPath(object):
+    """
+    Processor for Webdav Path:
+    - get all workspaces hierarchy from path
+    - get all content hierarchy from path
+    - provide useful properties to handle Webdav Request
```
raphj, line 41:
```suggestion
    - get the workspace hierarchy from path
    - get the content hierarchy from path
    - provide useful properties to handle the WebDAV request 
```
?

backend/tracim_backend/lib/webdav/dav_provider.py
```diff
@@ -38,9 +145,16 @@ def __init__(
...
-        self.path = path
+        self.processed_path = ProcessedWebdavPath(
+            path=path,
+            current_user=self.current_user,
+            session=self.dbsession,
+            app_config=self.app_config,
+        )
 
     @property
     def root_path(self) -> str:
```
raphj, line 160:
```suggestion
    def root_path(self) -> typing.Optional[str]:
```

backend/tracim_backend/lib/webdav/dav_provider.py
```diff
@@ -89,84 +203,34 @@ def current_workspace(self):
...
-            show_deleted=True,
-        )
-        return wapi.get_one_by_label(workspace_id)
-
-    def _get_current_workspace_label(self) -> str:
-        return webdav_convert_file_name_to_bdd(self.path.split("/")[1])
+        return self.processed_path.current_workspace
 
     @property
     def current_content(self):
```
raphj, line 209: Can you put the typing here?
(it is typing.Optional[X], with some X, if I understand the code correctly)

backend/tracim_backend/lib/webdav/dav_provider.py
```diff
@@ -89,84 +203,34 @@ def current_workspace(self):
...
-        return webdav_convert_file_name_to_bdd(self._destpath.split("/")[1])
+        self.processed_destpath = ProcessedWebdavPath(
+            path=destpath,
+            current_user=self.current_user,
+            session=self.dbsession,
+            app_config=self.app_config,
+        )
 
     @property
     def candidate_parent_content(self) -> Content:
```
raphj, line 225:
```suggestion
    def candidate_parent_content(self) -> typing.Optional[Content]:
```
?

backend/tracim_backend/lib/webdav/dav_provider.py
```diff
@@ -188,74 +252,32 @@ def getResourceInst(self, path: str, environ: dict):
...
             )
-        elif content.type == content_type_list.File.slug:
-            return resources.FileResource(
-                path=path, environ=environ, content=content, tracim_context=tracim_context,
-            )
-        else:
-            return resources.OtherFileResource(
-                path=path, environ=environ, content=content, tracim_context=tracim_context,
-            )
+        return None
```
raphj, line 280: Can you add the typing information on this method?

backend/tracim_backend/models/data.py
```diff
@@ -158,6 +158,18 @@ def get_children(self, recursively: bool = False) -> List["Workspace"]:
         else:
             return self.children
 
+    @property
+    def filemanager_filename(self):
```
raphj, line 162:
```suggestion
    def filemanager_filename(self) -> str:
```

backend/tracim_backend/lib/webdav/resources.py
```diff
@@ -68,143 +70,224 @@ def wrapper(self: "_DAVResource", *arg, **kwarg) -> typing.Callable:
...
+def get_content_resource(
+    content: Content,
+    path: str,
+    environ: typing.Dict,
+    workspace: Workspace,
+    tracim_context: "WebdavTracimContext",
+) -> _DAVResource:
     """
-    RootResource ressource that represents tracim's home, which contains all workspaces
+    Helper to get correct content webdav resource according to content type
```
raphj, line 93:
```suggestion
    Helper to get the correct content WebDAV resource according to the content type
```

backend/tracim_backend/lib/webdav/dav_provider.py
```diff
@@ -1,33 +1,140 @@
...
+                            pass
+                    try:
+                        content = self.content_api.get_one_by_filename(
+                            filename=section,
+                            workspace=self.workspaces[-1],
+                            parent=self.contents[-1] if self.contents else None,
+                        )
+                    except ContentNotFound:
+                        content = None
+                    self.contents.append(content)
```
grignards, line 86: Wouldn't separating the loop in two parts (one for workspaces, one for contents) be easier to understand?
Something like this:
```python
path_parts = self._path_splitter(self.path)
if not path_parts:
    self.workspaces.append(None)
    return
workspace_found = True
# INFO - G.M - 2020-10-07 - Get all workspace tree + content_tree
while workspace_found:
    try:
        part = path_parts[0]
        filemanager_filename = webdav_convert_file_name_to_bdd(part)
        self.workspaces.append(
            self.workspace_api.get_one_by_filemanager_filename(
                root_workspace_filemanager_filename
            )
        )
        path_parts = path_parts[1:]
    except (IndexError, WorkspaceNotFound):
        workspace_found = False

for part in path_parts:
    try:
        content = self.content_api.get_one_by_filename(
            filename=section,
            workspace=self.workspaces[-1],
            parent=self.contents[-1] if self.contents else None,
        )
    except ContentNotFound:
        content = None
    self.contents.append(content)
```

backend/tracim_backend/lib/webdav/resources.py
```diff
@@ -68,143 +70,224 @@ def wrapper(self: "_DAVResource", *arg, **kwarg) -> typing.Callable:
...
         tracim_context: "WebdavTracimContext",
     ) -> None:
-        super(AbstractContentContainer, self).__init__(path, environ)
+        """
+        Some rules:
+        - if content given is None, return workspace root contents
+        - if content given is correct, return subcontent of this content
+         add user known workspaces without any user known parent to the list.
+         - in case of content collision, only the first named content (sorted by content_id
+         from lower to higher) will be returned.
```
raphj, line 280:
```suggestion
        - if content given is None, return workspace root contents
        - if the given content is correct, return the subcontent of this content
         and user-known workspaces without any user-known parent to the list.
         - in case of content collision, only the first named content (sorted by content_id
         from lower to higher) will be returned.
```

backend/tracim_backend/lib/webdav/resources.py
```diff
@@ -68,143 +70,224 @@ def wrapper(self: "_DAVResource", *arg, **kwarg) -> typing.Callable:
...
     def __init__(
         self,
         path: str,
         environ: dict,
         label: str,
+        content: Content,
+        provider: "TracimDavProvider",
         workspace: Workspace,
         tracim_context: "WebdavTracimContext",
     ) -> None:
```
raphj, line 273:
```suggestion
    ):
```

backend/tracim_backend/lib/webdav/dav_provider.py
```diff
@@ -1,33 +1,140 @@
...
+        return the level type of the current path:
+        - return ResourceLevelType.ROOT is path is root path
+        - return ResourceLevelType.WORKSPACE if path is at workspace level (can be direct
+        workspace root or subworkspace root)
+        - return ResourceLevelType.CONTENT if path is at content level (can be direct content at workspace
+        root or content within directory hierarchy)
+        - Return None if path is not valid
+        """
+        if self.path == "/":
+            return ResourceLevelType.ROOT
```
grignards, line 130: Maybe only use `self.workspaces`/`self.contents`:
```python
type = None
if not self.current_workspace:
   type = ResourceLevelType.ROOT
elif not self.contents:
   type = ResourceLevelType.WORKSPACE
elif self.current_content:
   type = ResourceLevelType.CONTENT
return type
```

backend/tracim_backend/lib/webdav/resources.py
```diff
@@ -68,143 +70,224 @@ def wrapper(self: "_DAVResource", *arg, **kwarg) -> typing.Callable:
...
+        return OtherFileResource(
+            path=path, environ=environ, content=content, tracim_context=tracim_context
+        )
 
-    def __init__(self, path: str, environ: dict, tracim_context: "WebdavTracimContext"):
-        super(RootResource, self).__init__(path, environ)
+
+class WebdavContainer(ABC):
+    """
+    A Webdav Container should implement theses methods.
```
raphj, line 115:
```suggestion
    A Webdav Container should implement these methods.
```

"These" is always spelled "these"

backend/tracim_backend/lib/webdav/resources.py
```diff
@@ -68,143 +70,224 @@ def wrapper(self: "_DAVResource", *arg, **kwarg) -> typing.Callable:
...
+    def getMemberNames(self) -> [str]:
+        pass
+
+    @abstractmethod
+    def getMember(self, label: str) -> _DAVResource:
+        pass
+
+    @abstractmethod
+    def getMemberList(self) -> [_DAVResource]:
+        pass
```
raphj, line 136: Are you free to name these methods as you wish?

If so:
 - "member" is confusing with respect to what is usually a member in Tracim 
 - maybe use snake_case?

backend/tracim_backend/lib/webdav/resources.py
```diff
@@ -68,143 +70,224 @@ def wrapper(self: "_DAVResource", *arg, **kwarg) -> typing.Callable:
...
+    def __init__(
+        self,
+        path: str,
+        environ: dict,
+        label: str,
+        workspace: typing.Optional[Workspace],
+        provider: "TracimDavProvider",
+        tracim_context: "WebdavTracimContext",
+        list_orphan_workspaces: bool = False,
+    ) -> None:
```
raphj, line 153:
```suggestion
    ):
```
A constructor should never return anything else than None anyway.

backend/tracim_backend/lib/webdav/dav_provider.py
```diff
@@ -1,33 +1,140 @@
...
+
+    @property
+    def current_parent_content(self) -> typing.Optional[Content]:
+        try:
+            return self.contents[-2]
+        except IndexError:
+            return None
+
+    @property
+    def current_path_level(self) -> typing.Optional[ResourceLevelType]:
```
grignards, line 119: `level` made me think it would be the depth of the current path (ie. an integer), should the naming be `current_path_type`/`current_path_resource_type`?

backend/tracim_backend/lib/webdav/resources.py
```diff
@@ -68,143 +70,224 @@ def wrapper(self: "_DAVResource", *arg, **kwarg) -> typing.Callable:
...
+        """
+        Some rules:
+        - if workspace given is None, return workspaces with no parent
+        - if workspace given is correct, return
+        - if list_orphan_workspaces is True, it
+         add user known workspaces without any user known parent to the list.
+         - in case of workspace collision, only the first named workspace (sorted by workspace_id
+         from lower to higher) will be returned
```
raphj, line 161:
```suggestion
        - if the given workspace is None, return workspaces with no parent
        - if the given workspace is correct, return [WHAT - feel free to edit my suggestion to replace this]
        - if list_orphan_workspaces is True, it
         adds user-known workspaces without any user-known parent to the list.
         - in case of workspace collision, only the first named workspace (sorted by workspace_id
         from lower to higher) will be returned
```
This list of rules seems similar to another one present in this PR, would it be possible to put these all at a same place and refer to them instead? (maybe it's not a good idea, I don't know)

backend/tracim_backend/lib/webdav/resources.py
```diff
@@ -68,143 +70,224 @@ def wrapper(self: "_DAVResource", *arg, **kwarg) -> typing.Callable:
...
+        raise NotImplementedError()
+
+    def getMemberNames(self) -> List[str]:
         """
-        raise DAVError(HTTP_FORBIDDEN, contextinfo="Not allowed to create new root")
+        This method returns the names (here workspace's labels) of all its children
 
-    @webdav_check_right(is_trusted_user)
-    def createCollection(self, name: str):
+        Though for perfomance issue, we're not using this function anymore
```
raphj, line 227:
```suggestion
        This method returns the names (here, the labels of workspaces) of all its children.
        For performance concerns, we're not using this function anymore
```

Can we remove it then? And from the interface defined earlier too?

backend/tracim_backend/lib/webdav/resources.py
```diff
@@ -284,84 +408,171 @@ def getMemberNames(self) -> [str]:
...
+    def __init__(
+        self,
+        path: str,
+        environ: dict,
+        label: str,
+        content: typing.Optional[Content],
+        workspace: typing.Optional[Workspace],
+        provider: "TracimDavProvider",
+        tracim_context: "WebdavTracimContext",
+    ) -> None:
```
raphj, line 445:
```suggestion
    ):
```

backend/tracim_backend/lib/webdav/resources.py
```diff
@@ -284,84 +408,171 @@ def getMemberNames(self) -> [str]:
...
+                continue
+            else:
+                members.append(content_resource)
         return members
 
 
-class WorkspaceResource(AbstractContentContainer, DAVCollection):
+class RootResource(DAVCollection):
+    """
+    RootResource ressource that represents tracim's home, which contains all workspaces without
```
raphj, line 515:
```suggestion
    RootResource resource that represents Tracim's home, which contains all workspaces without
```

raphj (changes requested):

backend/tracim_backend/lib/webdav/resources.py
grignards, line 136: Maybe add documentation to explain that members here are sub-resources of this object.

backend/tracim_backend/lib/core/workspace.py
```diff
@@ -218,16 +218,40 @@ def get_one(self, workspace_id: int) -> Workspace:
                 "workspace {} does not exist or not visible for user".format(workspace_id)
             ) from exc
 
-    def get_one_by_label(self, label: str) -> Workspace:
+    def get_one_by_filemanager_filename(
```
grignards, line 221: As filemanager filename is a webdav specific part I would have left the handler of the extension in webdav specific code and only call `get_one_by_label`.

grignards (changes requested):

backend/tracim_backend/lib/webdav/dav_provider.py
inkhey: We do need to : 
- add item at the end 
- get last item without removing it
nice to have : 
- be able to debug the code by just reading the list.

If we do use a simple list, it will work but as one of on first goal is to get the current last items, it's a bit more logical to have a list that allow direct access to the last item (without fetching all items). I do choose deque for this, but maybe there is better solution.

backend/tracim_backend/lib/core/workspace.py
grignards: s/where the users/where the user/.

backend/tracim_backend/lib/webdav/resources.py
grignards: As far as I remember mypy doesn't do type checking in functions/methods which do not have a completely typed signature. @raphj Do you know if there is a special case for `__init__` that would make mypy work without the `-> None` typing?

backend/tracim_backend/lib/webdav/dav_provider.py
inkhey: i suggest : 
```suggestion
    - get the workspaces hierarchy from path
    - get the contents hierarchy from path
    - provide useful properties to handle Webdav Request
```
There is a hierarchy of multiple workspaces and an hierarchy of multiple contents.

backend/tracim_backend/lib/webdav/dav_provider.py
grignards: I think that python's lists have a constant time random access operation (see https://wiki.python.org/moin/TimeComplexity) so they also would be OK I think.

backend/tracim_backend/lib/webdav/resources.py
inkhey: In fact, i don't know where this comment came from, and i'm not completely sure this method is not used by wsgidav (it's, so i keep implementation of both getMember and getMemberList for all resources. I suggest to make a comment about this unclear point.

grignards:
Same question about the author of the PR: should it be asf-algoo?

backend/tracim_backend/lib/webdav/dav_provider.py
inkhey: are you sure ? `self.environ["http_authenticator.realm"]` may be None ?

backend/tracim_backend/lib/webdav/dav_provider.py
inkhey: Interesting point. I do not have realize that python list are array.

backend/tracim_backend/lib/webdav/resources.py
raphj: Oh, I didn't know and you are right:

On this code:

```python
class C:
        def __init(self) -> None:
                a: int = 2
                print('a' + a)
```
Mypy says:

```
test.py:4: error: Unsupported operand types for + ("str" and "int")
Found 1 error in 1 file (checked 1 source file)
```

and on this one:

```python
class C:
        def __init(self):
                a: int = 2
                print('a' + a)
```

Mypy says: 

```
Success: no issues found in 1 source file
```

So yeah, maybe keep `-> None`. Typescript complains on this, hence my initial remark.

backend/tracim_backend/lib/webdav/resources.py
inkhey: No, it's related to wsgidav : https://github.com/mar10/wsgidav . Version 2.* that we do use use CamelCase, we do need to update the dependencies to support snake_case. I don't do it now, as this may be an heavy change.
see this ticket https://github.com/tracim/tracim/issues/2024

backend/tracim_backend/lib/webdav/dav_provider.py
inkhey: Doesn't work for error 404 case. self.current_workspace will be none in this case.

inkhey: another way to do this:
```python
        type = None
        if self.current_content:
            type = ResourceType.CONTENT
        elif self.current_workspace:
            type = ResourceType.WORKSPACE
        elif self.path == "/":
            type = ResourceType.ROOT
        return type
```

backend/tracim_backend/lib/webdav/dav_provider.py
inkhey: I agree this code is not so nice but i don't really like the idea of changing  dynamically the list used this way.
I suggest to put a comment about finding a way to refactor this better. I try some things but this part is complex so i don't find a proper solution.

backend/tracim_backend/lib/webdav/dav_provider.py
inkhey: done.

backend/tracim_backend/lib/webdav/dav_provider.py
inkhey: converted to list.

backend/tracim_backend/lib/core/workspace.py
inkhey: not really sure about this one. it's seems to me more clear to have this into core api.

backend/tracim_backend/lib/webdav/dav_provider.py
inkhey: done.

backend/tracim_backend/lib/webdav/dav_provider.py
inkhey: done.

backend/tracim_backend/lib/webdav/dav_provider.py
grignards: Updated my suggestion to avoid changing the parts list.

backend/tracim_backend/lib/webdav/resources.py
inkhey: Theses list are differents, i will not mix them because it make things unclear.

backend/tracim_backend/lib/webdav/resources.py
inkhey: i add a more clear info comment about this point.

backend/tracim_backend/lib/core/workspace.py
inkhey: done.

backend/tracim_backend/lib/webdav/dav_provider.py
inkhey: Updated with suggestion (with some fix to make it work), i do keep a comment about fact that this mecanism does work but it's not so clear for distinction of some case like invalid path, proper move path and root.

backend/tracim_backend/lib/webdav/resources.py
inkhey: done.

backend/tracim_backend/lib/core/workspace.py

grignards (approved):

raphj (approved):

Issue #3706, opened 2020-10-16
Feat: Allow user to leave space


gbonaspetti:
## Feature description and goals

Since the users have an option to join spaces (https://github.com/tracim/tracim/issues/3587) they should have the option to leave. 

- [x] This option will be in the My Account page in place of the red X of the mockup below:

![Capture d’écran de 2020-10-16 13-16-22](https://user-images.githubusercontent.com/35422692/96252230-ede88a00-0fb1-11eb-968e-95ea8243e9cf.png)

And it should be a (non primary) button with the icon `fa-sign-out` and the label "Leave space" with the normal font color.

- [x] When we clicking on this button, a confirmation popup should appear (like when we delete a space).
- [x] The option "Spaces and notifications" should be visible even if the notifications by email are deactivated, but the notification column should not be visible in this case.
- [x] It should be renamed "Spaces"

## Translations

|English|French|Portuguese|
|--|--|--|
|Leave space|Quitter l'espace|Deixar espaço|
|Are you sure you want to leave the space?|Voulez-vous vraiment quitter l'espace ?|Tem a certeza de que quer deixar o espaço?|

grignards:
Should we have another column header *Leave Workspace* like we have for space/role/notification?

raphj:
I'd like a column "Leave Space" yes, or even better I think, a button "Leave space" on each row. What do you think?

gbonaspetti:
>  The option "Spaces and notifications" should be visible even if the notifications by email are deactivated, but the notification column should not be visible in this case.

Should we rename "Spaces and notifications" to "Spaces"?

celiamgtt:
The icon should be the 'log out' icon ;) I like the idea to have a 'leave space button' (with the icon).
<img width="249" alt="Capture d’écran 2020-10-20 à 09 19 08" src="https://user-images.githubusercontent.com/50453784/96553385-5f2b8400-12b5-11eb-9803-daef95b319f2.png">

lebouquetin:
> Should we rename "Spaces and notifications" to "Spaces"?

Looks logical

> Should we have another column header Leave Space like we have for space/role/notification?

I think a column header is not required, but a button on each line is the right solution (like sugested by @raphj )

Note: the leave feature must generate a waring like @celiamgtt suggested during last UI meetup

raphj:
This branch requires commit f3f4971fa3e3f8e015626051fdd4eade5d6b79ba.
[...]

PR #3726, opened 2020-10-22
Feat/3706 Leave spaces

asf-algoo:
#3706

## Checkpoints

**For developers**

**For code reviewers**

**For testers**


frontend/src/component/Account/UserSpacesConfig.jsx
```diff
@@ -0,0 +1,135 @@
...
+    const fetchResult = await props.dispatch(deleteWorkspaceMember(spaceId, props.userToEditId))
+    if (fetchResult.status !== 204) {
+      props.dispatch(newFlashMessage(props.t('Error while leaving the space'), 'warning'))
+    }
+  }
+
+  render = () => {
+    const { props } = this
+
+    const entries = props.workspaceList.map(space => {
```
Skylsmoi, line 30: this const could be a separated component

frontend/src/component/Account/UserSpacesConfig.jsx
```diff
@@ -0,0 +1,135 @@
...
+
+    return (
+      <div className='account__userpreference__setting__spacename'>
+        <div className='spaceconfig__sectiontitle subTitle ml-2 ml-sm-0'>
+          {props.t('Spaces')}
+        </div>
+
+        <div className='spaceconfig__text ml-2 ml-sm-0' />
+
+        {(entries.length ? (
```
Skylsmoi, line 81: we usually indent the ? and : at a new line

frontend/src/component/Account/UserSpacesConfig.jsx
```diff
@@ -0,0 +1,135 @@
...
+        ))}
+      </div>
+    )
+  }
+}
+
+const mapStateToProps = ({ workspaceList }) => ({ workspaceList })
+export default connect(mapStateToProps)(translate()(UserSpacesConfig))
+
+UserSpacesConfig.propTypes = {
```
Skylsmoi, line 125: userToEditId and admin props aren't marked as `required` but don't have any default values

frontend_lib/src/component/ConfirmPopup/ConfirmPopup.jsx
```diff
@@ -0,0 +1,50 @@
...
+  confirmLabel: PropTypes.string,
+  onConfirm: PropTypes.func,
+  onCancel: PropTypes.func,
+  msg: PropTypes.string
+}
+
+ConfirmPopup.defaultProps = {
+  confirmLabel: '',
+  onConfirm: null,
+  onCancel: null,
```
Skylsmoi, line 46: the propTypes expects a function but you give `null` as default value

Skylsmoi (changes requested):

frontend/src/component/Account/UserSpacesConfig.jsx
raphj: Good idea, I'll do this

raphj: Well, actually this is not so much a component that a list of components that could be empty, so this is a bit tricky to do this.

frontend/src/component/Account/UserSpacesConfig.jsx
raphj: Should we add an eslint rule for this, to avoid having to check this during review?

frontend/src/component/Account/UserSpacesConfig.jsx
raphj: Indeed, I didn't know `isRequired`. Fixed.

frontend_lib/src/component/ConfirmPopup/ConfirmPopup.jsx
raphj: it expects an optional function:
>   You can declare that a prop is a specific JS type. By default, these are all optional.

https://reactjs.org/docs/typechecking-with-proptypes.html

The code handles null correctly.

frontend/i18next.scanner/en/translation.json
```diff
@@ -59,12 +59,12 @@
   ", create the first content by clicking on the button 'Create'": ", create the first content by clicking on the button 'Create'",
   "Calendar": "Calendar",
   "Access your personal calendar": "Access your personal calendar",
-  "Spaces and notifications": "Spaces and notifications",
   "Space": "Space",
   "Spaces": "Spaces",
   "Role": "Role",
   "Notification": "Notification",
   "You are not a member of any space yet": "You are not a member of any space yet",
+  "This user is not a member of any space yet": "This member is not a member of any space yet",
```
gbonaspetti, line 67: Why the key and the English translation are different?

frontend/i18next.scanner/fr/translation.json
```diff
@@ -275,6 +276,9 @@
   "{{author}} modified the space {{space}}": "{{author}} a mis à jour l'espace {{space}}",
   "{{author}} deleted the space {{space}}": "{{author}} a supprimé l'espace {{space}}",
   "{{author}} restored the space {{space}}": "{{author}} a restauré l'espace {{space}}",
-  "Only an administrator can see this user's profile": "Seule un administrateur peut consulter ce profil utilisateur",
+  "Only an administrator can see this user's profile": "Seule un administrateur peut consulter ce profil utilisateur",  "Leave space": "Quitter l'espace",
```
gbonaspetti, line 279:
```suggestion
  "Only an administrator can see this user's profile": "Seule un administrateur peut consulter ce profil utilisateur",
  "Leave space": "Quitter l'espace",
```

frontend/i18next.scanner/fr/translation.json
```diff
@@ -275,6 +276,9 @@
...
+  "Are you sure you want to leave the space?": "Voulez-vous vraiment quitter cet espace ?",
+  "Remove from space": "Retirer de l'espace",
+  "Error while leaving the space": "Error while leaving the space",
```
gbonaspetti, line 282: Still in english

frontend/src/component/Account/UserSpacesConfig.jsx
```diff
@@ -0,0 +1,228 @@
...
+  TLM_ENTITY_TYPE as TLM_ET,
+  TLM_CORE_EVENT_TYPE as TLM_CET
+} from 'tracim_frontend_lib'
+
+import { newFlashMessage } from '../../action-creator.sync.js'
+import { deleteWorkspaceMember, getUserWorkspaceList, getWorkspaceMemberList } from '../../action-creator.async.js'
+
+export class UserSpacesConfig extends React.Component {
+  constructor (props) {
+    super()
```
gbonaspetti, line 21:
```suggestion
    super(props)
```

frontend/src/component/Account/UserSpacesConfig.jsx
```diff
@@ -0,0 +1,228 @@
...
+      { entityType: TLM_ET.SHAREDSPACE_MEMBER, coreEntityType: TLM_CET.MODIFIED, handler: this.handleMemberModified },
+      { entityType: TLM_ET.SHAREDSPACE_MEMBER, coreEntityType: TLM_CET.DELETED, handler: this.handleMemberDeleted }
+    ])
+  }
+
+  handleMemberModified = data => {
+    const { props } = this
+    if (Number(props.userToEditId) !== data.fields.user.user_id) return
+    this.setState(prev => ({
+      workspaceList: prev.workspaceList.map(space =>
```
gbonaspetti, line 39: Since you use space here, I think you can rename `workspaceList` into `spaceList`

frontend/src/component/Account/UserSpacesConfig.jsx
```diff
@@ -0,0 +1,228 @@
...
+          : space
+      )
+    }))
+  }
+
+  handleMemberDeleted = data => {
+    const { props } = this
+    if (Number(props.userToEditId) !== data.fields.user.user_id) return
+    this.setState(prev => ({
+      workspaceList: prev.workspaceList.filter(ws =>
```
gbonaspetti, line 57: Avoid abbreviations (`ws`)

frontend/src/component/Account/UserSpacesConfig.jsx
```diff
@@ -0,0 +1,228 @@
...
+  }
+
+  handleMemberCreated = async data => {
+    const { props } = this
+    if (Number(props.userToEditId) !== data.fields.user.user_id) return
+
+    const newSpace = await this.fillMemberList(data.fields.workspace)
+
+    this.setState(prev => ({
+      workspaceList: [...prev.workspaceList, newSpace].sort(
```
gbonaspetti, line 70: Can't you use the function `frontend_lib/src/helper/sortWorkspaceList` here?

frontend/src/component/Account/UserSpacesConfig.jsx
```diff
@@ -0,0 +1,228 @@
...
+      default: props.dispatch(newFlashMessage(props.t('Error while loading user')))
+    }
+  }
+
+  fillMemberList = async space => {
+    const fetchMemberList = await this.props.dispatch(getWorkspaceMemberList(space.workspace_id))
+
+    return {
+      ...space,
+      memberList: fetchMemberList.json || [] // handle error?
```
gbonaspetti, line 92: Forgotten comment?

frontend/src/component/Account/UserSpacesConfig.jsx
```diff
@@ -0,0 +1,228 @@
...
+    }
+  }
+
+  getUserWorkspaceListMemberList = async (wsList) => {
```
gbonaspetti, line 96: Avoid abbreviations (`wsList`)

frontend/src/component/Account/UserSpacesConfig.jsx
```diff
@@ -0,0 +1,228 @@
...
+    const workspaceList = await Promise.all(wsList.map(this.fillMemberList))
+    this.setState({ workspaceList })
+  }
+
+  handleConfirmDeleteSpace = async () => {
+    const { props } = this
+    const spaceId = this.state.spaceBeingDeleted
+    if (!spaceId) return
+    this.setState({ spaceBeingDeleted: null })
```
gbonaspetti, line 105: I think this should only happen if fetchResult is 204. 
If I understand well, if the person clicks once, has some problem and clicks again, he/she will lose the id in the first click, right?

frontend/src/component/Account/UserSpacesConfig.jsx
```diff
@@ -0,0 +1,228 @@
...
+
+    const fetchResult = await props.dispatch(deleteWorkspaceMember(spaceId, props.userToEditId))
+    if (fetchResult.status !== 204) {
+      props.dispatch(newFlashMessage(props.t('Error while leaving the space'), 'warning'))
+    }
+  }
+
+  componentDidMount = this.getWorkspaceList
```
gbonaspetti, line 113: Does this work?

frontend/src/component/Account/UserSpacesConfig.jsx
```diff
@@ -0,0 +1,228 @@
...
+
+  componentDidUpdate (prevProps) {
+    if (prevProps.userToEditId !== this.props.userToEditId) {
+      this.getWorkspaceList()
+    }
+  }
+
+  render = () => {
```
gbonaspetti, line 121:
```suggestion
  render () {
```

frontend/src/container/Account.jsx
```diff
@@ -48,7 +46,7 @@ import AgendaInfo from '../component/Dashboard/AgendaInfo.jsx'
 
 export class Account extends React.Component {
   constructor (props) {
-    super(props)
+    super()
```
gbonaspetti, line 49: Why are you removing props here?

frontend_lib/src/component/ConfirmPopup/ConfirmPopup.jsx
```diff
@@ -0,0 +1,52 @@
...
+          onClick={props.onConfirm}
+        >
+          {props.confirmLabel || props.t('Confirm')}
+        </button>
+      </div>
+    </div>
+  </CardPopup>
+)
+
+ConfirmPopup.propType = {
```
gbonaspetti, line 38:
```suggestion
ConfirmPopup.propTypes = {
```

gbonaspetti (changes requested):

frontend_lib/src/component/ConfirmPopup/ConfirmPopup.jsx
Skylsmoi: if you ask for a function that means you will bind some handlers to these parameters
If it is null, it might be converted to `null()` which will trigger a js error.
Also, optional parameters will be `undefined` if not given without a default value

frontend/src/component/Account/UserSpacesConfig.jsx
Skylsmoi: standardjs already state that rule
If it doesn't show a warning in this case is more likely to be a standardjs issue
So no, we won't add an eslint rule for that case

frontend/i18next.scanner/en/translation.json
Skylsmoi: odd indeed, we are not supposed to manually edit the english translation file since they must be the translation key

frontend/src/component/Account/UserSpacesConfig.jsx
raphj: I don't think so.

frontend_lib/src/component/ConfirmPopup/ConfirmPopup.jsx
raphj: But since it is optional, I check before use.

frontend/src/component/Account/UserSpacesConfig.jsx
raphj: This was discouraged by VSCodium, but it seems to be an issue with VSCodium: https://stackoverflow.com/questions/63883401/react-native-context-and-deprecated-super-usage

Some interesting read on the subject: https://overreacted.io/why-do-we-write-super-props/

frontend/src/component/Account/UserSpacesConfig.jsx
Skylsmoi: My comment isn't on the right line
I was talking about the `<tr>` code that should be in a separated component

Also, on a second look, instead of returning undefined on some cases and then .filter(entries => entries) afterwards, you could do the filter before the map.

And finally, React doesn't like components that are undefined. That's why you filter them. But it doesn't mind components as `null`, it just renders nothing

frontend/i18next.scanner/fr/translation.json
```diff
@@ -276,7 +276,8 @@
   "{{author}} modified the space {{space}}": "{{author}} a mis à jour l'espace {{space}}",
   "{{author}} deleted the space {{space}}": "{{author}} a supprimé l'espace {{space}}",
   "{{author}} restored the space {{space}}": "{{author}} a restauré l'espace {{space}}",
-  "Only an administrator can see this user's profile": "Seule un administrateur peut consulter ce profil utilisateur",  "Leave space": "Quitter l'espace",
+  "Only an administrator can see this user's profile": "Seule un administrateur peut consulter ce profil utilisateur",
```
Skylsmoi, line 279: `Seul un administrateur`
minus the -e

Skylsmoi (changes requested):

frontend/src/component/Account/UserSpacesConfig.jsx
raphj: yes :-)

frontend/src/container/Account.jsx
raphj: See my other comment ;-)
```suggestion
    super(props)
```

frontend/i18next.scanner/fr/translation.json
raphj:
```suggestion
  "Only an administrator can see this user's profile": "Seul un administrateur peut consulter ce profil utilisateur",
```

frontend/src/component/Account/UserSpacesConfig.jsx
raphj: I need to know whether there are no spaces to show so we can show the "no spaces here" message. I'll try filtering before but I'm not sure this is easy (I would prefer to filter before)

frontend/src/component/Account/UserSpacesConfig.jsx
Skylsmoi: https://standardjs.com/rules.html
look for "ternary operator"

frontend_lib/src/component/ConfirmPopup/ConfirmPopup.jsx
Skylsmoi: mmh I don't see where

frontend/src/component/Account/UserSpacesConfig.jsx
raphj: Ah right, this is strange that standard did not complain then

frontend/src/component/Account/UserSpacesConfig.jsx
Skylsmoi: It's also a bit confusing to me, it would be easier to call `this.getWorkspaceList` in the `componentDidMount` body

frontend/src/component/Account/UserSpacesConfig.jsx
raphj: workspaceList is used everywhere in the frontend, I'd suggest to keep it like this for now so the same thing has the same name everywhere.

I opened an issue: https://github.com/tracim/tracim/issues/3734

frontend/src/component/Account/UserSpacesConfig.jsx
raphj: If the user clicks again, this will open the confirmation popup again and set back `spaceBeingDeleted` to the correct value. `spaceBeingDeleted` is intended to have a value only when the confirmation popup is open. Is it ok for you?

frontend_lib/src/component/ConfirmPopup/ConfirmPopup.jsx
raphj: onClick={`props.onCancel`} will be `onClick={null}` if `props.onCancel` is `null`, which is fine.

Anyway, Il make these props required, it does not make sense to not set them.

frontend/src/component/Account/UserSpacesConfig.jsx
gbonaspetti: Does the error close the popup? If so, it's ok for me and you can solve this conversation.

frontend/src/component/Account/UserSpacesConfig.jsx
raphj: I used reduce to avoid the filter and extracted the line in a separate component as requested.

frontend_lib/src/component/ConfirmPopup/ConfirmPopup.jsx
raphj: done.

frontend/src/component/Account/UserSpacesConfig.jsx
raphj: ok, done

frontend/src/component/Account/UserSpacesConfig.jsx
raphj: The user clicks on the confirm button, and then the request is sent. The popup is not shown at the moment an error could happen.

frontend/src/component/Account/UserSpacesConfig.jsx
raphj: Ah, this is a moved code. But you are right, i'll rename this variable.

frontend/src/component/Account/UserSpacesConfig.jsx
raphj: Good idea, but I want to make sure that the order is always the same so I'll update sortWorkspaceList so it also sorts by id when the name are the same.

frontend/src/component/Account/UserSpacesConfig.jsx
raphj: nope, this is a comment from a moved code. I decided to leave it here

frontend/src/component/Account/UserSpacesConfig.jsx
raphj: done

frontend/src/component/Account/UserSpacesConfig.jsx
```diff
@@ -0,0 +1,208 @@
...
+import { deleteWorkspaceMember, getUserWorkspaceList, getWorkspaceMemberList } from '../../action-creator.async.js'
+
+import UserSpacesConfigLine from './UserSpacesConfigLine.jsx'
+
+export class UserSpacesConfig extends React.Component {
+  constructor (props) {
+    super(props)
+
+    this.state = {
+      workspaceList: [],
```
grignards, line 24: Does this workspace list have a different behavior than the one in redux? If not you should use the redux one as it would save much code here (no need to handle tlms to update this list/the list of members/…).

gbonaspetti (approved):

frontend/src/component/Account/UserSpacesConfigLine.jsx
```diff
@@ -0,0 +1,53 @@
...
+            onChange={() => props.onChangeSubscriptionNotif(space.workspace_id, !member.do_notify)}
+          />
+        </td>
+        <td data-cy='spaceconfig__table__leave_space_cell'>
+          <IconButton
+            className='outlineTextBtn primaryColorBorder primaryColorBgHover primaryColorBorderDarkenHover'
+            onClick={(() => props.onLeaveSpace(space.workspace_id))}
+            icon='sign-out'
+            text={props.admin ? props.t('Remove from space') : props.t('Leave space')}
+          />
```
grignards, line 46: As you'll merge after my PR, be aware that `IconButton` changed a little.
To have the same button as this one you will have to declare it as:
```javascript
<IconButton
  mode="light"
  onClick={(() => props.onLeaveSpace(space.workspace_id))}
  icon='sign-out'
  text={props.admin ? props.t('Remove from space') : props.t('Leave space')}
/>
```

grignards (changes requested):

frontend/src/component/Account/UserSpacesConfigLine.jsx
raphj: thx

frontend/src/component/Account/UserSpacesConfig.jsx
Skylsmoi: this is moved code that I originally made a while ago
I think there is reason to not use the redux space list, I don't remember it right now, I'm looking into it quickly

raphj: https://github.com/tracim/tracim/issues/3735

grignards (approved):

gbonaspetti (approved):

gbonaspetti:
Some bugs were found during tests:
- Sidebar active space bug
Capture d’écran de 2020-10-28 09-10-23
- The leave button is in three lines
Capture d’écran de 2020-10-28 08-53-40
- If I'm the only responsible and I can't leave the space, so I open a new tab and add a user as responsible, the disabled of the button is not real-time updated, I have to refresh the page
- The notification columnis visible even if the notification by email are deactivated

raphj:
Fixed the bug, IconButtons are know whitespace:pre

frontend/src/component/Account/UserSpacesConfigLine.styl
```diff
@@ -0,0 +1,8 @@
+@import "~tracim_frontend_lib/src/css/Variable.styl"
+
+@media (max-width: max-sm)
+  .spaceconfig__table__leave_space_cell
+    & .iconbutton__label
```
grignards, line 5:
```suggestion
    & .iconbutton__text
```
As the property is named `text` in `IconButton`

frontend_lib/src/css/IconButton.styl
```diff
@@ -33,3 +34,5 @@
   &__secondary_dark
     border-color: var(--primaryDarkColor)
     border-width: 1px
+  &__label
```
grignards, line 37:
```suggestion
  &__text
```
as the property is named `text`.

frontend_lib/src/component/Button/IconButton.jsx
```diff
@@ -23,6 +23,7 @@ const IconButton = props => {
       style={style}
       onClick={props.onClick}
       disabled={props.disabled}
+      title={props.title}
```
grignards, line 26:
```suggestion
      title={props.title || props.text}
```
Like this we always have a tooltip.

gbonaspetti (approved):

frontend_lib/src/component/Button/IconButton.jsx
```diff
@@ -23,6 +23,7 @@ const IconButton = props => {
       style={style}
       onClick={props.onClick}
       disabled={props.disabled}
+      title={props.title || props.text}
       data-cy={props.dataCy}
     >
       <i className={`fa fa-fw fa-${props.icon}`} /> {props.text}
```
gbonaspetti, line 29:
```suggestion
      <i className={`fa fa-fw fa-${props.icon}`} /> <span className='iconbutton__text'>{props.text}</span>
```

gbonaspetti:
Tested on branch

PR #3732, opened 2020-10-23
Feat/3587  join spaces

asf-algoo:
See #3587

## Checkpoints

**For developers**

**For code reviewers**

**For testers**


frontend/src/component/Home/HomeNoWorkspace.jsx
```diff
@@ -1,10 +1,19 @@
...
 
+import { IconButton } from 'tracim_frontend_lib'
+
 require('./Home.styl')
 
+const getWorkspaceMessage = (canCreate, canJoin) => {
+  if (canCreate && canJoin) return 'You can join an existing space or create your first space'
+  if (canCreate && !canJoin) return 'You can create your first space'
+  if (!canCreate && canJoin) return 'You can join an existing space'
+  return 'Please ask access to spaces to an administrator or space manager'
```
Skylsmoi, line 12: if the string isn't wrapped in a t() function, it wont be parsed by the build-translation script.
To be clearer, your translation process works. It is only the translation keys that won't be added in the translation files by the script build-translation

Skylsmoi: the solution is to call props.t on each strings of this functino instead of on the function return value

frontend/src/container/JoinWorkspace.jsx
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  createRequestComponent (workspace) {
+    const { props } = this
+
+    const subscription = props.workspaceSubscriptionList.find(s => s.workspace.workspace_id === workspace.id)
+
+    if (subscription !== undefined) {
+      var text = 'Unknown request state'
+      var icon = 'question'
```
Skylsmoi, line 91: don't use var keyword, prefer const. Or, `let` here since you reassign it
also, `text` isn't translated

frontend/src/container/JoinWorkspace.jsx
```diff
@@ -0,0 +1,181 @@
...
+          />)
+      case SPACE_TYPE.open.slug:
+        return (
+          <IconButton
+            icon='sign-in'
+            text={props.t('Join the space')}
+            onClick={() => this.joinWorkspace(workspace.id)}
+          />)
+      default:
+        return 'Unknown space access type'
```
Skylsmoi, line 121: if your function usually returns a component, I think it would be better if your default state would also return a "component like" thing.
Wrapping the string in `<span>` or `<div>` should be enough

frontend/src/container/JoinWorkspace.jsx
```diff
@@ -0,0 +1,181 @@
...
+  createIconForAccessType (accessType) {
+    const spaceType = SPACE_TYPE_LIST.find(t => t.slug === accessType)
+    return spaceType
+      ? <i class={`fa fa-fw fa-2x fa-${spaceType.faIcon}`} title={spaceType.label} />
+      : <i class='fa fa-fw fa-2x fa-search' title='Unknown space type' />
+  }
+
+  handleWorkspaceFilter (filter) {
+    const re = new RegExp(escapeRegExp(filter), 'i')
+    const filterFunc = workspace => workspace.label.search(re) >= 0 || workspace.description.search(re) >= 0
```
Skylsmoi, line 134: I'm not sure what this exactly does but, if it is only to search if a workspace label or description contains the string `filter` you can use the function `.includes()` on a string.
It would avoid having to use regex

frontend/src/container/ReduxTlmDispatcher.jsx
```diff
@@ -85,15 +95,27 @@ export class ReduxTlmDispatcher extends React.Component {
     }
   }
 
+  handleWorkspaceCreated = data => {
+    const { props } = this
+    if (ACCESSIBLE_SPACE_TYPE_LIST.find(s => s.slug === data.fields.workspace.access_type) !== undefined) {
```
Skylsmoi, line 100: alternatively, you can use `Array.some()` which returns a boolean instead of using `Array.find() !== undefined`

frontend/test/src/container/JoinWorkspace.spec.js
```diff
@@ -0,0 +1,73 @@
...
+      expectedIcon: SUBSCRIPTION_TYPE.rejected.faIcon,
+      expectedText: 'Request rejected'
+    }
+  ]
+
+  createRequestComponentTestCasesWithSubscription.forEach(testCase => {
+    const workspace = { id: 1, accessType: SPACE_TYPE.onRequest.slug }
+    const workspaceSubscriptionList = [{ workspace: { workspace_id: 1 }, state: testCase.subscriptionState }]
+    describe(`its createRequestComponent() function with a subscription state '${testCase.subscriptionState}'`, () => {
+      const wrapper = shallow(<JoinWorkspaceWithoutHOC {...{ ...props, workspaceSubscriptionList: workspaceSubscriptionList }} />)
```
Skylsmoi, line 64: do you need the to objects assignment here ?
the `{... { ...props, ... }}`

frontend_lib/src/component/Input/TextInput.jsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react'
+import PropTypes from 'prop-types'
+
+require('./TextInput.styl')
+
+// A thin-layer around <input type='text'> that sets the Tracim style and behaviour
```
Skylsmoi, line 6: comment must use the required syntax:
INFO - developer - date - text

frontend_lib/src/component/Input/TextInput.jsx
```diff
@@ -0,0 +1,50 @@
...
+// Also provide a onValidate handler which is called when Enter is pressed by the user.
+export class TextInput extends React.Component {
+  validateIfEnterKey(e) {
+    if (e.key !== 'Enter' || !props.onValidate) return
+    props.onValidate(e)
+  }
+
+  render() {
+    const { props } = this
+    return <div class='textinput__box'>
```
Skylsmoi, line 16: to align vertically the opening and closing tags, we usually use parenthesis:
``` jsx
return (
  <div className=''>
    ...
  </div>
)

Comment (Skylsmoi) [reply frontend_lib/src/component/Input/TextInput.jsx:16]: also, `class` keyword is forbidden in jsx. you must use `className`.
The reason is that jsx is translated into js and the `class` keyword is a reserved js keyword

File: frontend_lib/src/component/Input/TextInput.jsx [hunk 9]
```diff
@@ -0,0 +1,50 @@
...
+      <input
+        type='text'
+        class='textinput__text form-control primaryColorBorder'
```
Skylsmoi, line 19: className

Skylsmoi (changes requested):

frontend/src/container/JoinWorkspace.jsx
grignards: I didn't translate the text as this is more a "bug detection" to avoid breaking the interface rather than a string that should usually be displayed to users.

frontend/src/component/Home/HomeNoWorkspace.jsx
grignards: I wasn't aware that we'd use such a script to detect translation keys, will change my code then.

frontend/src/container/JoinWorkspace.jsx
raphj: or possibly the empty element `<>..</>`?

frontend/src/container/JoinWorkspace.jsx
raphj: `workspace.label.toLowerCase().includes(filter.toLowerCase())` for case insensitivity.

frontend/test/src/container/JoinWorkspace.spec.js
grignards: Nope, this is overly complex, I'll change it.

frontend/i18next.scanner/en/translation.json
```diff
@@ -277,5 +275,16 @@
...
+  "Access request": "Access request",
+  "Request access": "Request access",
+  "Request sent": "Request sent",
+  "Join the space": "Join the space",
+  "Filter spaces": "Filter spaces",
+  "You can join an existing space or create your first space": "You can join an existing space or create your first space",
+  "You can create your first space": "You can create your first space",
+  "You can join an existing space": "You can join an existing space",
+  "Please ask access to your spaces to an administrator or space manager": "Please ask access to your spaces to an administrator or space manager"
+}
```
raphj, line 290: missing new line?

frontend/i18next.scanner/fr/translation.json
```diff
@@ -276,5 +275,16 @@
...
+  "Access request": "Demande d'accès",
+  "Request access": "Demander l'accès",
+  "Request sent": "Demande envoyée",
+  "Join the space": "Rejoindre l'espace",
+  "Filter spaces": "Filtrer les espaces",
+  "You can join an existing space or create your first space": "Vous pouvez rejoindre un espace existant ou créer votre premier espace",
+  "You can create your first space": "Vous pouvez créer votre premier espace",
+  "You can join an existing space": "Vous pouvez rejoindre un espace existant",
+  "Please ask access to your spaces to an administrator or space manager": "Veuillez demander accès à vos espaces à un administrateur ou un gestionnaire d'espace"
+}
```
raphj, line 290: missing newline here too

frontend/i18next.scanner/pt/translation.json
```diff
@@ -277,5 +276,16 @@
...
+  "Access request": "Pedido de acesso",
+  "Request access": "Pedir acesso",
+  "Request sent": "Pedido enviado",
+  "Join the space": "Junte-se ao espaço",
+  "Filter spaces": "Filtrar espaços",
+  "You can join an existing space or create your first space": "Você pode juntar-se a um espaço existente ou criar o seu primeiro espaço",
+  "You can create your first space": "Você pode criar o seu primeiro espaço",
+  "You can join an existing space": "Você pode juntar-se a um espaço já existente",
+  "Please ask access to your spaces to an administrator or space manager": "Por favor peça acesso aos seus espaços a um administrador ou gestor de espaço"
+}
```
raphj, line 291: and here :-)

frontend/src/component/Home/HomeNoWorkspace.jsx
```diff
@@ -16,32 +25,36 @@ export const HomeNoWorkspace = props =>
...
-              {props.t("You aren't member of any space yet")}
-            </div>
-            <div>{props.t('Please refer to an administrator or a trusted user')}</div>
-          </div>
-        )
-      )}
+      <div className='homepagecard__text__user__top'>
+        {props.t("You aren't member of any space yet")}
+      </div>
+      {getWorkspaceMessage(props.canCreateWorkspace, props.canJoinWorkspace, props.t)}
```
raphj, line 31: I'm not completely sure that you can pass props.t in parameter because of the way translation strings are parsed when building the translation list. Maybe you'll need to pass props, or use `const {props} = this` in `getWorkspaceMessage` (and yes, unfortunately make it depend on `this`.

(@Skylsmoi ?)

frontend/src/css/index.styl
```diff
@@ -24,3 +24,4 @@
 @import 'Account.styl'
 
 @import 'SearchResult.styl'
+@import 'JoinWorkspace.styl'
```
raphj, line 27: is this needed?
(missing new line)

frontend_lib/src/component/Input/TextInput.jsx
raphj: It seems to me that a comment documenting the intent of a class should not need this format.
This is something we do not do in the backend for instance.

This format seems useful for wargings, fixmes, todos, or notes on something tricky in the code.

raphj (approved):

frontend/src/action-creator.async.js
```diff
@@ -914,3 +916,69 @@ export const getUserMessagesSummary = userId => dispatch => {
...
+        ...FETCH_CONFIG.headers
+      },
+      method: 'GET'
+    },
+    actionName: WORKSPACE_SUBSCRIPTION_LIST,
+    dispatch
+  })
+}
+
+export const joinWorkspace = (workspaceId, userId) => dispatch => {
```
gbonaspetti, line 950: I have a slight preference for making it clear in the name of the function what method it is implementing, for example `postSpaceMember` or `postUserSpace`, but if you prefer to leave it like this, you can just solve this conversation. If you decide to change the name, I suggest you also change it to the `subscribeToWorkspace` function (maybe `postSpaceRequest`).

raphj: This is indeed how the other functions are named in these files, the HTTP method appears in the name.

functionnal_tests/cypress/fixtures/onRequestWorkspace.json
```diff
@@ -0,0 +1,8 @@
+{
+    "description": "A super description of my space.",
+    "label": "My ON REQUEST space",
+    "access_type": "on_request",
+    "default_user_role": "reader",
+    "workspace_id": 3,
+    "parent_id": null
+}
```
raphj, line 8: new line

functionnal_tests/cypress/fixtures/openWorkspace.json
```diff
@@ -0,0 +1,8 @@
+{
+    "description": "A super description of my space.",
+    "label": "My OPEN space",
+    "access_type": "open",
+    "default_user_role": "reader",
+    "workspace_id": 2,
+    "parent_id": null
+}
```
raphj, line 8: new line

raphj (approved):

frontend/src/container/JoinWorkspace.jsx
Skylsmoi: this is called React Fragment and its purpose is to avoid adding an extra node to the DOM.
So I don't think it will fit the needs here

frontend/src/component/Home/HomeNoWorkspace.jsx
Skylsmoi: It will work because build-translation script looks for `props.t()` and `t()` functions. I don't think that this is a great pattern though.

Alternatively, you can keep your old way of calling props.t on the function return and add a fake function `const f = p => p` for `getWorkspaceMessage` to trick the build-translation script into adding the key. But that might be an even worst pattern

If I remember correctly, there is a way to `require` or `import` the translation function from i18n. The way we use in Tracim is using the HOC to get it but that isn't the only way. It might be cleaner that passing it through a parameter. But then, your function `getWorkspaceMessage` wont be pure anymore. I don't know what is the best way. The one that is currently commited is fine to me

Skylsmoi (approved):

frontend_lib/src/component/Button/IconButton.jsx
```diff
@@ -1,31 +1,55 @@
...
 }
 
 IconButton.defaultProps = {
-  className: '',
-  onClick: () => {},
+  onClick: undefined,
   disabled: false,
-  text: '',
-  style: {}
+  color: GLOBAL_primaryColor, // eslint-disable-line no-undef
```
raphj, line 48: Wouldn't it be better to GLOBAL_primaryColor in field `globals` of `package.json`?

Skylsmoi: it is set by the backend in index.max
that's why it needs to be a gobal
I agree that it's not ideal but that's the easier solution

frontend/src/container/JoinWorkspace.jsx
raphj: I know what it is. Why?

frontend_lib/src/component/Button/IconButton.jsx
raphj: I suggested to add `GLOBAL_primaryColor` to the `globals` standard's field of package.json to avoid disabling the eslint rule on this line

gbonaspetti:
I found some problems during the tests: 

- We can't request again to join a space if the request was rejected
- Shouldn't we have the tree view too? (as in creation space popup)
- The icon in the filter bar is not aligned
- A big description or space name is not chopped with ellipsis (...) and it has not a tooltip
- Problem with button in Portuguese translation
join
- Scroll problem
scroll

raphj:
> Shouldn't we have the tree view too? (as in creation space popup)

I don't think so, I'd say let's see if it would be needed

raphj (approved):

gbonaspetti:
- Still having a problem with scroll, I can't see all the spaces
scroll

- The buttons at the sidebar and home page, if you are an admin and don't have any space, do not have a margin between them.

- A big name is still not "overflowed" by ellipsis.

- There are some "refresinhg" problems, sometimes I can see a space that I'm already a member, sometimes when I refresh I have a different list that before (it feels like the list is not updated when I change the user)
Screenshot from 2020-10-27 09-13-12
user

- If a request is rejected, when the user tries to request again, nothing happens even if the request was send (if the user refresh the page we have a "request sent")
nothing

- White screen when connecting first time with a user that don't have any spaces
Screenshot from 2020-10-27 09-21-46
Screenshot from 2020-10-27 09-21-42
Screenshot from 2020-10-27 09-21-29

frontend/src/container/JoinWorkspace.jsx
```diff
@@ -0,0 +1,183 @@
...
+          />)
+      case SPACE_TYPE.open.slug:
+        return (
+          <IconButton
+            icon='sign-in'
+            text={props.t('Join the space')}
+            onClick={() => this.joinWorkspace(workspace.id)}
+          />)
+      default:
+        return <span>Unknown space access type</span>
```
raphj, line 118:
```suggestion
        return <span>Unknown space access type</span>
```
I know this is a "should never happen" kind of message, but should this be translated anyway?

grignards:
@gbonaspetti I cannot reproduce "There are some "refresinhg" problems, sometimes I can see a space that I'm already a member, sometimes when I refresh I have a different list that before (it feels like the list is not updated when I change the user)", do you have a way of reproducing this behavior?

gbonaspetti:
> @gbonaspetti I cannot reproduce "There are some "refresinhg" problems, sometimes I can see a space that I'm already a member, sometimes when I refresh I have a different list that before (it feels like the list is not updated when I change the user)", do you have a way of reproducing this behavior?

After re-test, your changes have fixed the problem. Good job ^^

gbonaspetti:
Tested on branch.

Issue #3771, opened 2020-10-30
Bug: The border of an active space moves the name of the space.


gbonaspetti:
## Description and expectations

The border of an active space moves the name of the space. Since the character shifts, it can be confused with a sub-space of lower rank (in the image below Sous⁵ and Sous⁵ child are on the same level, both children of Sous⁴ child).

![Screenshot from 2020-10-30 16-17-36](https://user-images.githubusercontent.com/35422692/97723967-a511f480-1acc-11eb-91f6-36eebb128297.png)

Related issue to fix: `<span>` elements are children of <ul> and parents of `<li>` elements in the sidebar.

To do:
 - [x] fix the margin
 - [x] remove the `<span>`

grignards:
Tested on branch.

gbonaspetti:
After testing, this issue was not completely fixed.

raphj:
tested on branch.

PhilippeAccorsi:
Done.

PR #3803, opened 2020-11-04
Fix/3771  border active space should not move the title at sidebar

asf-algoo:
#3771 

## Checkpoints

**For developers**

**For code reviewers**

**For testers**

### DIFF
--- a/frontend/src/component/Sidebar/WorkspaceListItem.jsx
+++ b/frontend/src/component/Sidebar/WorkspaceListItem.jsx
@@ -69,6 +69,10 @@ class WorkspaceListItem extends React.Component {
 
   render () {
     const { props, state } = this
+    const INDENT_WIDTH = 20
+    const BASE_MARGIN = 10
+    const ACTIVE_BORDER_WIDTH = 5
+
     return (
       <li
         id={props.workspaceId}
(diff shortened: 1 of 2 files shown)


frontend/src/component/Sidebar/WorkspaceListItem.jsx
```diff
@@ -94,8 +94,18 @@ class WorkspaceListItem extends React.Component {
...
+            // If the space is active the border width (5px) is subtracted
             props.level > 0 && (
-              <div style={{ marginLeft: `${(props.level - 1) * 20 + 10}px` }}>&#8735;</div>
+              <div
+                style={{
+                  marginLeft:
+                  `${(props.level - 1) * 20 + 10 -
+                    (props.location.pathname.includes(`${PAGE.WORKSPACE.ROOT}/${props.workspaceId}/`) ? 5 : 0)
+                  }px`
+                }}
```
raphj, line 105: Would it be possible to move this style in a class?

raphj (changes requested):

frontend/src/component/Sidebar/WorkspaceListItem.jsx
gbonaspetti: is there a way to pass the variables to the class? doesn't it make it more complicated knowing that we even have a conditional in it?

raphj: You are right. I'll try to think of an improvement for later.

Maybe put 10 and 20 in named constant though?

gbonaspetti: I don't think is needed here, normally I prefer constants, but in this case they are just number without a meaning, like:
const TEN = 10
const TWENTY = 20
`${(props.level - 1) * TWENTY + TEN - ...

raphj (approved):

Issue #3809, opened 2020-11-05
Bug: Breadcrumbs need to be improved


raphj:
## Description and expectations

Everywhere in Tracim, breadcrumbs currently do not show the full path. They show the space, the app and the content name.
Other problem is that they don't handle big words:

![98917562-d4f9c880-24cc-11eb-9e67-4f243119866e](https://user-images.githubusercontent.com/35422692/98924054-b26bad80-24d4-11eb-831d-138851efb740.png)

Expected (frontend):
Tracim has two types of breadcrumbs: page related and content related.

![breadcrumbs_page](https://user-images.githubusercontent.com/35422692/101490903-4eb59280-3963-11eb-9e54-45bfc02c7b80.png)

![breadcrumbs_content](https://user-images.githubusercontent.com/35422692/101491077-989e7880-3963-11eb-95ad-f7aea3f3138f.png)

 - [x] Page related breadcrumbs should always have Home page in it. In small and medium screens (xs, sm and md) and if we have more than 5 layers, we show only the icon `fa-home`.
 - [x] The current page at page related breadcrumbs should not be clickable.
 - [x] All clickable elements should have a tootltip.
 - [x] If the breadcrumbs have more than 5 layers (with the home page, 4 without) we put non-clickable ellipsis between the space and last element name, e.g. `home > space > ... > folder > sub-folder > content` and it should have a tooltip with the whole path in this ellipsis.
 - [x] Breadcrumbs should always be in one line only. It should handle big words with ellipsis.
 - [x] The breadcrumbs should include folders, if pertinent.

Expected (backend):
 - [x] New endpoint `GET /api/workspaces/{workspace_id}/contents/{content_id}/path` that returns a ordered array with the path to the content. Something like this _(this is just an example, it can be changed/adapted during development. please update this description if the implementation is different)_:
```
{
  "items": [
    {
      "content_id": 42,
      "content_type": "folder",
      "label": "A folder"
    },
   {
      "content_id": 44,
      "content_type": "folder",
      "label": "A sub-folder"
    },
   {
      "content_id": 46,
      "content_type": "file",
      "label": "A file"
    }
  ]
}
```
## Work branch

fix/3809__improve_breadcrumbs

## Extra information

- In the `frontend/src/container/SearchResult.jsx` component of frontend we have the `getPath` function that gets the pathname of a content.

### Should we still show the content app in the breadcrumb?

Decision: No, the breadcrumbs should not have elements there are not navigable.

[...]

PR #3943, opened 2020-12-09
Fix/3809  improve breadcrumbs

gbonaspetti:
#3809 

## Checkpoints

**For developers**

- [x] If there are modifications in the Tracim configuration files (eg. `development.ini`), make sure it is documented in `backend/doc/setting.md` and notify relevant people of this change

**For code reviewers**

**For testers**

**Testing method**

Create contents with several parent folders, check that the `…` is displayed as specified in the issue. Check that the last breadcrumb is not a link in page title (but others are and that links are functional).
Navigate in all application pages and check that the breadcrumbs are correct.


backend/tracim_backend/models/context_models.py
```diff
@@ -1637,6 +1637,11 @@ def __init__(self, page: Page) -> None:
         self.items = page
 
 
+class ListItemsObject(object):
```
inkhey, line 1640: Not sure how to name this, i do need to encapsulate in "items" my result.

backend/tracim_backend/views/core_api/schemas.py
```diff
@@ -1326,6 +1326,13 @@ def make_content_creation(self, data: typing.Dict[str, typing.Any]) -> object:
         return ContentCreation(**data)
 
 
+class ContentMinimalSchema(marshmallow.Schema):
```
inkhey, line 1329: ContentDigestSchema contains many data, i'm not sure we do want to return so much data just to get breadcrumb.

On the other side, this mean one more schema for content (we have 3 with this on)

backend/tracim_backend/models/context_models.py
grignards:
```suggestion
T = typing.TypeVar("T")


class ListItemsObject(typing.Generic[T]):
```

backend/tracim_backend/models/context_models.py
```diff
@@ -1637,6 +1637,11 @@ def __init__(self, page: Page) -> None:
...
+    def __init__(self, items_list: List) -> None:
+        self.items = items_list
```
grignards, line 1642:
```suggestion
    def __init__(self, items: List[T]) -> None:
        self.items = items
```

backend/tracim_backend/models/data.py
```diff
@@ -1395,6 +1395,21 @@ def recursive_children(self) -> List["Content"]:
             )
         return []
 
+    @property
```
grignards, line 1398: Why not directly constructing the content path here?

backend/tracim_backend/views/core_api/schemas.py
grignards: That's ok, we should really review the usage of digest schemas because this is quite strange that almost all the object is returned in it…

grignards (changes requested):

backend/tracim_backend/models/data.py
inkhey: somes reason:
- Where at the the level of the content itself, for me the concept of "path" doesn't make sense at this level.
- We do already have recursive_children, it make sense to have also recursive_parent.
- recursive parent may be useful for some user case.
- If we decide later that the path of a content is "workspace path + content path", it will be way more easier to do the work at the Lib level than on content level.

grignards: I think `path` does make sense at this level: when you got a file in the filesystem it is usual to be able to get its path directly I think.
I we want to keep the data classes as only "data" we could move this method to `ContentApi`.

inkhey: i suggest that we move the content_path here (which will be a content based path (not workspace) but we do keep recursive_parents method (content_path will use recursive_parents content).
There 2 differences : 
- path contain content itself
- reversed order.

grignards (approved):

raphj (approved):

frontend_lib/src/helper.js
```diff
@@ -693,8 +694,12 @@ export const buildContentPathBreadcrumbs = async (apiUrl, content, props) => {
...
             url: PAGE.WORKSPACE.CONTENT(content.workspace_id, crumb.content_type, crumb.content_id),
             label: crumb.label,
-            link: null,
-            type: BREADCRUMBS_TYPE.APP_FEATURE
+            link: (
+              <Link to={PAGE.WORKSPACE.CONTENT(content.workspace_id, crumb.content_type, crumb.content_id)}>
+                {crumb.label}
+              </Link>
+            ),
+            type: BREADCRUMBS_TYPE.CORE
```
gbonaspetti, line 702:
```suggestion
            type: BREADCRUMBS_TYPE.APP_FEATURE
```

grignards (approved):

PR #3950, opened 2020-12-14
PoC: Kanban

asf-algoo:
The Kanban PoC

## Checkpoints

**For developers**

- [ ] If there are modifications in the Tracim configuration files (eg. `development.ini`), make sure it is documented in `backend/doc/setting.md` and notify relevant people of this change

**For code reviewers**

**For testers**


frontend_app_file/src/component/Kanban.jsx
```diff
@@ -0,0 +1,386 @@
...
+require('./Kanban.styl')
+
+function moveElem (array, fromPosition, toPosition, elem) {
+  const a = array.slice()
+  a.splice(fromPosition, 1)
+  a.splice(toPosition, 0, elem)
+  return a
+}
+
+export default translate()(class Kanban extends React.Component {
```
Skylsmoi, line 25: we usually set the export default at the end of file

frontend_app_file/src/component/Kanban.jsx
```diff
@@ -0,0 +1,386 @@
...
+  handleFetchResult,
+  putFileContent,
+  removeLocalStorageItem,
+  getLocalStorageItem,
+  setLocalStorageItem
+} from 'tracim_frontend_lib'
+
+require('./Kanban.styl')
+
+function moveElem (array, fromPosition, toPosition, elem) {
```
Skylsmoi, line 18: I was about to advice using a lodash utility for that but it appears lodash doesn't have such a function.

So just don't forget to write a few unit tests for it. If you took it from stack, we usually add a comment with the source link

frontend_app_file/src/component/Kanban.jsx
```diff
@@ -0,0 +1,386 @@
...
+    super(props)
+    this.state = {
+      loading: false,
+      mustSave: false,
+      board: {
+        columns: [],
+        ...JSON.parse(props.content.raw_content || '{}')
+      },
+      isDrafAvailable: !!getLocalStorageItem(
+        'rawContent',
```
Skylsmoi, line 36: I think you can use the constant for this string

frontend_app_file/src/component/Kanban.jsx
```diff
@@ -0,0 +1,386 @@
...
+        props.content,
+        'kanban'
```
Skylsmoi, line 38: this label should exist as appName like other apps do

frontend_app_file/src/component/Kanban.jsx
```diff
@@ -0,0 +1,386 @@
...
+      // TODO handle the "Refresh?" button
+      this.updateBoard({
+        ...this.state.board,
+        ...JSON.parse(this.props.content.raw_content || '{}')
+      }, true)
+    }
+  }
+
+  handleRestoreDraft = () => {
+    this.updateBoard(
```
Skylsmoi, line 54: update board does a setState() and there is another one bellow
It is a good practice in react to try and factorize the setState to avoid multiple rerenders and side effects
Here, your updateBoard() could return the new board or you could pass any other state update to updateBoard
Those are examples, I haven't checked if they are good ideas

frontend_app_file/src/component/Kanban.jsx
```diff
@@ -0,0 +1,386 @@
...
+      )
+    )
+  }
+
+  removeCard = (card) => {
+    this.updateColumns(
+      this.state.board.columns.map(
+        col => ({
+          ...col,
+          cards: col.cards.filter(c => c !== card)
```
Skylsmoi, line 105: shouldn't you compare ids of cards ? I don't know the structure of cards, I just think this looks like you are comparing objects

frontend_app_file/src/component/Kanban.jsx
```diff
@@ -0,0 +1,386 @@
...
+        })
+      )
+    )
+  }
+
+  renderCard = (card, arg2, arg3) => {
```
Skylsmoi, line 111: arg2 and arg3 seems unused

frontend_app_file/src/component/Kanban.jsx
```diff
@@ -0,0 +1,386 @@
...
+          className='file__contentpage__statewrapper__kanban__card__description'
+          onClick={() => this.handleEditCardContent(card)}
+        >
+          {card.description}
+        </div>
+      </div>
+    )
+  }
+
+  handleAddCardClicked (column) {
```
Skylsmoi, line 133: I think the naming usually are in the active voice in Tracim
So I would advice `handleClickAddCard` instead

frontend_app_file/src/component/Kanban.jsx
```diff
@@ -0,0 +1,386 @@
...
+        description
+      })
+    }
+  }
+
+  addCardToColumn (column, card) {
+    this.updateColumns(
+      this.state.board.columns.map(
+        col => (
+          col === column
```
Skylsmoi, line 150: same question about comparing objects

frontend_app_file/src/component/Kanban.jsx
```diff
@@ -0,0 +1,386 @@
...
+        <strong onClick={() => this.handleColumnRenameClick(column)}>{column.title}</strong>
+        <IconButton
+          text=''
+          icon='plus'
+          tooltip={this.props.t('Add a card')}
+          onClick={() => this.handleAddCardClicked(column)}
+        />
+        <IconButton
+          text=''
+          icon='trash'
```
Skylsmoi, line 321: tooltip is missing

frontend_app_file/src/component/Kanban.jsx
```diff
@@ -0,0 +1,386 @@
...
+  }
+
+  handleColumnRenameClick = (column) => {
+    const newName = prompt(this.props.t('Please enter the new name of the column'), column.title)
+    if (newName) {
+      this.handleColumnRename(column, newName)
+    }
+  }
+
+  handleColumnRename = (column, title) => {
```
Skylsmoi, line 295: "handlers" are for user's action, I wouldn't call this function with the word "handle"

frontend_app_file/src/component/Kanban.jsx
```diff
@@ -0,0 +1,386 @@
...
+
+  handleColumnRemove = (column) => {
+    // console.log("handleColumnRemove", column)
+    this.updateColumns(this.state.board.columns.filter(col => col !== column))
+  }
+
+  handleCardDragEnd = (card, { fromColumnId, fromPosition }, { toColumnId, toPosition }) => {
+    // console.log("handleCardDragEnd", card, { fromColumnId, fromPosition }, {toColumnId, toPosition })
+    this.updateColumns(
+      this.state.board.columns.map(
```
Skylsmoi, line 226: to reduce this complexity, this data structure might be interesting: https://medium.com/javascript-in-plain-english/https-medium-com-javascript-in-plain-english-why-you-should-use-an-object-not-an-array-for-lists-bee4a1fbc8bd

Skylsmoi: or maybe decomposing the function

My point is, I wouldn't like to debug this function :sweat_smile:

asf-algoo: I can't use another data structure here, this is the one of the kanban component. But sure I can split this function :-)

raphj: I can't use another data structure here, this is the one of the kanban component. But sure I can split this function :-)

frontend_app_file/src/component/Kanban.jsx
raphj: This is called back by the kanban component as after a user action

frontend_app_file/src/component/Kanban.jsx
raphj: Thanks! the UI might change though

frontend_app_file/src/component/Kanban.jsx
raphj: Mhm, I like your name better but I took inspiration from something else in Tracim here. I'll rename it.

frontend_app_file/src/component/Kanban.jsx
raphj: React will batch state changes: https://stackoverflow.com/questions/33613728/what-happens-when-using-this-setstate-multiple-times-in-react-component

I'd rather keep a separation of concerns: each function does what it should do. But we can discuss.

frontend_app_file/src/component/Kanban.jsx
raphj: Yes, I planned to do this when we agree that this is the right design. Moreover, we are in app file so we need to discuss how to handle this case. This is not really an appName, but more a "local storage slot"

frontend_app_file/src/component/Kanban.jsx
raphj: I wrote this myself, good point for the unit testing of this function.

frontend_app_file/src/component/Kanban.jsx
raphj:
```suggestion
  renderCard = (card) => {
```

raphj: Yes, this is a leftover for when I was guessing the API

frontend_app_file/src/component/Kanban.jsx
raphj: Yes, It's probably better indeed as it is how the component works. I'll fix this.

raphj:
at this point, all the failing cypress tests pass locally

frontend_app_file/src/component/Kanban.jsx
Skylsmoi: the issue with setState batching is that you don't handle it. React does it for perf optimization but you can't know which setStates are going to be batched.
Therefore, you can't be sure that the ones you expect to be batched always will be. And if they don't, there will 2 rerenders that might create unexpected behavior.

I agree with the separation of concerns though, my advise would be to have a function that act as a controller that calls separate functions that each returns their own new state part. And the "controller" do a final setState with the returns of every separated functions

frontend_app_file/src/component/Kanban.jsx
Skylsmoi: indeed

frontend_app_file/src/component/Kanban.jsx
Skylsmoi: then it's not a handler, its more of a utility function

PR #4005, opened 2021-01-04
add support for preview_generator cache for non-local storage type

inkhey:
Introduce preview_generator cache for non-local storage like s3.
related to #4000 
## Checkpoints

**For developers**

- [x] If relevant, manual tests have been done to ensure the stability of the whole application and that the involved feature works
- [ ] The original issue is up to date w.r.t the latest discussions and contains a short summary of the implemented solution
- [x] Automated tests covering the feature or the fix, have been written, deemed irrelevant (give the reason), or an issue has been created to implement the test (give the link)
- [ ] Make sure that:
  - if there are modifications in the Tracim configuration files (eg. `development.ini`), they are documented in `backend/doc/setting.md`
  - any migration process required for existing instances is documented
  - relevant people for these changes are notified

**For code reviewers**

- [x] The code is clear enough
- [x] If there are FIXMEs in the code, related issues are mentioned in the FIXME
- [x] If there are TODOs, NOTEs or HACKs in code, the date and the developer initials are present

**For testers**

- [ ] Manual, quality tests have been done

Issue #4000, opened 2020-12-30
Bug: S3 storage (minio) too many preview file created for one file uploaded in tracim


PhilippeAccorsi:
## Description and expectations

I have uploaded one file in my tracim instance and clean preview folder before uploaded this file (no other file exist in this test instance).
This is a odp file and his size is near 21Mo.

After uploading this file, app file is open and I can seen de preview. But if i look in preview folder I can seen many file and space used is 5 time bigger than file stored in tracim.

~~~
total 129M
-rw-rw-rw- 1 www-data www-data 16M déc.  30 10:00 2cbd2ed344591899e76bfd43a1cab403-500x500-page0
-rw-rw-rw- 1 www-data www-data 33K déc.  30 10:00 2cbd2ed344591899e76bfd43a1cab403-500x500-page0.jpeg
-rwxrwxrwx 1 www-data www-data   0 déc.  30 10:00 2cbd2ed344591899e76bfd43a1cab403.lock
-rw-rw-rw- 1 www-data www-data   2 déc.  30 10:00 2cbd2ed344591899e76bfd43a1cab403_page_nb
-rw-rw-rw- 1 www-data www-data 16M déc.  30 10:00 2cbd2ed344591899e76bfd43a1cab403.pdf
-rw-rw-rw- 1 www-data www-data   2 déc.  30 10:00 58870ca566246a789e3ff7998c026cfc_page_nb
-rw-rw-rw- 1 www-data www-data 16M déc.  30 10:00 58870ca566246a789e3ff7998c026cfc.pdf
-rw-rw-rw- 1 www-data www-data 16M déc.  30 10:06 59043e4cd2b1f9ba0596ebf7846c49e9-500x500-page0
-rw-rw-rw- 1 www-data www-data 33K déc.  30 10:06 59043e4cd2b1f9ba0596ebf7846c49e9-500x500-page0.jpeg
-rwxrwxrwx 1 www-data www-data   0 déc.  30 10:06 59043e4cd2b1f9ba0596ebf7846c49e9.lock
-rw-rw-rw- 1 www-data www-data   2 déc.  30 10:06 59043e4cd2b1f9ba0596ebf7846c49e9_page_nb
-rw-rw-rw- 1 www-data www-data 16M déc.  30 10:06 59043e4cd2b1f9ba0596ebf7846c49e9.pdf
-rw-rw-rw- 1 www-data www-data   2 déc.  30 09:59 7acd9736718ebe72199db0603ae2c739_page_nb
[...]

PR #4009, opened 2021-01-05
add support for preview_generator cache for non-local storage type

asf-algoo:
Introduce preview_generator cache for non-local storage like s3.
related to #4000 
## Checkpoints

**For developers**

- [x] Make sure that:

**For code reviewers**

**For testers**

previous PR: https://github.com/tracim/tracim/pull/4005

### DIFF
--- a/backend/tracim_backend/lib/core/content.py
+++ b/backend/tracim_backend/lib/core/content.py
@@ -8,6 +8,7 @@ import typing
 from depot.io.interfaces import StoredFile
 from depot.io.utils import FileIntent
 from depot.manager import DepotManager
+import filelock
 from preview_generator.exception import UnavailablePreviewType
 from preview_generator.exception import UnsupportedMimeType
 from preview_generator.manager import PreviewManager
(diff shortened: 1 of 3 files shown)


grignards (approved):

backend/tracim_backend/lib/core/content.py
```diff
@@ -735,23 +736,53 @@ def get_one_revision(
...
+            try:
+                # HACK - G.M - This mecanism is inefficient because it
+                # generate a temporary file each time
+                # Improvement need to be made in preview_generator itself.
+                # to handle more properly theses issues
+                # We do rely on consistent path based on gettemdir(),
+                # normally /tmp to give consistent path, this is a quick fix which does
+                # not need change in preview-generator.
+                # note: this base path is configurable through env var according to python doc
+                # https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir
```
raphj, line 777:
```suggestion
        # FIXME - G.M - This will create a lockfile for
        # each depot file we need (each content revision)
        # This file will NOT be removed at the end on Linux (FileLock use flock):
        # see  https://stackoverflow.com/questions/17708885/flock-removing-locked-file-without-race-condition
        # some investigation needs to be conducted to see if another solution is possible avoiding creating
        # too many files.
        with filelock.FileLock(lockfile_path):
            try:
                # HACK - G.M - This mechanism is inefficient because it
                # generates a temporary file each time
                # Improvements need to be made in preview_generator itself
                # to handle more properly these issues.
                # We rely on consistent path based on gettemdir(),
                # normally /tmp to give consistent path, this is a quick fix which does
                # not need any change in preview-generator.
                # note: this base path is configurable through an environment variable according to the Python doc
                # https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir
```
Please put the dates in the comments and open an issue that you mention here to handle this FIXME.

raphj (changes requested):

backend/tracim_backend/lib/core/content.py
inkhey: done.

raphj (approved):

grignards (approved):

Issue #3999, opened 2020-12-30
Bug: Link at frontend_lib causes white page using servdev


gbonaspetti:
## Description and expectations

`frontend_lib/helper.js` and `frontend_lib/PageTitle.jsx` have _Link_ imports, but `frontend_lib` is not inside the _Router_, this causes some errors like `Error: Invariant failed`, `The above error occurred in the <Context.Consumer> component`, `Cannot update during an existing state transition (such as withinrender). Render methods should be a pure function of props and state.`, when using servdev.

### How to reproduce

1. Activate servdev
2. At your browser go to a space's dashboard

### Version information

- Tracim version: at least v3.4.1
- Web browser: at least Google Chrome
- Device / System: PC

________________________________________________

Skylsmoi:
The general issue is about having different instances of react-router-dom in the project.

I have tested:
- passing the link component as parmeter where it is required
- with and without putting react-router-dom in forntend_vendor
- removing the call to frontend_lib and frontend_vendor in the index.html of servdev
- simplifying the app to the maximum by only calling Breadcrumb component in frontend/src/index.js
  - even removing every other HOC from frontend/src/index.js

None of this worked.

Side note: importing Breadcrumbs directly through `tracim_frontend_lib/src/component/Breadcrumbs/Breadcrumbs.jsx` worked. Which is obvious because there isn't any external library in that case.

Skylsmoi:
Solution that works (currently without any apps included in Tracim, only frontend_lib and frontend):

In frontend_lib, setting react-router-dom as an external dependency.

This is the case by the `.external` key in `frontend_lib/webpack.config.js` for the `standalone` build and by putting it in the frontend_vendors bundle (adding 'react-router-dom' in `frontend_vendors/src/index.js`) for the `optimized` build

That way, there always is only once instance of react-router-dom which is:
- for standalone build: the one in frontend/
- for optimized build: the one in frontend_vendors/

This working solution hasn't been tested with apps yet.

Skylsmoi:
List of others fixes done during investigation:
- use BrowserRouter instead of Router in frontend/
- remove any import and usage of react-router, only use react-router-dom
- removed unused `optimized-dev` case from frontend_lib/build-script.sh
- fixed frontend_lib/build-script.sh error output to standard output instead of redirecting error to standard
- removed .dev extension in frontend_lib builds that where never used
  - because, in package.json, the main property is `tracim_frontend_lib.lib.js`
[...]

PR #4046, opened 2021-01-12
Fix/3999  fix servdev

Skylsmoi:
https://github.com/tracim/tracim/issues/3999

## Checkpoints

**For developers**

- [x] Make sure that:

**For code reviewers**

**For testers**


Skylsmoi:
list of changes:
- new file `build_full_frontend_standalone.sh` which build every apps in standalone mode and copy the output to frontend/dist/app
- frontend/dist/index.html (used by servdev) now uses only .standalone files
- removed react-router everywhere
  - now only use react-router-dom
- harmonized react-router-dom version number
- frontend now uses BrowserRouter instead of Router
- removed the .dev extension of all script
  - the .dev extension is now .standalone and is generated by `build` and `build-dev` and `build_full_frontend_standalone.sh`
- fix servdev for all apps
  - updated index.html
  - added missing import 'regenerator-runtime/runtime' in some index.dev.js
  - update debug.js.sample
  - harmonized servdev usage for all apps
- fix script building in frontend_lib
  - in `build-script.sh`
    - removed unused case
    - fix error output
    - reduced variables and parameters
    - now create a .optimized.js output
- frontend_lib Breadcrumbs component is now the only one to generate the `<Link />` components for the breadcrumb
  - refactored all breadcrumb generation with the new props
    - which imply the refactor of the APPEND_BREADCRUMB TLM
- frontend_lib, added 'react-router-dom' as externals
  - this is the fix that allows frontend_lib to create `<Link>` in its components
  - making Breadcrumbs component the only one to generate the `<Link />` is the fix that allows apps to update the breadcrumbs
- frontend_vendors
  - react-router removed and react-router-dom added

build_full_frontend_standalone.sh
```diff
@@ -0,0 +1,126 @@
+#!/bin/bash
```
raphj, line 1: Could we add an option to build_full_frontend instead of replicating its features?

I can help if needed.

raphj:
So react-router-dom is always external even in standalone builds?

frontend_app_admin_workspace_user/dist/index.html
```diff
@@ -55,20 +104,32 @@
...
 
   <div id='content'></div>
 
-  <script type='text/javascript' src='./app/tracim_frontend_vendors.js'></script>
-  <script type='text/javascript' src='./app/tracim_frontend_lib.lib.js'></script>
+  <!--  <script type='text/javascript' src='./app/tracim_frontend_vendors.js'></script>-->
+  <!--  <script type='text/javascript' src='./app/tracim_frontend_lib.lib.js'></script>-->
+
+  <!--  <script type='text/javascript' src='../frontend/src/util/appInterface.js'></script>-->
+  <!--  <script type='text/javascript' src='../frontend/src/util/tinymceInit.js'></script>-->
```
gbonaspetti, line 126: Are these comments intentional or forgotten?

frontend_app_admin_workspace_user/src/container/AdminWorkspaceUser.jsx
```diff
@@ -219,25 +218,21 @@ export class AdminWorkspaceUser extends React.Component {
...
-      page = PAGE.ADMIN.USER
+      pageUrl = PAGE.ADMIN.USER
       label = props.t('User account management')
     }
     breadcrumbsList.push({
-      link: <Link to={page}>{label}</Link>,
+      link: pageUrl,
       type: BREADCRUMBS_TYPE.CORE,
       label: label
     })
```
gbonaspetti, line 234: `isALink: true`  is not necessary in this case?

frontend_app_admin_workspace_user/src/index.js
```diff
@@ -13,6 +13,7 @@ const appInterface = {
   renderAppFullscreen: data => {
     document.getElementById(data.config.domContainer).classList.add('fullWidthFullHeight')
 
+    // INFO - CH - 2020-01-11 - User Router instead of BrowserRouter because BrowserRouter doesn't have a history props
```
gbonaspetti, line 16: At `/index.dev.js` you used `const history = require('history').createBrowserHistory()` can't you do the same here?

```suggestion
    // INFO - CH - 2020-01-11 - Use Router instead of BrowserRouter because BrowserRouter doesn't have a history props
```

build_full_frontend_standalone.sh
Skylsmoi: it might be better, but I chose not to edit the script to avoid any side effects in production

it also keep things a bit simpler that way

Skylsmoi:
for optimized build, yes. It is handled by frontend_vendors

for standalone, it is only external for frontend_lib but still bundled in each standaslone build that requires it

frontend/src/index.js
```diff
@@ -17,17 +17,15 @@ require('./css/index.styl')
 require('./util/appInterface.js')
 require('./util/tinymceInit.js')
 
-// export const history = require('history').createBrowserHistory()
-
 ReactDOM.render(
   <Provider store={store}>
-    <Router history={history}>
+    <BrowserRouter history={history}>
```
gbonaspetti, line 22: I'm not sure that I understood, at `frontend_app_admin_workspace_user/src/index.js` you add the comment `// INFO - CH - 2020-01-11 - User Router instead of BrowserRouter because BrowserRouter doesn't have a history props` but here you use BrowserRouter and history.

gbonaspetti (changes requested):

frontend/src/index.js
Skylsmoi: it's a mistake, I'm fixing it.
Only frontend must use BrowserRouter actually. Because apps needs to get their history from the frontend, so they must use Router

frontend_app_admin_workspace_user/src/index.js
Skylsmoi: no, I use createHistory in index.dev.js because I don't have the history that comes from the frontend.
Which index.js has

Skylsmoi: will fix the typo though

frontend_app_admin_workspace_user/src/container/AdminWorkspaceUser.jsx
Skylsmoi: is should yes, will fix

frontend_app_admin_workspace_user/dist/index.html
Skylsmoi: mmh intentional.
To mark the fact that we deliberately don't use theses scripts in standalone mode.

But I should add some comment for explaination

frontend/src/index.js
Skylsmoi: actually, there might be a small misconception here

Skylsmoi: ok the issue was in frontend, it was using a BrowserRouter with a history props, which makes no sense.

Skylsmoi: fixed

frontend_app_admin_workspace_user/src/index.js
Skylsmoi: typo fixed

frontend_app_admin_workspace_user/src/container/AdminWorkspaceUser.jsx
Skylsmoi: fixed

frontend_app_admin_workspace_user/dist/index.html
Skylsmoi: fixed

Skylsmoi:
regarding tests:
- I have updated the tests of the component Breadcrumbs
- I don't know how to test that webpack-dev-server works to ensure non regression and I'm not sure there is a way

build_full_frontend_standalone.sh
raphj: I'm concerned by the duplication that may cause maintenance issues, keeping both in sync may be harder.

raphj:
> for optimized build, yes. It is handled by frontend_vendors
> 
> for standalone, it is only external for frontend_lib but still bundled in each standalone apps builds that requires it

Okay. I'll need an explanation on why it is bundled for frontend_lib but not apps.

gbonaspetti (approved):

build_full_frontend_standalone.sh
```diff
@@ -0,0 +1,126 @@
...
+yarn workspace tracim_frontend_lib run build$dev && loggood "success" || logerror "Could not build tracim_frontend_lib"
+
+#for app in "$DEFAULTDIR"/frontend_app_*; do
+#	if [ -f "$app/.disabled-app" ]; then
+#		log "Skipping $app because of the existence of the .disabled-app file"
+#	else
+#		cd "$app" || exit 1
+#		./build_*.sh $appdev || logerror "Failed building $app."
+#	fi
+#done
```
grignards, line 56: If this loop can work it would make the script easier to understand/maintain I think.
If not maybe remove it?

frontend/dist/index.mak
```diff
@@ -111,14 +110,19 @@
     <script type="text/javascript" src="/assets/bootstrap/popper-1.12.3.js?token=${cache_token}" nonce="${csp_nonce}"></script>
     <script type="text/javascript" src="/assets/bootstrap/bootstrap-4.0.0-beta.2.js?token=${cache_token}" nonce="${csp_nonce}"></script>
 
+    <script type="text/javascript" src="/assets/tracim.app.js?token=${cache_token}" nonce="${csp_nonce}"></script>
+
```
grignards, line 114: What is the reason behind the move of this <script> tag? Adding a note explaining that this script must be loaded here could avoid unwanted changes later on (if the order is important for example).

raphj (approved):

frontend_app_thread/src/index.dev.js
```diff
@@ -2,22 +2,22 @@ import React from 'react'
 import 'core-js/stable'
 import 'regenerator-runtime/runtime'
 import ReactDOM from 'react-dom'
-// import Thread from './container/Thread.jsx'
+import Thread from './container/Thread.jsx'
 import { debug } from './debug.js'
 import { LiveMessageManager } from 'tracim_frontend_lib'
-import PopupCreateThread from './container/PopupCreateThread.jsx'
+// import PopupCreateThread from './container/PopupCreateThread.jsx'
```
grignards, line 8: Maybe remove this line?

frontend_lib/src/helper.js
```diff
@@ -679,37 +678,21 @@ export const scrollIntoViewIfNeeded = (elementToScrollTo, fixedContainer) => {
...
-      GLOBAL_dispatchEvent({
-        type: CUSTOM_EVENT.ADD_FLASH_MSG,
-        data: {
-          msg: props.t('Error while getting breadcrumbs'),
-          type: 'warning',
-          delay: undefined
-        }
-      })
+      console.error('Error getting breadcrumbs data', fetchGetContentPath)
+      throw new Error('')
```
grignards, line 696: Shouldn't the exception also contain the message?

frontend_vendors/src/index.js
```diff
@@ -11,6 +11,8 @@ module.exports = {
   react: require('react'),
   'react-dom': require('react-dom'),
   'react-image-lightbox': require('react-image-lightbox'),
-  'react-router': require('react-router'),
+  // INFO - CH - Keep the line bellow or Link components will fail
```
grignards, line 14:
```suggestion
  // INFO - CH - Keep the line below or <Link> components will fail
```

build_full_frontend_standalone.sh
Skylsmoi: it doesn't work, you are right, I forgot to remove it. It was a copast from build_full_frontend.sh

frontend_app_thread/src/index.dev.js
Skylsmoi: I chose to keep it because there is no way of switching between the components PopupCreateThread and Thread other than commenting one or the other in the ReactDOM.render().

It's not the cleanest but it's something that was here before my refactor

frontend_lib/src/helper.js
Skylsmoi: It should yes, but since we don't have any exception management, I didn't know what to put in it and it slipped my mind.

My expectations would be to something like an exception code to be handled by a generic process. 
Only consistent possibility here would be to put a generic message like `Breadcrumb exception` ?

frontend/dist/index.mak
Skylsmoi: It was to fix an inconsistency during my tests, but I shouldn't have kept. Will redo some tests without the change and remove it if it's ok

Skylsmoi: After some tests I'm a bit confused. The issue I ran into is that, with the previous position of this line (in the middle of the scripts tags), sometimes, apps aren't loaded when frontend tries to call them. Using pushpin server.

I tried back on develop and I'm having the same issue. I can reproduce it consistently (not 100% though).

So I tried the same on unstable.tracim.fr but I'm unable to reproduce it.

Skylsmoi: So I'm reverting this fix
But we might need to keep an eye on it

Skylsmoi: fix

frontend_vendors/src/index.js
Skylsmoi: fixed

build_full_frontend_standalone.sh
Skylsmoi: fixed

grignards (approved):

frontend_lib/src/helper.js
Skylsmoi: added the same message as console.error

Skylsmoi: fixed

frontend_app_thread/src/index.dev.js
Skylsmoi: won't fix

Skylsmoi:
For manual tests:
- Check some app's servdev (don't forget to update debug.js)
- run build_full_frontend_standalone.sh then run frontend servdev and check it works fine
- in the frontend's servdev, open an app feature for a content and check that the breadcrumbs is well updated
- check the breadcrumbs for some other pages

grignards (approved):

gbonaspetti:
Tested on branch

Issue #4102, opened 2021-01-28
Bug: Hide the profile edit button if the data json is empty


gbonaspetti:
## Description and expectations

By default, Tracim doesn't use custom properties and the endpoint related to that returns an empty json.
Frontend handle the empty json in the same way that others, so we have a "Edit" button at Information and/or Personal Page that gives the impression of doing nothing.

### Expected
- [x] Do not show the edit button if json is empty

### How to reproduce

1. Do not set any custom properties configuration
2. Go to your profile
3. Click at edit

### Version information

- Tracim version: v3.5
- Web browser: any
- Version: any
- Device / System: any

raphj:
tested on branch, but the problem related to the required field of json schemas really needs to be addressed.

inkhey:
tested on branch, does work correctly in all case:
- no json schema (get the `e.required is undefined` issue but this is unrelated to edit button.)
- json schema with no value in each session
- json schema with value in just one section
- json schema with value in just the other one section

PhilippeAccorsi:
Done.

PR #4112, opened 2021-01-28
Fix/4102  hide edit button when empty json at profile

asf-algoo:
#4102 

## Checkpoints

**For developers**

- [x] Make sure that:

**For code reviewers**

**For testers**

### DIFF
--- a/frontend/src/container/PublicProfile.jsx
+++ b/frontend/src/container/PublicProfile.jsx
@@ -371,12 +371,14 @@ export class PublicProfile extends React.Component {
     }
   }
 
-  isPublicProfileEditable = (connectedUser, publicProfileId, profileObject) => {
+  isPublicProfileEditable = (connectedUser, publicProfileId, profileObject, schemaObject) => {
+    if ((schemaObject && Object.keys(schemaObject).length === 0) ||
+      (schemaObject && schemaObject.properties && Object.keys(schemaObject.properties).length === 0)
+    ) return false
     const isConnectedUserOnHisOwnProfile = connectedUser.userId === publicProfileId
     const isUserAdmin = connectedUser.profile === profileObject.administrator.slug
-    const isInformationSchemaObjectEmpty = Object.keys(this.state.informationSchemaObject).length === 0
 
-    return (isConnectedUserOnHisOwnProfile || isUserAdmin) && !isInformationSchemaObjectEmpty
+    return isConnectedUserOnHisOwnProfile || isUserAdmin
   }
 
   handleCloseUploadPopup = () => this.setState({ displayUploadPopup: undefined })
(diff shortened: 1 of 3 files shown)


raphj (approved):

frontend/src/container/PublicProfile.jsx
```diff
@@ -374,8 +374,9 @@ export class PublicProfile extends React.Component {
   isPublicProfileEditable = (connectedUser, publicProfileId, profileObject) => {
     const isConnectedUserOnHisOwnProfile = connectedUser.userId === publicProfileId
     const isUserAdmin = connectedUser.profile === profileObject.administrator.slug
+    const isInformationSchemaObjectEmpty = Object.keys(this.state.informationSchemaObject).length === 0
```
Skylsmoi, line 377: shouldn't you also test the personnalDataSchemaObject ?

Also, I'd like this function to stay pure

And finally, you could use and early return if isInformationSchemaObjectEmpty is true
It would simplify the final return

Skylsmoi (changes requested):

frontend/src/container/PublicProfile.jsx
gbonaspetti: When you said `personalDataSchemaObject` I think you meant `informationDataSchema`, correct if I'm wrong. At this case, the object can be empty even if `informationSchemaObject` is not empty. The first time that the user go to the profile, for example.

I'll change the rest.

Skylsmoi: no, I mean that `isPublicProfileEditable` is used by both form (information and personal page) and I suppose it should return a result for each form based on its own schema

Skylsmoi: Information form: https://github.com/tracim/tracim/blob/develop/frontend/src/container/PublicProfile.jsx#L448

Personal page form: https://github.com/tracim/tracim/blob/develop/frontend/src/container/PublicProfile.jsx#L472

gbonaspetti: ah, ok, will fix

frontend/src/container/PublicProfile.jsx
```diff
@@ -371,12 +371,14 @@ export class PublicProfile extends React.Component {
     }
   }
 
-  isPublicProfileEditable = (connectedUser, publicProfileId, profileObject) => {
+  isPublicProfileEditable = (connectedUser, publicProfileId, profileObject, schemaObject) => {
```
Skylsmoi, line 374: seeing the usage of the function in CoverImage and Avatar makes me realize that the function now to more that what it's supposed to do.

As seen with you, I advise to split the function into to parts, one to check the users ids and profile and one to check if the schema is empty.
And use const in the render to put the different values for the 3 usages:
- CoverImage and Avatar which are the same value
- Information
- Personal Page

Skylsmoi (changes requested):

frontend/src/container/PublicProfile.jsx
gbonaspetti: updated

Skylsmoi (approved):

Issue #4083, opened 2021-01-25
Feat: Add search page interface for advanced search


gbonaspetti:
## Feature description and goals

The advanced search will be improved by adding some result filters and more results possibilities, for this we will add a new interface for advanced search.
This issue will not implement the filters menu (see https://github.com/tracim/tracim/issues/4097) nor the different types of results (see https://github.com/tracim/tracim/issues/4085). Just update the interface and some interactions for now.

- [x] Add search input at the search screen.
- [x] Update `N best results for "X"` message to `Showing N of M results` with lightDarkGrey color.
- [x] Update title adding the keyword.
- [x] After modifying a keyword the user must click on the search icon or press the Enter key to validate. After this validation, the title and both input are updated with the new keywords.
- [x] Allow the click at search icon even if there is no keyword
    - Tooltip: Search
    -  Hover: fontColor
    - Cursor: pointer
- [x] Add a filter button in the top-right corner
    - On click it'll open the filter menu
    - In sm and xs screens the label should be hidden
    - Icon: fas fa-sliders-h
    - Label: Filter
    - Tooltip: Search filters
- [x] Add a filter menu: 
    - For now just the rectangle, the title and the close option
        - Width: 300px to md, lg and xl screens and 100% to sm and xs screens
        - Background color: offWhite
        - Border color: grey
        - Icon: fas fa-sliders-h
        - Title: Search filters
- [x] By default the filter's menu is closed.
- [x] Update loader if needed

## Prerequisites

https://github.com/tracim/tracim/issues/4094 (see https://github.com/tracim/tracim/issues/4083#issuecomment-767553590)

## Translations

|English|French|Portuguese|
|-|-|-|
|Filter|Filtrer|Filtrar|
|Search filters|Filtres de recherche|Filtros de pesquisa|
|Search|Rechercher|Pesquisar|
|Showing {{number_display_results}} of {{number_total_results}} results|{{number_display_results}} résultats affichées sur {{number_total_results}}|Mostrando {{number_display_results}} de {{number_total_results}} resultados|
|Results for "{{keywords}}"|Résultats pour "{{keywords}}"|Resultados para "{{keywords}}"|

## Interface
_Obs: These interface examples are only templates, not everything will be developed in this issue and some details can be changed according to the implementation._

- Button view
    - md, lg, xl screens
    ![search_page](https://user-images.githubusercontent.com/35422692/105695965-fdf1fd00-5f02-11eb-86a7-c97ab9b8521e.png)
[...]

PR #4165, opened 2021-02-09
Feat/4083  add advanced search interface

asf-algoo:
#4083 
- Added AdvancedSearch component, for now it has the same content list as the simple search.
- Added a Search folder at frontend/component to put all things related to the search
- Added a SearchFilterMenu component (empty for this issue)
    - to make the responsive behavior I added a `advancedSearch__openMenu` class and use it to hide things at a small screen
- Fixed some FA5 issues that I founded during development
## Checkpoints

**For developers**

Dev note: Travis configuration uses simple search.
- [x] Make sure that:

**For code reviewers**

**For testers**

To test: just see if the interface with the elastic search has the elements mentioned at the issue description and with simple search we have the same interface that before. The filters menu height is not set because it doesn't have any content yet.


frontend/src/container/AdvancedSearch.jsx
```diff
@@ -0,0 +1,303 @@
...
+  componentDidMount () {
+    this.setHeadTitle()
+    this.buildBreadcrumbs()
+    this.loadSearchUrl()
+  }
+
+  componentDidUpdate (prevProps) {
+    const { props } = this
+    const prevSearchedKeywords = qs.parse(prevProps.location.search).q
+    const currentSearchedKeywords = parseSearchUrl(qs.parse(props.location.search)).searchedKeywords
```
Skylsmoi, line 64: why isn't it the same function call in these 2 lines ?
I would assume calling the same function with only the parameter changing from prevProps to props

especially because parseSearchUrl convert the q query param to searchedKeywords

frontend/src/container/AdvancedSearch.jsx
```diff
@@ -0,0 +1,303 @@
...
+      searchObject.showDeleted,
+      searchObject.showActive
+    ))
+
+    switch (fetchGetSearchedKeywords.status) {
+      case 200:
+        props.dispatch(setCurrentNumberPage(NEXT_PAGE))
+        props.dispatch(appendSearchResultsList(fetchGetSearchedKeywords.json.contents))
+        this.setState({ totalHits: fetchGetSearchedKeywords.json.total_hits })
+        props.history.push(PAGE.SEARCH_RESULT + '?' + qs.stringify({ ...qs.parse(props.location.search), p: NEXT_PAGE }, { encode: true }))
```
Skylsmoi, line 146: I think this might be a bit of a misconception.

You are updating the local state then updating the route. It should be the other way around, the interface should react to the new route and trigger the new search when the route change.

Currently: click "see more" > trigger api GET search result > update result > update route for consistency
Expected: click "see more" > update route.
And the updated route triggers the rerender of the component that will GET the new search result

This is all theoretical, if it's an update of the old search process maybe it's too big of a change to do it now. But that is something important to keep in mind when building apps

frontend/src/container/AdvancedSearch.jsx
```diff
@@ -0,0 +1,303 @@
...
+
+  handleClickFilterMenu = () => this.setState(prev => ({ isFilterMenuOpen: !prev.isFilterMenuOpen }))
+
+  handleClickSearch = searchedKeywords => {
+    const { props } = this
+    const FIRST_PAGE = 1
+
+    props.dispatch(setSearchedKeywords(searchedKeywords))
+    props.dispatch(setCurrentNumberPage(FIRST_PAGE))
+    props.history.push(PAGE.SEARCH_RESULT + '?' + qs.stringify({
```
Skylsmoi, line 162: This is the same issue with routing, clicking the search button should only do a history.push and the component should update redux on reaction to the new route

Skylsmoi: this won't be "required changed", and we can talk about it a bit more if you need

frontend/src/container/AdvancedSearch.jsx
gbonaspetti: No specific reason, I will fix that.

frontend/src/container/AdvancedSearch.jsx
gbonaspetti: All this content part I just copy+paste from the simple search, I don't think it's too big to change it now (in both components), I'll fix it.

Skylsmoi (approved):

frontend/src/container/Tracim.jsx
```diff
@@ -526,7 +526,7 @@ export class Tracim extends React.Component {
 
           <Route
             path={PAGE.SEARCH_RESULT}
-            component={true // TODO - props.system.config.search_engine === 'elasticsearch'
+            component={props.system.config.search_engine === 'elasticsearch'
```
Skylsmoi, line 529: 'elasticsearch' should be a constant

Skylsmoi (changes requested):

Skylsmoi (approved):

Skylsmoi (approved):

Issue #4180, opened 2021-02-16
Feat: Add different types of search


gbonaspetti:
## Feature description and goals

Now it will be possible to search not only for contents but also in spaces and users for advanced search. Some development has already been done on problem #4085 and should be finished here.

- [x] Add radio buttons for the three different types of result.
- [x] The frontend must search for the 3 types in parallel so that there is no loading time each time the user changes the type.
    - Add three lists at Redux at root
- [x] By default the page should display the results of the content search
- With sm and xs screens:
    - [x] The breadcrumbs should be hidden
    - [x] The icon type/avatar should have fontSizeMedium/AVATAR_SIZE.MINI size
- With xl, lg and md screens:
    - [x] The icon type/avatar should have fontSizeTitle/AVATAR_SIZE.SMALL size
- [x] Add unit tests to advanced search (content type included), if needed
- [x] Deduplicate currentSearch ifs

### Users list

- [x] Columns at xl, lg, md and sm screens:
    - Avatar
    - Full Name
    - Information (fontSizeSmall)
        - Username
- [x] Columns at xs screens:
    - Avatar and Full Name
- [x] Clicking anywhere redirects to the user's profile
- [x] If and only if the logged user is an administrator, the frontend will have to get email for each result to display it. If user isn't an admin, the email is never displayed
    - Update: the e-mail will not be displayed, even for admin 
- The avatar is not send by backend
  - but the front will have all the required information to generate the component Avatar (id, public name ...)

### Spaces list

- [x] Columns at any screen:
    - Type 
    - Name (if there's not enough space at sm and/or xs screens type and name should be put in the same column)
    - Information (fontSizeSmall - and if there's not enough space at  xs screens this column should be hidden)
        - Number of contents
        - Number of members
- Click on :
    - [x] a space that the user is a member redirects to the home page of the space
    - [x] a space that the user is not a member redirects to the page to join the space with the same keyword entered in the search (compare with the list of spaces that the logged user is a member of)
- [x] Each type must have its name written under the icon on the lg and xl screens in fontSizeSmall size

## Prerequisites

- https://github.com/tracim/tracim/issues/4095

## Translations

|English|French|Portuguese|
|-|-|-|
|Contents|Contenus|Conteúdos|
|Users|Utilisateurs|Utilizadores|
|Spaces|Espaces|Espaços|
|Type|Type|Tipo|
|Title and Path|Titre et chemin|Título e caminho|
|Last modification|Dernière modification|Última modificação|
[...]

PR #4225, opened 2021-02-26
Feat/4180  add different types of search

asf-algoo:
#4180 

## Checkpoints

**For developers**

- [x] Make sure that:

**For code reviewers**

**For testers**


raphj (approved):

frontend/src/container/AdvancedSearch.jsx
```diff
@@ -333,21 +282,9 @@ export class AdvancedSearch extends React.Component {
...
-      if (searchObject.searchType === ADVANCED_SEARCH_TYPE.USER) {
-        currentSearchLength = props.userSearch.resultList.length
-      }
-
-      if (searchObject.searchType === ADVANCED_SEARCH_TYPE.SPACE) {
-        currentSearchLength = props.spaceSearch.resultList.length
-      }
-    */
-    this.getSearchResult(searchObject, currentSearchLength)
+    this.setState({ searchType: searchObject.searchType })
```
raphj, line 285: Maybe check here if searchType is in Object.values(ADVANCED_SEARCH_TYPE). If not, you could fallback to contents, and maybe push a new "correct" location.

gbonaspetti: fix

backend/tracim_backend/lib/search/elasticsearch_search/elasticsearch_search.py
```diff
@@ -681,10 +682,8 @@ def search_user(
                 "range", newest_authored_content_date=newest_authored_content_date_range
             )
 
-        if size:
-            search = search.extra(size=size)
-        if page_nb:
-            search = search.extra(from_=self.offset_from_pagination(size, page_nb))
+        if size and page_nb:
```
inkhey, line 685: I don't understood this code, you should keep possibility to set size (and not just offset) and allow 0 size too (for facet only case)

backend/tracim_backend/lib/search/elasticsearch_search/elasticsearch_search.py
```diff
@@ -739,10 +750,8 @@ def search_workspace(
             search = search.filter("terms", member_ids=member_ids)
         if not show_deleted:
             search = search.exclude("term", is_deleted=True)
-        if size:
-            search = search.extra(size=size)
-        if page_nb:
-            search = search.extra(from_=self.offset_from_pagination(size, page_nb))
+        if size and page_nb:
```
inkhey, line 753: Same here, i don't understood this change. It make more sense to have similar code as for content here.

inkhey (changes requested):

backend/tracim_backend/views/search_api/schemas.py
```diff
@@ -322,6 +360,19 @@ class WorkspaceSearchQuerySchema(marshmallow.Schema):
...
+        required=False, default=10, validate=strictly_positive_int_validator
+    )
+    page_nb = marshmallow.fields.Int(
+        required=False, default=1, validate=strictly_positive_int_validator
+    )
+    show_deleted = marshmallow.fields.Int(
+        example=0,
+        default=0,
+        description="if set to 1, then show deleted contents."
+        " Default is 0 - hide deleted content",
```
inkhey, line 373:
```suggestion
        description="if set to 1, then show deleted workpaces."
        " Default is 0 - hide deleted workspaces",
```
?

backend/tracim_backend/lib/search/elasticsearch_search/elasticsearch_search.py
grignards: ? setting `size=0/None` had no effect, even with the previous code. So this means it wasn't possible to only get facets and not results. By the way those two values (`page_nb` and `size`) are enforced to be strictly positive by the schema and defaults to `1`/`10`. So I'm even wondering whether we should care about false values here…

inkhey: According to content search specification (last version), we should now allow 0 for size, as this permit (that's a elasticsearch feature) to have facet only search (this work with content search so i see no reason this should not work here). Also, as far i understood your fix, you remove the ability to set size, which is not something we want.

see `test_api___elasticsearch_search_ok__wildcard__facet_only` for example.

grignards: Corrected, I wasn't aware of the `size=0` case.

inkhey (approved):
Backend seems ok, don't know for the frontend part.

raphj (approved):

raphj:
looks good to me for everything.

Issue #4912, opened 2021-09-13
Bug:  subdomain email are not valid through backend email validation


inkhey:
## Description and expectations

Email validation in tracim backend doesn't validate email of subdomain like `me@my.subdomain.fr` whereas `me.@mydomain.fr` work properly. This kind of email should be considered as valid and work like any valid email address.

### How to reproduce

1. Connect to tracim
2. Change email to a subdomain email in user config panel.

### Version information

- Tracim version: =<3.9.2 (found in 3.9.2 but all older version are most probably concerned).

## Diagnostic

This problem concern all place we do email validation using Email validator of Marshmallow. So, it make not possible to create user with this email, but also make not possible to modify the user mail and broke some email based feature like file sharing with this kind of email.

The problem is related to the email validator of the library marshmallow, 
there is no current issue on it upstream but a similar problem was found on url validation and there is a comment about a possible issue related to email validation:
https://github.com/marshmallow-code/marshmallow/issues/664#issuecomment-340264736

## Extra information

:warning:  We need to decide how we deal with the support transition:

- Problem doesn't seems to be solved to latest marshmallow versions. So propose an upstream patch seems the way to go.
- Tracim currently doesn't support marshmallow 3+, so a fix in the source code seems to be needed anyway.

inkhey:
https://github.com/tracim/tracim/issues/4056 seems to be the same issue. We need to verify it after fixing this one.

lebouquetin:
I suggest to fix this as a maintainance issue.

lebouquetin:
@inkhey could you make a bugfix fast? (I mean ... it's not too much work, is it?). Same question @gbonaspetti for #4056.

gbonaspetti:
> @inkhey could you make a bugfix fast? (I mean ... it's not too much work, is it?). Same question @gbonaspetti for #4056.

https://github.com/tracim/tracim/issues/4056 it's quick.

buxx:
Note : Think to advise here : https://community.tracim.fr/ui/workspaces/157/contents/thread/18010

inkhey:
> @inkhey could you make a bugfix fast? (I mean ... it's not too much work, is it?). Same question @gbonaspetti for #4056.

I did an hotfix here : https://github.com/tracim/tracim/pull/4913 and did some manual try on it as a fast bugfix.
[...]

Issue #4056, opened 2021-01-15
Bug: its not possible to share file when domain of email have more than one dot


PhilippeAccorsi:
## Description and expectations

I tried to test if shared file email sended by Tracim is considering like a spam.
To test that, I tried to use https://www.mail-tester.com/ service, but this service give me an email like this `test-ri8itwf0j@srv1.mail-tester.com`.

We need to accept more complex email in Tracim.

For information: this control is just done in shared feature. No control of email structure to create a new user.

### How to reproduce

- open app file
- open share tab
- add an email with more than one dot in domain name

### Version information

- Tracim version: all tracim since july-august 2019

## Screenshots

![Screenshot_2021-01-15 (99+) Login · preprod-algoo](https://user-images.githubusercontent.com/25848257/104750788-da85c000-5754-11eb-875b-c302ad77af35.png)

________________________________________________

## Diagnostic

Frontend use this file `frontend_lib/src/helper` (line 423) to check email structure
```js
// INFO - GB - 2019-07-31 - This function check if the email has three parts arranged like somethig@something.somethig
export const checkEmailValidity = email => {
  const parts = email.split('@')
  if (parts.length !== 2) return false

  const domainParts = parts[1].split('.')
  return domainParts.length === 2
}
```

PhilippeAccorsi:
Related to #2030

MariePrandstetter:
tested on branch

PR #4913, opened 2021-09-13
Fix/4912 allow subdomain emails

inkhey:
related to #4912 and #4056

## Checkpoints

**For developers**

- [x] Make sure that:

**For code reviewers**

**For testers**


backend/tracim_backend/app_models/email_validators.py
```diff
@@ -0,0 +1,41 @@
...
+    USER_REGEX = re.compile(
+        r"(^[-!#$%&'*+/=?^`{}|~\w]+(\.[-!#$%&'*+/=?^`{}|~\w]+)*\Z"  # dot-atom
+        # quoted-string
+        r'|^"([\001-\010\013\014\016-\037!#-\[\]-\177]' r'|\\[\001-\011\013\014\016-\177])*"\Z)',
+        re.IGNORECASE | re.UNICODE,
+    )
+
+    DOMAIN_REGEX = re.compile(
+        # domain
+        r"(?:[A-Z0-9](?:[A-Z0-9-]{0,61}[A-Z0-9])?\.?)+" r"(?:[A-Z]{2,6}|[A-Z0-9-]{2,})\Z"
```
inkhey, line 19: Simple fix inspirated by this commit: https://github.com/sduthil/marshmallow/commit/126112f3c2f6f0d7ba35e9f960529e8a5d8addd2 .

I'm open to any suggestion to refactor the regex we do use for email. I'm not sure we need such complicated regex.

backend/tracim_backend/app_models/email_validators.py
```diff
@@ -0,0 +1,41 @@
...
+        # literal form, ipv4 address (SMTP 4.1.3)
+        r"|^\[(25[0-5]|2[0-4]\d|[0-1]?\d?\d)" r"(\.(25[0-5]|2[0-4]\d|[0-1]?\d?\d)){3}\]\Z",
+        re.IGNORECASE | re.UNICODE,
+    )
```
raphj, line 23: We need to be consistent between the frontend and the backend here. Please make sure the frontend part will run the same regular expression.
(cc @gbonaspetti)

backend/tracim_backend/app_models/email_validators.py
```diff
@@ -0,0 +1,41 @@
+import re
+
+from marshmallow import ValidationError
+from marshmallow.validate import Email
+
+
+class TracimEmailValidator(Email):
+    """ Special Email Validator who accept email from subdomain"""
```
raphj, line 8: ...and many other things really

```suggestion
    """ Email Validator accepting a wider range of email addresses than marshmallow's Email field"""
```

raphj (approved):

backend/tracim_backend/app_models/email_validators.py
grignards: As I wrote in the issue, I think we should only check for `@`: email validation is **very** fragile.

grignards (changes requested):
See https://github.com/tracim/tracim/pull/4913/files#r708090017

frontend_lib/src/helper.js
```diff
@@ -449,13 +449,12 @@ export const parserStringToList = (string, separatorList = [',', ';', '\n']) =>
   return parsedString.split(',').filter(notEmptyString => notEmptyString !== '')
 }
 
-// INFO - GB - 2019-07-31 - This function check if the email has three parts arranged like somethig@something.somethig
+// INFO - GB - 2019-07-31 - This function check if the email has two parts arranged like somethig@something
 export const checkEmailValidity = email => {
   const parts = email.split('@')
```
inkhey, line 454:
```suggestion
  const parts = email.split('@')
```
To be consistent with the backend behavior, i suggest you to check if something like "value.rsplit("@", 1)" is possible.
The rfc say the @ to check is the last one as far i understood, and splitting will obtains weird result if you have more than 2 elements.

raphj: I confirm that the domain is after the last @ in an email.

frontend_lib/src/helper.js
```diff
@@ -449,13 +449,12 @@ export const parserStringToList = (string, separatorList = [',', ';', '\n']) =>
...
-  if (parts.length !== 2) return false
+  if (parts.length < 2) return false
 
-  const domainParts = parts[1].split('.')
-  return domainParts.length === 2
+  return parts[0] !== '' && parts[1] !== ''
```
raphj, line 457:
```suggestion
  return parts[0] && parts[parts.length - 1]
```

MariePrandstetter (approved):

frontend_lib/src/helper.js
```diff
@@ -449,13 +449,12 @@ export const parserStringToList = (string, separatorList = [',', ';', '\n']) =>
   return parsedString.split(',').filter(notEmptyString => notEmptyString !== '')
 }
 
-// INFO - GB - 2019-07-31 - This function check if the email has three parts arranged like somethig@something.somethig
+// INFO - GB - 2019-07-31 - This function check if the email has two parts arranged like somethig@something
```
raphj, line 452:
```suggestion
// INFO - GB - 2019-07-31 - This function checks if the string looks like an email (username@domain) with a non empty username and a non-empty domain
```

frontend_lib/src/helper.js
gbonaspetti: Updated

frontend_lib/src/helper.js
gbonaspetti: Updated

frontend_lib/src/helper.js
gbonaspetti: Outdated with new function

raphj:
review ok for the frontend part

frontend_lib/src/helper.js
```diff
@@ -449,13 +449,15 @@ export const parserStringToList = (string, separatorList = [',', ';', '\n']) =>
...
-  if (parts.length !== 2) return false
-
-  const domainParts = parts[1].split('.')
-  return domainParts.length === 2
+  const firstPart = email.substr(0, email.lastIndexOf('@'))
+  const secondPart = email.substr(email.lastIndexOf('@') + 1)
+  return firstPart !== '' && !firstPart.includes(',') && !firstPart.includes(';') &&
+    !firstPart.includes('>') && !firstPart.includes('<') &&
+    secondPart !== '' && !secondPart.includes(',') && !secondPart.includes(';') &&
+    !secondPart.includes('>') && !secondPart.includes('<')
```
raphj, line 460:
```suggestion
  return firstPart !== '' && secondPart !== '' && !/[,;<>]/.test(email)
```

backend/tracim_backend/app_models/email_validators.py
```diff
@@ -0,0 +1,57 @@
...
+        if not value or "@" not in value:
+            raise ValidationError(message)
+
+        user_part, domain_part = value.rsplit("@", 1)
+
+        if not self.USER_REGEX.match(user_part):
+            raise ValidationError(message)
+        if self.DOMAIN_REGEX.match(domain_part):
+            return value
+        raise ValidationError(message)
```
raphj, line 39:
```suggestion
    FORBIDDEN_CHARS = re.compile(r"[<>,;\n]")

    default_message = "Not a valid email address."

    def __init__(self, error=None):
        self.error = error or self.default_message

    def _format_error(self, value):
        return self.error.format(input=value)

    def __call__(self, value):
        message = self._format_error(value)

        if not value or "@" not in value or self.FORBIDDEN_CHARS.match(user_part):
            raise ValidationError(message)

        user_part, domain_part = value.rsplit("@", 1)

        if not user_part or not domain_part:
            raise ValidationError(message)
```
(not tested, solves duplicate regex, make it clear that we forbid empty parts)

frontend_lib/src/helper.js
raphj: The backend also forbids new line characters, so if you settle on forbidding it:
```suggestion
  return firstPart !== '' && secondPart !== '' && !/[,;<>\n]/.test(email)
```

backend/tracim_backend/app_models/email_validators.py
raphj: edited to fix some mistakes

inkhey: This does not work "a@>" is valid and I don't like the matching in inverted side, it's really confusing me.

inkhey: i made a better version of my code, removing duplicate regex, and reordering  the last part, but keeping regex for validation of full part.

raphj: Ah, right, you need to use search instead of match.

You probably find it confusing because you think in terms of allowed characters and not of forbidden characters. But you actually specify a list of forbidden characters in your regular expression and apply a double negation, which is actually arguably harder to follow for the reader.

raphj: I've edited my suggestion, it should work now.

raphj: If I had to caricature, my suggestions says "is this string containing one of those forbidden characters", while your code says "is each character not a non-allowed character", and makes the "non empty" requirement implicit.

inkhey: if we go this way, i found it useless to keep regex, checking with in seems more clear.

raphj: I prefer my version, but I'm fine with merging yours.

inkhey: I adapt you code without regex.

raphj: Sorry, your comments just showed up in both my email inbox and on the github page. I saw your new version and its better. My points still remain and I don't see a reason for having both DOMAIN_REGEX and USER_REGEX if there are doing the same thing.

But as I said, I'm okay with merging, it is still very easy to read.

raphj: These comments appearing late got to be a bit annoying. I'm okay with your last changes.

frontend_lib/src/helper.js
gbonaspetti: Fixed

MariePrandstetter (approved):

gbonaspetti:
One of the reviewers can check the check-boxes, please

raphj:
done

MariePrandstetter:
tested on branch

Issue #5047, opened 2021-10-29
Feat: Use Pyramid >=2.0 as python dependencies


inkhey:
## Feature description and goals

Update python dependencies for pyramid to support pyramid>=2.0
:warning: pyramid plugins may need to be updated too.

## Prerequisites

- https://github.com/tracim/tracim/issues/5046

## Implemented solution:

- remove pyramid-retry
- use compatibility layer (TracimSecurityPolicy) to keep current authentifications
- few fix in our custom authentifications to avoid infinite loop issue due to behavior change in pyramid 2.

PR ref: #5460

inkhey:
When updating pyramid to version 2.0, some check should be done about pyramid-retry, as the not found error may be broken related to it, check : https://github.com/tracim/tracim/issues/5046

grignards:
Could also be the time to rationalize the use of pyramid plugins (pyramid-retry usefulness is unclear) and authorization mechanism.

knodelle:
done

PR #5460, opened 2022-02-23
add support for pyramid 2

inkhey:
related to #5047 

## Checkpoints

**For developers**

- [x] Make sure that:

**For code reviewers**

**For testers**

### DIFF
--- a/backend/tracim_backend/__init__.py
+++ b/backend/tracim_backend/__init__.py
@@ -51,6 +51,7 @@ from tracim_backend.lib.utils.authentification import RemoteAuthentificationPoli
 from tracim_backend.lib.utils.authentification import TracimBasicAuthAuthenticationPolicy
 from tracim_backend.lib.utils.authorization import TRACIM_DEFAULT_PERM
 from tracim_backend.lib.utils.authorization import AcceptAllAuthorizationPolicy
+from tracim_backend.lib.utils.authorization import TracimSecurityPolicy
 from tracim_backend.lib.utils.cors import add_cors_support
 from tracim_backend.lib.utils.logger import logger
 from tracim_backend.lib.utils.request import TracimRequest
@@ -235,9 +236,12 @@ def web(global_config: OrderedDict, **local_settings) -> Router:
     # make sure to add this before other routes to intercept OPTIONS
     configurator.add_cors_preflight_handler()
     # Default authorization : Accept anything.
-    configurator.set_authorization_policy(AcceptAllAuthorizationPolicy())
-    authn_policy = MultiAuthenticationPolicy(policies)
-    configurator.set_authentication_policy(authn_policy)
+    configurator.set_security_policy(
+        TracimSecurityPolicy(
+            authn_policy=MultiAuthenticationPolicy(policies),
+            authz_policy=AcceptAllAuthorizationPolicy(),
+        )
+    )
     # INFO - GM - 11-04-2018 - set default perm
     # setting default perm is needed to force authentification
     # mechanism in all views.
--- a/backend/tracim_backend/__init__.py
+++ b/backend/tracim_backend/__init__.py
@@ -238,8 +238,8 @@ def web(global_config: OrderedDict, **local_settings) -> Router:
     # Default authorization : Accept anything.
     configurator.set_security_policy(
         TracimSecurityPolicy(
-            authn_policy=MultiAuthenticationPolicy(policies),
-            authz_policy=AcceptAllAuthorizationPolicy(),
+            authentification_policy=MultiAuthenticationPolicy(policies),
+            authorization_policy=AcceptAllAuthorizationPolicy(),
         )
     )
     # INFO - GM - 11-04-2018 - set default perm
--- a/backend/tracim_backend/lib/utils/authorization.py
+++ b/backend/tracim_backend/lib/utils/authorization.py
@@ -42,6 +42,31 @@ if TYPE_CHECKING:
 TRACIM_DEFAULT_PERM = "tracim"
 
 
+class TracimSecurityPolicy:
+    """
+    Adapter to keep existing authentification method compatible with pyramid 2.0
+    source:
+    https://docs.pylonsproject.org/projects/pyramid/en/latest/whatsnew-2.0.html#upgrading-from-third-party-policies
+    """
+
+    def __init__(self, authn_policy, authz_policy):
+        self.authn_policy = authn_policy
+        self.authz_policy = authz_policy
+
+    def authenticated_userid(self, request):
+        return self.authn_policy.authenticated_userid(request)
+
+    def permits(self, request, context, permission):
+        principals = self.authn_policy.effective_principals(request)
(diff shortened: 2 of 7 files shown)


backend/tracim_backend/lib/utils/authorization.py
```diff
@@ -42,6 +42,31 @@
 TRACIM_DEFAULT_PERM = "tracim"
 
 
+class TracimSecurityPolicy:
+    """
+    Adapter to keep existing authentification method compatible with pyramid 2.0
+    source:
```
inkhey, line 48:
```suggestion
    Adapter to keep existing authentification method compatible with pyramid 2.0
    source:
```
Note: we may want to support the new security policy mecanism as we still use deprecated mecanism this way, but i decided for now to just do the necessary to make things work.

backend/tracim_backend/models/setup_models.py
```diff
@@ -122,9 +122,6 @@ def init_models(configurator: Configurator, app_config: "CFG") -> None:
     # use pyramid_tm to hook the transaction lifecycle to the request
     configurator.include("pyramid_tm")
 
-    # use pyramid_retry to retry a request when transient exceptions occur
-    configurator.include("pyramid_retry")
```
inkhey, line 126: as pyramid_retry as some caveat and we have no clear reason to use it, i do think it's better to remove it. 
https://docs.pylonsproject.org/projects/pyramid-retry/en/latest/#caveats

The idea behind pyramid_retry is that is some specific case the transaction manager (with sqlalchemy) is able to know that an error can be fixed by retrying doing the request, so they may be some hypothetical case where this change something but it's something that in never encounter and i do think it's better to know these issue instead of hidding them behind some weird magic we don't completely understood.

backend/tracim_backend/__init__.py
```diff
@@ -235,9 +236,12 @@ def web(global_config: OrderedDict, **local_settings) -> Router:
     # make sure to add this before other routes to intercept OPTIONS
     configurator.add_cors_preflight_handler()
     # Default authorization : Accept anything.
-    configurator.set_authorization_policy(AcceptAllAuthorizationPolicy())
-    authn_policy = MultiAuthenticationPolicy(policies)
-    configurator.set_authentication_policy(authn_policy)
+    configurator.set_security_policy(
+        TracimSecurityPolicy(
+            authn_policy=MultiAuthenticationPolicy(policies),
```
inkhey, line 241: Note: the pyramid_multiauth is not up to date for pyramid v2, so we still use the old authentication policy logic (deprecated) there.

backend/tracim_backend/views/base.py
```diff
@@ -1,13 +1,25 @@
...
 from tracim_backend import PageNotFound
 from tracim_backend import TracimRequest
 from tracim_backend.views.controllers import Controller
+from tracim_backend.views.errors import ErrorSchema
 
 
 class BaseController(Controller):
     def not_found_view(self, context, request: TracimRequest):
-        raise PageNotFound("{} is not a valid path".format(request.path)) from context
+        # FIXME  - G.M - 2022-02-24 - Explicit error return as the global handle_exception does
```
raphj, line 14: Don't forget to create a GitHub issue for this FIXME.

raphj (approved):
I see nothing out of place at first sight but I don't know much about Pyramid.

backend/tracim_backend/lib/utils/authorization.py
```diff
@@ -42,6 +42,31 @@
...
+    https://docs.pylonsproject.org/projects/pyramid/en/latest/whatsnew-2.0.html#upgrading-from-third-party-policies
+    """
+
+    def __init__(self, authn_policy, authz_policy):
+        self.authn_policy = authn_policy
+        self.authz_policy = authz_policy
```
grignards, line 54: Maybe more explicit naming would help understand?
And should the policies be configurable? Or would it be enough to instantiate them here?

grignards (changes requested):

backend/tracim_backend/lib/utils/authorization.py
inkhey: Yes, it's possible to change the naming.
The policies are configurable and instantiated before running into this code, in fact, we instantiale multiple authentication policy into a single multiauth policy and a "allow all" authorization policy as we handle authorization with check_right decorator that doesn't use pyramid specific code.

backend/tracim_backend/views/base.py
Haskaris: Is this issue created?

backend/tracim_backend/lib/utils/authorization.py
Haskaris: How long do you think it would take to support it? I think it's a good thing to not keep deprecated mecanisms.

inkhey: It really depend how far we go. We do rely on a package that is not compatible with this for now, so if we do want to support this properly we add to replace it/update it for pyramid 2.0 which may take time.

Also, as we completely skip the authorization mecanism of pyramid, i'm not sure we should still use the authentification mecanism, but this may take some time also.

backend/tracim_backend/views/base.py
inkhey: yes : #5487

inkhey:
```suggestion
        # FIXME  - G.M - 2022-02-24 - Explicit error return as the global handle_exception does, 
        # see https://github.com/tracim/tracim/issues/5487
```

backend/tracim_backend/views/base.py
```diff
@@ -1,13 +1,26 @@
...
+from tracim_backend.views.errors import ErrorSchema
 
 
 class BaseController(Controller):
     def not_found_view(self, context, request: TracimRequest):
-        raise PageNotFound("{} is not a valid path".format(request.path)) from context
+        # FIXME  - G.M - 2022-02-24 - Explicit error return as the global handle_exception does, 
+        # see https://github.com/tracim/tracim/issues/5487
+        # not work for notfound_view in pyramid 2
+        error_body = ErrorSchema().build_from_exception(
```
inkhey, line 17:
```suggestion
        # FIXME  - G.M - 2022-02-24 - Explicit error return as the global handle_exception does 
        # not work for notfound_view in pyramid 2,
        # see https://github.com/tracim/tracim/issues/5487
        error_body = ErrorSchema().build_from_exception(
```

backend/tracim_backend/lib/utils/authorization.py
inkhey: ticket created about this https://github.com/tracim/tracim/issues/5488

backend/tracim_backend/lib/utils/authorization.py
inkhey: renaming done.

Haskaris (approved):

inkhey:
As soon as this is manually tested, we can merge this.

Issue #5526, opened 2022-03-16
Bug: Can't mark a group of notification as read


Haskaris:
## Description and expectations

#5497 bring a regression. With the refactoring of notifications, the possibility of marking a group as read by clicking on the little dot has disappeared.

### How to reproduce

1. Make a notification group
2. Click on the blue dot

## Linked PR
PR #5903

## Implemented solution
I've modified the `read every messages` endpoint to accept a list of messages to read. If there is no list then it will read every messages.
Also put `read one message` as deprecated since we are not using it anymore.

Haskaris:
This was started during integration phase of Tracim 4.3.2 build_91
Since it has some _major_ change I decided to create a dedicated PR (#5903).

Haskaris:
I've created a ticket for deprecated code #5941

Quiwy:
Tested on branch

Millefeuille42:
Done

PR #5903, opened 2022-09-29
fix(frontend, backend): Mark notification group as read

Haskaris:
Issue ref #5526

This PR also bring:
- Update of `messages/read` endpoint: now you can pass a list of notification to read, this was required since a notification group is a frontend entity
- Deprecate `messages/{event_id}/read` to use `messages/read` so we don't have to maintain the code twice, remove duplicated logic

## Checkpoints

**For developers**

- [ ] Make sure that:
- [ ] Original authors of the features included in a multi-feature branch (maintenance fixes -> develop, security fixes -> develop, …) should be part of the reviewers, especially if you encountered merge conflicts.

**For code reviewers**

**For testers**

### DIFF
--- a/backend/tracim_backend/views/core_api/schemas.py
+++ b/backend/tracim_backend/views/core_api/schemas.py
@@ -704,6 +704,15 @@ class UserIdPathSchema(marshmallow.Schema):
     )
 
 
+class EventIdListSchema(marshmallow.Schema):
+    event_id_list = marshmallow.fields.List(
+        marshmallow.fields.Int(
+            example=3, description="id of a valid event", validate=strictly_positive_int_validator,
+        ),
+        required=False,
+    )
+
+
 class EventIdPathSchema(marshmallow.Schema):
     event_id = marshmallow.fields.Int(
         example=5,
--- a/frontend/src/action-creator.async.js
+++ b/frontend/src/action-creator.async.js
@@ -946,6 +946,35 @@ export const getNotificationList = (
   return fetchGetNotificationWall
 }
 
+/**
+ * Put a list of notifications as read
+ * @param {String} userId
+ * @param {int[]} notificationIdList
+ * @returns
+ */
+export const putNotificationListAsRead = (userId, notificationIdList) => dispatch => {
+  return fetchWrapper({
+    url: `${FETCH_CONFIG.apiUrl}/users/${userId}/messages/read`,
+    param: {
+      credentials: 'include',
+      headers: FETCH_CONFIG.headers,
+      method: 'PUT',
+      body: JSON.stringify({
+        event_id_list: notificationIdList
+      })
+    },
+    actionName: NOTIFICATION_LIST,
+    dispatch
+  })
+}
+
+/**
+ * @deprecated
+ * Put a notification as read
+ * @param {String} userId user that read the notification
+ * @param {String} eventId id of the notification to read
+ * @returns
+ */
 export const putNotificationAsRead = (userId, eventId) => dispatch => {
   return fetchWrapper({
     url: `${FETCH_CONFIG.apiUrl}/users/${userId}/messages/${eventId}/read`,
--- a/frontend/src/action-creator.sync.js
+++ b/frontend/src/action-creator.sync.js
@@ -197,16 +197,18 @@ export const APPLIED_FILTER = (searchType) => `AppliedFilter_${searchType}`
 export const setAppliedFilter = (key, value, searchType) => ({ type: `${SET}/${APPLIED_FILTER(searchType)}`, key, value })
 export const resetAppliedFilter = (searchType) => ({ type: `${RESET}/${APPLIED_FILTER(searchType)}` })
 
+export const EVERY_NOTIFICATIONS = 'EveryNotifications'
 export const NEXT_PAGE = 'NextPage'
-export const NOTIFICATION_LIST = 'NotificationList'
 export const NOTIFICATION = 'Notification'
+export const NOTIFICATION_LIST = 'NotificationList'
 export const UNREAD_MENTION_COUNT = 'UnreadMentionCount'
 export const UNREAD_NOTIFICATION_COUNT = 'UnreadNotificationCount'
 export const appendNotificationList = (notificationList, spaceList) => ({ type: `${APPEND}/${NOTIFICATION_LIST}`, notificationList, spaceList })
 export const addNotification = (notification, spaceList) => ({ type: `${ADD}/${NOTIFICATION}`, notification, spaceList })
 export const updateNotification = (notificationId, notificationList) => ({ type: `${UPDATE}/${NOTIFICATION}`, notificationId, notificationList })
 export const readNotification = notificationId => ({ type: `${READ}/${NOTIFICATION}`, notificationId })
-export const readNotificationList = () => ({ type: `${READ}/${NOTIFICATION_LIST}` })
+export const readNotificationList = notificationIdList => ({ type: `${READ}/${NOTIFICATION_LIST}`, notificationIdList })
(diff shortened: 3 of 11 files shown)


frontend/src/action-creator.sync.js
```diff
@@ -197,16 +197,17 @@ export const APPLIED_FILTER = (searchType) => `AppliedFilter_${searchType}`
 export const setAppliedFilter = (key, value, searchType) => ({ type: `${SET}/${APPLIED_FILTER(searchType)}`, key, value })
 export const resetAppliedFilter = (searchType) => ({ type: `${RESET}/${APPLIED_FILTER(searchType)}` })
 
+export const EVERY_NOTIFICATIONS = 'EveryNotifications'
```
Skylsmoi, line 200: I'd rather use the NOTIFICATION key and add "EVERY"

same for NOTIFICATION_LIST that should use NOTIFICATION

frontend/src/component/GroupedNotificationItem/GroupRender.jsx
```diff
@@ -29,24 +36,19 @@ const GroupRender = props => {
...
-    //   const fetchPutNotificationAsRead = { status = 404 }
-    //   switch (fetchPutNotificationAsRead.status) {
-    //     case 204: {
-    //       props.dispatch(readGroupNotification(notificationId))
-    //       break
-    //     }
-    //     default:
-    //       props.dispatch(newFlashMessage(props.t('Error while marking the notification as read'), 'warning'))
-    //   }
+  const handleReadNotificationGroup = async (groupNotification) => {
```
Skylsmoi, line 39: async function's body should always be wraped in a try catch

frontend/src/action-creator.sync.js
```diff
@@ -197,17 +197,17 @@ export const APPLIED_FILTER = (searchType) => `AppliedFilter_${searchType}`
 export const setAppliedFilter = (key, value, searchType) => ({ type: `${SET}/${APPLIED_FILTER(searchType)}`, key, value })
 export const resetAppliedFilter = (searchType) => ({ type: `${RESET}/${APPLIED_FILTER(searchType)}` })
 
-export const EVERY_NOTIFICATIONS = 'EveryNotifications'
-export const NEXT_PAGE = 'NextPage'
 export const NOTIFICATION = 'Notification'
-export const NOTIFICATION_LIST = 'NotificationList'
+export const NEXT_PAGE = 'NextPage'
+export const NOTIFICATION_LIST = `${NOTIFICATION}List`
+export const EVERY_NOTIFICATION_LIST = `Every${NOTIFICATION_LIST}`
```
Skylsmoi, line 203: you must separate keywords with a slash

Skylsmoi (changes requested):

frontend/src/reducer/notificationPage.js
```diff
@@ -146,7 +146,7 @@ export default function notificationPage (state = defaultNotificationsObject, ac
       }
     }
 
-    case `${READ}/${NOTIFICATION_LIST}`: {
+    case `${READ}/${EVERY_NOTIFICATION}`: {
```
gbonaspetti, line 149:
```suggestion
    case `${READ}/${ALL_NOTIFICATIONS}`: {
```
Just a suggestion, you don't have to apply if you prefer the current name: since in most of the code we use ALL and not EVERY. But if you apply, it would be good to change on `readEveryNotifications` too. If you decide not to apply it, it would still be good to normalize the plural, because here we have `EVERY_NOTIFICATION` without s and in the `readEveryNotifications` function with s.

frontend/src/action-creator.async.js
```diff
@@ -947,15 +947,41 @@ export const getNotificationList = (
...
+/**
+ * Put a list of notifications as read
+ * @param {String} userId
+ * @param {int[]} notificationIdList
+ * @returns
+ */
+export const putNotificationListAsRead = (userId, notificationIdList) => dispatch => {
+  return fetchWrapper({
+    url: `${FETCH_CONFIG.apiUrl}/users/${userId}/messages/read` +
+      `?notification_ids=${notificationIdList.join(',')}`,
```
gbonaspetti, line 959:
```suggestion
    url: `${FETCH_CONFIG.apiUrl}/users/${userId}/messages/read?notification_ids=${notificationIdList.join(',')}`,
```

gbonaspetti: Just a suggestion, don't need to apply

gbonaspetti (changes requested):

frontend/src/action-creator.async.js
Haskaris: I've splited the line here to keep the line under 100 characters while easily readable

frontend/src/reducer/notificationPage.js
Haskaris: I feel like ``every`` is more adapted here since we are reading every one of the notifications.
For the ``s`` it's after a debate with Skylsmoi. His first review was ``EVERY_NOTIFICATIONS`` to ``EVERY_NOTIFICATION_LIST``. However this key (``EVERY_NOTIFICATIONS`) is meant to say _We are reading every notifications_. Since the ``s`` is confusing (see Skylsmoi) we've decided to remove the plural.

lebouquetin: I think the naming should be decided in ordre to improve thé understandability. If you process a liste, a plural form is a better naming choice, I think.

A naming convention should fixe this (what is the rule for plural form var naming)

Haskaris: I'm not going to lie, I'm very confused about this..
So I've done some researches and here is what I've got:
- ``all`` refers to the whole group
- ``Every`` focuses on each individual member of the whole group

With all -> plural
With every -> singular

_So, in every case ``every_notifications`` is wrong grammatically speaking_

I can change to ``all_notification_list`` (with ``list`` and not just plural because the code is using this convention in the frontend) but as I said, the key to redux here is to read _every_ notification. And put it as read _all_ notifications confuse me as possible is human possible since we are speaking of each notifications that aren't read already

lebouquetin: What I say is *convention* is better than *grammar* for code maintenance.

If the rule applied is different, when maintaining code I will ask myself : "why is the naming différent? Is there a specific behavior?"

I juste reacted to the discussion about this is more correct vs the code is full or "all".

Conventions may change. I have no opinion here.

lebouquetin: I mean:

- same behavior -> same naming rule
- different behavior -> different naming rule

frontend/src/action-creator.async.js
gbonaspetti: ok

backend/tracim_backend/views/core_api/schemas.py
```diff
@@ -2226,13 +2226,26 @@ class PathSuffixSchema(marshmallow.Schema):
 class UserMessagesMarkAsReadQuerySchema(marshmallow.Schema):
     content_ids = StrippedString(
         validate=regex_string_as_list_of_int,
-        example="3,4",
-        description="comma separated list of content_ids to check for marking event as read",
+        example="1,4",
+        description="Comma separated list of content ids. Every event related to these contents\
+            will be marked as read.",
+    )
+    notification_ids = StrippedString(
```
grignards, line 2233:
```suggestion
   event_ids = StrippedString(
```
Notifications do not exist for the backend.

grignards (changes requested):

gbonaspetti (approved):

Skylsmoi (approved):

Quiwy:
Tested on branch

Issue #5818, opened 2022-07-29
Feat: Borderless design


gbonaspetti:
## Feature description and goals

As a user, I consider a borderless design to be modern and clean.

## Prerequisites

None

## PR

https://github.com/tracim/tracim/pull/5963

Some colors had to be changed to ensure contrast and to help with this change, an update was made to the nomenclature, making it easier to understand.

Millefeuille42:
Tested on branch

Millefeuille42:
Done

PR #5963, opened 2022-10-22
Feat/5818 remove borders and update colors

gbonaspetti:
#5818

## Checkpoints

**For developers**

  - This PR just change some colors and borders, no tests needed for that.
- [x] Make sure that:

**For code reviewers**

**For testers**

### DIFF
--- a/frontend/src/component/Activity/ActivityList.styl
+++ b/frontend/src/component/Activity/ActivityList.styl
@@ -20,7 +20,7 @@ previewHeight = 400px
 
     &__overflowOverlay
       background-image:
-        linear-gradient(to top, rgba(0, 0, 0, 0.25), transparent),
+        linear-gradient(to top, rgba(0, 0, 0, 0.1), transparent),
         radial-gradient(ellipse at bottom, rgba(0, 0, 0, 0.1), transparent)
       background-repeat no-repeat
       background-position bottom
--- a/backend/color-test.json
+++ b/backend/color-test.json
@@ -1,7 +1,7 @@
 {
   "primary": "#7d4e24",
-  "sidebar": "#f6f6f6",
-  "sidebar/logo": "#f6f6f6",
+  "sidebar": "#f3f3f3",
+  "sidebar/logo": "#f3f3f3",
   "sidebar/font": "#252525",
   "contents/html-document": "#3f52e3",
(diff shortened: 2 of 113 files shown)


Haskaris:
I just went for a little try before any review/test
Did we have considered removing some colors?

For example:
image
To
image

The white block made sense when we had borders since buttons were white too. But since they aren't white anymore, maybe we can remove the "block" background to reduce the interface weight

gbonaspetti:
@Haskaris Thanks for all the feedbacks, they will be taken into account for the improvements of this PR, the idea is to make a first version that even not being perfect, can be merged today so we can test it during the week and make the improvements pointed out.

frontend/src/component/Activity/ActivityList.styl
```diff
@@ -20,7 +20,7 @@ previewHeight = 400px
 
     &__overflowOverlay
       background-image:
-        linear-gradient(to top, rgba(0, 0, 0, 0.25), transparent),
+        linear-gradient(to top, rgba(0, 0, 0, 0.1), transparent),
```
Haskaris, line 23: `linear-gradient(to top, rgba(0, 0, 0, 0.1), transparent)` should be `linear-gradient(to top, rgba(1, 1, 1, 1), transparent)`
So the gradient will fade to the white

gbonaspetti: `rgba(1, 1, 1, 1)` it's black, to fade to white it's `rgba(255, 255, 255, 1)`. I can update this, the initial change was just "make it a little closer to the gray of the card".

Millefeuille42:
About the emojis reactions, to make the reactions User did to stand-out I'd suggest using a distinct color, and not playing with shades. For example, a shade of main color or success color could be used, using the alpha layer to give a soft but distinguishable color

gbonaspetti:
Yes, we use the primary color. After it was a lighten color of the primary, but after the comment I changed to use an opacity of it

Millefeuille42:
My bad I did not see the changes before commenting, this is indeed way better. Unfortunately this is still improvable, but the current color palette is not permitting better so I think it's OK

Millefeuille42:
Tested on branch

Millefeuille42 (approved):

PR #5968, opened 2022-10-24
fix(frontend): affecting value to props inside of component in FeedItemWithPreview

Millefeuille42:
[#5969](https://github.com/tracim/tracim/issues/5969)
## Checkpoints

**For developers**

- [x] Make sure that:

**For code reviewers**

**For testers**

### DIFF
--- a/frontend/src/container/FeedItemWithPreview.jsx
+++ b/frontend/src/container/FeedItemWithPreview.jsx
@@ -324,8 +324,10 @@ export class FeedItemWithPreview extends React.Component {
         ? this.getFirstComment()
         : null
     )
+
+    const previewContent = props.content
     if (shouldShowComment && props.inRecentActivities) {
-      props.content.firstComment = commentToShow
+      previewContent.firstComment = commentToShow
     }
 
     const commentList = this.getTimelineData()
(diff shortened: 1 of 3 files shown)


frontend/src/container/FeedItemWithPreview.jsx
```diff
@@ -324,8 +324,10 @@ export class FeedItemWithPreview extends React.Component {
         ? this.getFirstComment()
         : null
     )
+
+    const previewContent = props.content
     if (shouldShowComment && props.inRecentActivities) {
-      props.content.firstComment = commentToShow
+      previewContent.firstComment = commentToShow
```
Skylsmoi, line 330: js variables are always assigned by reference. So this code doesn't fix the issue, you are still modifying props.content.firstComment

I'm not sure what you are trying to do but I expect something like this to solve your issue:
```jsx
const commentToDisplay = shouldShowComment && props.inRecentActivities
  ? commentToShow
  : props.content.firstComment
```

Skylsmoi (changes requested):

frontend/src/container/FeedItemWithPreview.jsx
Skylsmoi: but having variables "commentToDisplay" and "commentToShow" feels like it can be improved

Millefeuille42: Due to the structure of `Comment`, I need to pass a copy of props.content to it as the `apiContent` prop. Changing how `Comment` works could be a solution but it could involve a handful of additional changes. I'll just edit the code to init `previewContent` as a deep copy of `props.content`

Millefeuille42: Actually comment is only used twice, changing how comment works is way more performant than creating a deep copy

Issue #5802, opened 2022-07-25
Bug: No more padding in the favorites list


Relf-Algoo:
## Description and expectations

When there are too many favourites in the list, the padding disappears. See topic https://algoo.tracim.fr/ui/workspaces/49/contents/thread/73848 for more informations

### How to reproduce

1. Go to…
2. Click on…
3. Scroll down to…

### Version information

- Tracim version: 4.3.0, our tracim
- Web browser: Firefox 
- Version: 102
- Device / System: work laptop

### Screenshots

________________________________________________

## PR
#5984

Millefeuille42:
Done

Issue #5983, opened 2022-10-26
🐛 Bug: Favorites are hard to read due to new font size


Millefeuille42:
## Description and expectations

Due to the new font size, favorites are dense and packed making them hard to read

## PR
#5984

gbonaspetti:
Done

PR #5984, opened 2022-10-26
fix(frontend): Favorites are hard to read due to new font size

Millefeuille42:
#5983
#5802

## Result

image

## Checkpoints

**For developers**

- [x] Make sure that:

**For code reviewers**

**For testers**

### DIFF
--- a/frontend/src/component/ContentListItem.styl
+++ b/frontend/src/component/ContentListItem.styl
@@ -10,7 +10,7 @@
   min-height 65px
   align-items center
   & > *
-    margin 0 8px
+    margin 0 standardSpacing
   &:hover
     color fontColor
   &__link
--- a/frontend/src/component/ContentListItem.styl
+++ b/frontend/src/component/ContentListItem.styl
@@ -7,9 +7,8 @@
   display flex
   justify-content space-between
   color fontColor
-  height 60px
+  height 65px
   align-items center
-  min-height 65px
   & > *
     margin 0 8px
   &:hover
--- a/frontend/src/component/ContentListItem.styl
+++ b/frontend/src/component/ContentListItem.styl
@@ -9,6 +9,9 @@
   color fontColor
   height 60px
   align-items center
+  min-height 65px
+  & > *
+    margin 0 8px
   &:hover
     color fontColor
   &__link
--- a/frontend/src/component/ContentListItem.styl
+++ b/frontend/src/component/ContentListItem.styl
@@ -7,7 +7,7 @@
   display flex
   justify-content space-between
   color fontColor
-  height 65px
+  min-height 65px
   align-items center
   & > *
     margin 0 8px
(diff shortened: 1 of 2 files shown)


frontend/src/component/ContentListItem.styl
```diff
@@ -9,6 +9,9 @@
   color fontColor
   height 60px
```
grignards, line 10: Shouldn't the `min-height` be 60px? Then `height` can probably be removed.

Millefeuille42: Indeed height could be set to 65px, removing the min-height property

frontend/src/component/ContentListItem.styl
```diff
@@ -7,8 +7,10 @@
   display flex
   justify-content space-between
   color fontColor
-  height 60px
+  height 65px
   align-items center
+  & > *
+    margin 0 8px
```
gbonaspetti, line 13: Maybe I do not understand the fix/issue, but it seems that it just add the left-right margins to all lists, like for notification wall:

Capture d’écran de 2022-10-27 09-31-41

Shoudn't the fix be about #5802 (that could be fix with a min-height) ?

gbonaspetti (changes requested):

frontend/src/component/ContentListItem.styl
Millefeuille42: The margins purpose is to keep a blank space between each field because one of the problems with the readability were also caused by horizontal spacing, no just the vertical one. I should indeed used min-height (I'm fixing this right now) but the whole goal was to bring more readability in the favorites, not just linked to the vertical spacing

### How it looks with the fix
image

### How it looked before
image

As you can see, the different fields were too close to each other, resulting in poor readability. This fix brings a less dense content display for the favorites.

I also checked for the other usages of the content list and fortunately it give also a lighter interface.

Millefeuille42: However, this fix could also fix #5802 as it solves this problem too.

gbonaspetti: You are right, I applied the fix in the wrong place. I'm still not 100% convinced because it generate different spacing inside and outside the elements (orange in the screenshot), but I think it was already the case, so not a strong opinion.
Capture d’écran de 2022-10-27 10-10-51

I suggest use the standardSpacing

```suggestion
    margin 0 standardSpacing
```

frontend/src/css/Favorites.styl
```diff
@@ -4,10 +4,11 @@
 .favoritesHeader
   display flex
   width 100%
-  height 60px
+  min-height 65px
   justify-content space-between
   align-items center
-
+  & > *
+    margin 0 8px
```
gbonaspetti, line 11:
```suggestion
    margin 0 standardSpacing
```

Issue #6048, opened 2022-11-25
🐛 Bug: Mention autocompletion not taking public name into account


Millefeuille42:
4.5 : #6049 
maintenance: #6054 

## Description and expectations

When trying to mention someone, typing their public name does not bring up associated user. Since this is a regression, it is required to include e2e tests in the patch.

### How to reproduce

Context: Trying to mention someone named `John Smith`, username: `j__s__`

1. In any text field type @ to start the mention autocomplete
2. Type john to search for aforementioned user
3. Their profile is not showing up, you'd have to type `j_` for it to show up. 

### Version information

Seen in build_095, might have appeared sooner.

PR #6049, opened 2022-11-25
fix(frontend): mention autocompletion not taking public name into account

Millefeuille42:
#6048 

## Checkpoints

**For developers**

- [x] Make sure that:

**For code reviewers**

**For testers**


gbonaspetti (approved):

frontend_lib/src/appContentFactory.js
```diff
@@ -1052,12 +1053,13 @@ export function appContentFactory (WrappedComponent) {
           await getSpaceMemberList(this.apiUrl, workspaceId)
         )
 
+        const includesKeyword = stringIncludes(keyword)
```
Skylsmoi, line 1056: I've just remembered that on frontend/ we have lodash library available
And, in lodash, there is _.includes that does the same thing as stringIncludes ([lodash _.includes doc](https://lodash.com/docs/4.17.15#includes))

It's generally better to use helper functions from vendors libraries that already are in the project because:
- it avoid adding new code
- the function have better documentation and examples
- the function is more tested
- other developers tends to be more familiars with it

I've checked the code base and it seems only frontend_app_workspace_advanced/ is using stringIncludes without having lodash in its package.json.
And that is making me think about adding lodash to frontend_vendors

And last comment about it, stringIncludes isn't unit tested

Conclusion:
- Quick solution: Add unit test to stringIncludes
- Middle solution: Refactor stringIncludes usage with _.includes and add lodash to frontend_app_workspace_advanced/
- Long solution: Refactor stringIncludes usage with _.includes, add lodash to frontend_vendors and remove lodash from every other package.json
- Alternative solution: change the body of stringIncludes to use _.includes. It would avoid having to change lodash imports and stringIncludes usages but will still require unit tests

Whether the quick, middle or the long solution is chosen, I'd like one of them to be implemented
I'm open to discuss which one to choose

Side note: that doesn't take away the fact that your implementation of stringIncludes is nice

Skylsmoi: as seen together, we will implement the quick solution because judging by the usage of your function, it makes sense.
Also since this might might during the content list refactor, let's not spend a lot of time on it for now.

grignards (approved):

gbonaspetti:
After testing, there is a bug on the autocomplete position. We can see that it's not the same between demo (4.4.2-build_095) and this branch. 

https://user-images.githubusercontent.com/35422692/204546391-07bb1c98-d311-458a-b7ea-41416b7dd20b.mp4

Millefeuille42:
@gbonaspetti This is not brought by this development, but by build_096. Should it be treated in this PR as well or addressed in a separate issue ?

gbonaspetti:
> @gbonaspetti This is not brought by this development, but by build_096. Should it be treated in this PR as well or addressed in a separate issue ?

No problem, we can do it on another issue.

gbonaspetti:
Tested on branch

Issue #6007, opened 2022-11-03
✨ Feat: Better mention support


Haskaris:
## Feature description and goals

With the increase amount of technical debt, it's time for a good old refactoring of mentions in Tracim.

#### General
Maybe we should use Tiny to the simple AND advanced text editor, with the only difference that we hide the header bar, this way we have a _simple_ text editor with every _advanced_ feature.

#### TinyMCE
We can  _easily_ create a plugin for auto-completion (mention). This way we can scale the auto-completion for mention and/or content link. See [autocompletion.ts](https://github.com/tinymce/tinymce/blob/develop/modules/tinymce/src/plugins/emoticons/main/ts/ui/Autocompletion.ts) from TinyMCE pulgins.

#### TextArea
There is quite few plugins to mentions in js available.

#### General
A mention should be a special tag with an id of a specific user.
If we think about the future, the idea of having multiple role/other mentions like `here` announce that the tag should be able to distinct and carry theses information. If we are tagging a person, a role. However, a role should be liked to a user, therefore this is not linked to this issue _but we have to keep it in mind_.

## Related PRs

PR #6058 (merged)
PR #6109 (merged)
PR #6119 (merged)

Haskaris:
Update time:
- I've made a POC (quick development) to use TinyMCE for simple and advanced edition, this way I can focus the development about mention on TinyMCE only
- I've _replaced_ the current TinyMCE to use the [React TinyMCE](https://github.com/tinymce/tinymce-react)
- I've explored/made the auto completion about mention
  - **:warning: I currently fetch every mention after each key typed once the auto completion is opened**
- I've created a HTML _custom-element_ which allows me to put a custom HTML element that will be replaced by a React component that will render the user name
  - For example: `<html-mention>{userId}</html-mention>`
- We (Skylsmoi and I) thought about the extensibility of it:
  - We will be able to add custom roles. Theses roles will have a custom ID but won't be able to have a translation. For example:
    - Dev
    - Commercial
  - We will be able to add Tracim roles easily. Theses roles will have a hard coded ID and a translation. For example:
    - Contributor
    - Administrator

Haskaris:
# Update time

## TinyMCE
- I've tried to create a plugin for TinyMCE. But in order to access member and for easier I removed the plugin code to put it in the `setup` section of the TinyMCE `Editor` _reminder: Editor is a [react component officially supported by TinyMCE](https://github.com/tinymce/tinymce-react)_

## Custom Element
[...]

Issue #5831, opened 2022-08-09
Feat: scroll in the mention list


Relf-Algoo:
## Feature description and goals

User wants to be able to scroll the user list (@) or content list (#) when the list is too big or the exact name is not known. This would be especially useful in big instances with a lot of content, where names tend to be reused or very similar. 

## Related PR
PR #6058

Issue #5442, opened 2022-02-17
Misc: Make TinyMCE's height adapt to its container's height 


raphj:
## Feature description and goals

Since migrating to TinyMCE 5, the `height: 100%` style that is applied to it does not work (anymore?).
To avoid custom styles and workarounds and arbitrary sizes wherever TinyMCE is embedded, find a way to make it adjust to its container automatically.

## Related PR
PR #6058

raphj:
We shall decide rules to size TinyMCE appropriately, including on small and mobile screens

Quiwy:
I guess I'm facing the same issue here.

I want to edit a "large" discussion (more than 25 words) and it's quite annoying to have to scroll a lot, even more if you have images on the discussion :

![image](https://user-images.githubusercontent.com/20048264/166455249-b13ea6bf-45ba-4e01-a697-b7d8812b49fa.png)

raphj:
@Quiwy This issue is a technical debt (only an issue for developers). Your issue is more that you find the TinyMCE component too small. You should file a new issue for this, something like "The tinyMCE component is too small when editing a comment".

As a workaround, I suggest you to use the full screen feature of TinyMCE (which corresponds to the last icon in your screenshot. That should allow you to comfortably write a long comment with big images.

See also: https://github.com/tracim/tracim/issues/4583

Issue #5436, opened 2022-02-16
Misc: TinyMCE-related end to end Cypress tests are unstable and therefore disabled since the migration to TinyMCE 5


raphj:
## Description and expectations

The related tests shall be fixed.

### Version information

- Tracim version: upcoming 4.2

PR #6058, opened 2022-11-29
feat(frontend, backend): better mention support

Haskaris:
# **:warning: When merging this PR, be careful to update every mention in scripts**

Issue related:
- #4136
- #6007 

## Also fix:

[TESTS]
- #3986
  - Fixed with TinyMCE React -> we use one component (to verify tho)

[FEAT]
- #2817
- #4396
- #4727
- #4795
- #5394
- #5442
- #5831

[BUG]
- #3519
- #3520
- #3590
- #3640
- #3671
- #4437
- #4463
- #4658
- #4714
- #4972
- #5080
- #5132
- #5407
- #5429

[DOCUMENTATION]
- #3892
  - Now mention are `<html-mention roleid="0" />` or `<html-mention userid="0" />`

Partially fix:
- #4436 (simple edition)
- #4358 (remove the duplication in `frontend`)
- #5436 (I didn't check every tests)

## Checkpoints

**For developers**

- [x] Make sure that:

**For code reviewers**

**For testers**

## Known limitation

~~- There is no picture of the members in the auto-completion~~
  ~~- We can only put image (with a source) or a text. Therefore composed HTML (like we currently have for user without avatar) doesn't work.~~
- Mentions and links in Kanban are broken && workspace-advanced
  - Still need to replace the old mechanism in Kanban and in Workspace-advanced
- Username live update (when someone change their username) doesn't work in `RecentActivities` (not thoses in the space activities)


frontend_lib/src/helper.js
```diff
@@ -927,6 +928,213 @@ export const handleClickCopyLink = (contentId) => {
...
+ * @returns {List[String]} List of content id found
+ * Example:
+ * - Input: `<p>Test #844</p>`
+ * - Output: `['#844']`
+ */
+export const searchContent = (text) => {
+  // Regex explanation: https://regex101.com/r/z1WUUu/1
+  // Match (#XXX part): '#XXX', '#XXX ', ' #XXX', '#XXX:', ':#XXX', '(#XXX)', '#XXX!', ...
+  // Don't match: 'XXX#XXX', '#<span>XXX</span>'
+  const contentRegex = /(?<=^|\s|\W)#([0-9]+)\b/g
```
Haskaris, line 1094: This regex is currently wrong. It will match `"#XXX"` and cause the text sent after an edit where a content were already typed to be broken?

1. Create a comment with a content link
2. Edit the previous comment so it can be send again
3. See the broken link/text

Haskaris: Changed to use `/(?<=^|\s|\W)(?<!title=")#([0-9]+)\b/g`
See https://regex101.com/r/z1WUUu/3

frontend/src/component/Mention/Mention.jsx
```diff
@@ -0,0 +1,106 @@
...
+    </Provider>
+  )
+}
+
+export const Mention = props => {
+  const DEFAULT_MENTION = props.t('UnknownUser')
+
+  const [mention, setMention] = useState({
+    id: undefined, // User id of the mention
+    level: undefined, // 0: all, 1: reader, 2: contributor, 4: content manager, 8: workspace manager
```
Skylsmoi, line 25: this naming invalidates the "roleId" name elsewhere

frontend/src/component/Mention/Mention.jsx
```diff
@@ -0,0 +1,106 @@
...
+    text: DEFAULT_MENTION // Text to display in the mention
+  })
+
+  const completeRoleList = ROLE_LIST.concat([{
+    id: 0,
+    label: props.t('All'),
+    slug: props.t('all')
+  }])
+
+  if (props.userid) {
```
Skylsmoi, line 36: move this handling into a function so you can use switch and/or early returns

frontend/src/component/Mention/Mention.jsx
```diff
@@ -0,0 +1,106 @@
...
+        }
+
+        if (mention.text === DEFAULT_MENTION) {
+          // Fetch from API
+          props.dispatch(getUser(mention.id)).then((response) => {
+            if (response.status === 200) {
+              const user = response.json
+              setMention({ ...mention, text: user.username })
+            }
+          })
```
Skylsmoi, line 62: I'd like an async function

frontend/src/component/Mention/Mention.jsx
```diff
@@ -0,0 +1,106 @@
...
+    const role = completeRoleList.find(r => Number(r.id) === mention.level)
+
+    if (role) {
+      mention.text = role.slug
+    }
+
+    if (mention.level === 0) {
+      mention.isToMe = true
+    } else {
+      // FIXME - Doesn't work in recent activities
```
Skylsmoi, line 81: can remove this comment?

frontend/src/component/Mention/Mention.jsx
```diff
@@ -0,0 +1,106 @@
...
+      const myRole = props.currentWorkspace.memberList.find(m => m.id === props.user.userId).role
+      const myRoleLevel = completeRoleList.find(r => r.slug === myRole).id
+      mention.isToMe = (myRoleLevel >= mention.level)
+    }
+  }
+
+  const data = classnames(
+    'mention',
+    'mention-v2', // NOTE - MP - 2023-01-02 - Not used yet
```
Skylsmoi, line 90: can be removed?

frontend/src/container/Publications.jsx
```diff
@@ -542,6 +459,8 @@ export class Publications extends React.Component {
     const currentPublicationId = Number(props.match.params.idcts || 0)
     const isPublicationListEmpty = props.publicationPage.list.length === 0
 
+    console.log('props.currentWorkspace.memberList', props.currentWorkspace.memberList)
```
Skylsmoi, line 462: can remove?

frontend_lib/src/appContentFactory.js
```diff
@@ -586,45 +564,23 @@ export function appContentFactory (WrappedComponent) {
...
           content.content_namespace
         )
       )
 
       switch (response.apiResponse.status) {
         case 200:
-          setState({ newComment: '', showInvalidMentionPopupInComment: false })
-          if (isCommentWysiwyg) tinymce.get(`wysiwygTimelineComment${id}`).setContent('')
-
           removeLocalStorageItem(
```
Skylsmoi, line 584: can this be removed?
it should be already removed by comment area

frontend_lib/src/appContentFactory.js
```diff
@@ -678,24 +636,17 @@ export function appContentFactory (WrappedComponent) {
...
-          newCommentAsFileList.map(newCommentAsFile => this.saveCommentAsFile(content, newCommentAsFile))
-        )
-        const uploadFailedList = responseList.filter(oneUpload => isFileUploadInErrorState(oneUpload))
-        uploadFailedList.forEach(fileInError => sendGlobalFlashMessage(fileInError.errorMessage))
+      setState({ isFileCommentLoading: true })
+      const responseList = await Promise.all(
+        fileList.map(file => this.appContentSaveNewCommentFile(content, file))
+      )
+      const uploadFailedList = responseList.filter(upload => isFileUploadInErrorState(upload))
+      uploadFailedList.forEach(fileInError => sendGlobalFlashMessage(fileInError.errorMessage))
```
Skylsmoi, line 647: I think globalFlashMessage cannot appear more than once. So it should be better to merge all message into one globalFlashMessage

Skylsmoi (changes requested):
Review on a first batch of file

frontend_lib/src/helper.js
```diff
@@ -927,6 +926,213 @@ export const handleClickCopyLink = (contentId) => {
...
+ * Replace the given HTML string containing the mention with the role slug
+ * @param {Array[object]} roleList List of roles
+ * @param {String} html Current html text with balise mention
+ * @returns Html text without mention balise
+ * Example:
+ * - Input: `<p>Test <html-mention roleid="0"><\html-mention></p>`
+ * - Output: `<p>Test @all</p>`
+ */
+const replaceHTMLElementWithMentionRole = (roleList, html) => {
+  const mentionRegex = /<html-mention roleid="(\d+)"><\/html-mention>/g
```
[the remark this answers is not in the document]
Haskaris: I think ``rolelevel`` is more appropriate here, since it's compared to level (in backend to send the notification and should be compared to the level in the frontend to know if the mention is for us)

However, in the frontend everywhere we reference to the ``rolelevel`` we use the ``id`` of the ``role``... That's why I used ``roleid``

frontend/src/component/Mention/Mention.jsx
Haskaris: Hum... Since this function is used for render, I can't use ``await`` in a render...

Skylsmoi: as seen together, your solution doesn't work on every cases

frontend_app_html-document/src/component/HtmlDocument.jsx
```diff
@@ -28,7 +31,18 @@ export const HtmlDocument = props => {
 
   useEffect(() => {
     setText(props.text)
-  }, [])
+  }, [props.text])
```
Skylsmoi, line 34: could you rename the state "text" to be more clear that this state is the text that user will update and props.text is the currently saved version?

I'm saying this because, in the render, you sometimes pass props.text and sometimes state text and it confused me a bit

frontend_lib/src/mentionOrLink.js
```diff
@@ -201,3 +201,224 @@ const wrapLinksFromText = async (text, doc, apiUrl) => {
...
   return fragment
 }
+
+// /////////////////////////////////////////////////////////////////////////////
+// NOTE - MP - 2022-12-02 - MENTION SECTION
+// /////////////////////////////////////////////////////////////////////////////
+
+// NOTE - MP - 2023-01-11 - This should be fusionned with ROLE_LIST. However, since we only support
+// `all` role, it requires some additional processing.
+export const DEFAULT_ROLES = [
```
Skylsmoi, line 211: please rename DEFAULT_ROLE_LIST

frontend_lib/src/mentionOrLink.js
```diff
@@ -201,3 +201,224 @@ const wrapLinksFromText = async (text, doc, apiUrl) => {
...
+      const linkBalise = `<a class="internal_link primaryColorFont" href="${
+        PAGE.CONTENT(contentId)
+      }" title="${content}">${contentTitle}</a>`
+      newHtml = newHtml.replace(content, linkBalise)
+    } else {
+      invalidContentList.push(content)
+    }
+  }))
+
+  console.log('NEW HTML: ', newHtml)
```
Skylsmoi, line 421: missing trace

Skylsmoi (changes requested):
There are a lot a comments that don't respect the comment rules, it would be nice to update

Skylsmoi (approved):
I'm only approving the frontend part

Skylsmoi (approved):
only the frontend part

grignards (changes requested):

backend/tracim_backend/lib/core/mention.py
```diff
@@ -23,30 +22,43 @@
...
 
     def __repr__(self) -> str:
-        return "<Mention(recipient={}, id={})>".format(self.recipient, self.id)
+        return (
+            f"<Mention(type={self.type}, recipient={self.recipient}, content_id={self.content_id})>"
+        )
 
     def __hash__(self) -> int:
-        return hash(self.id)
+        return hash(self.content_id)
```
Millefeuille42, line 56: The hash will not be unique when mentioning multiple users in the same content, set it to a mix of type, recipient and content_id.

backend/tracim_backend/lib/core/mention.py
```diff
@@ -63,32 +75,39 @@ class DescriptionMentionParser(BaseMentionParser):
...
+        return self.get_mentions_from_html(revision.content_id, revision.raw_content)
 
     @classmethod
     def is_html_mention_tag(cls, tag: Tag) -> bool:
-        return (
-            cls.MENTION_TAG_NAME == tag.name
-            and tag.has_attr("id")
-            and tag["id"].startswith(cls.MENTION_ID_START)
-        )
+        return cls.MENTION_TAG_NAME == tag.name
```
Millefeuille42, line 92: Must check if the mention has the necessary attributes; userid or roleid

Millefeuille42 (changes requested):

Millefeuille42:
Bugs I encountered while testing:

- Pressing tab does not autocomplete the mention
- Changing language does not change the keyword for the `@all` mention. It is always `@tous`.
- In french, the autocompletion for `@tous` does not show up. It only shows up in other languages

Skylsmoi:
issues found during tests:
- 2 mentions in a single comment generates 2 mentions
- a tricky case when I write:
```
      @aze salut ça va ?

      @azeazeaze

      oui bonjour @aze ça va ?
```
displays a mention on the second line on @ azeazeaze where it shouldn't and the last @ aze isn't displayed as a mention

Haskaris:
@Skylsmoi 
> 2 mentions in a single comment generates 2 mentions
This is this issue #4799 (which is already existing before my refactoring). I've got a fix, however as said in [this comment](https://github.com/tracim/tracim/issues/4799#issuecomment-878925573), it might be a choice. Unfortunally I don't have the knowledge about that and feel it as a bug. But since the point "A note could contain multiple tasks and one needs to be able to notice all the mentions." make sense to me... What is your thought about that?
> a tricky case when I write
This issue was also for content. Both fixed.

Haskaris:
@Millefeuille42 
> - Changing language does not change the keyword for the ``@all`` mention. It is always ``@tous``.
> - In french, the autocompletion for ``@tous`` does not show up. It only shows up in other languages

The translations for ``all``, ``reader``, ``contributor``, ``content-manager`` and ``space-manager`` aren't done yet. So it make sense that it's not translated. However, I don't understand why you have ``tous``, maybe a bad compilation?

> Pressing tab does not autocomplete the mention

That's quite tricky actually... The autocompleter from TinyMCE doesn't seems to support it. I'll investigate to create a custom validation.

lebouquetin:
> > 2 mentions in a single comment generates 2 mentions
> This is this issue https://github.com/tracim/tracim/issues/4799 (which is already existing before my refactoring). I've got a fix, however as said in https://github.com/tracim/tracim/issues/4799#issuecomment-878925573, it might be a choice. Unfortunally I don't have the knowledge about that and feel it as a bug. But since the point "A note could contain multiple tasks and one needs to be able to notice all the mentions." make sense to me... What is your thought about that?

The user will probably think tracim is buggy, so I suggest to notify user once only. It would be interesting to get two notifications if the two notifications were different (eg like having each a specific sentence).

Example:

```
- @ matthew you should read xxx
- @ matthew what do you think of the last release of QT?
```

-> in this case, you could get 2 distinct mentions and it would be interesting _if_ the text associated to the mention is included in the notification.

The other way to process this is to directly send 1 notification which says "lebouquetin mentionned you twice"

lebouquetin:
> The translations for all, reader, contributor, content-manager and space-manager aren't done yet. So it make sense that it's not translated. However, I don't understand why you have tous, maybe a bad compilation?

I do not understand this kind of answer. These are translations which are already available, no? I do not expect any work in order to re-establish translations ... do I misunderstand something?

Haskaris:
@lebouquetin I can fix in this PR the double mention, but not the double mention in a list change the notification text. It's a whole new development to do.
For the translation thing, yes we already do, but they are hard codded in the frontend code. This means, for every new language, we had to manually add the string. With the new mechanism, we can use the power of weblate to translate `@all`. Therefore they are not _yet_ translated. It was an arbitrary choice to not put already translated strings into weblate ; give the opportunity to everyone to contribute.

lebouquetin:
It makes sense @Haskaris for both subjects. Thanks for the explanation. Focus on the goal of your development, both subjects are distincts and may be fixed on association with an other issue

Quiwy:
I don't know if it should be considered as a regression.
With the old system, when inserting an “@”, a list of people was displayed.
Now, with this PR, we need at least 1 character to be written after the “@”.

If you have blackouts when writing sometimes, it was very helpfull to have autocompletion without remembering username / name.

 
mention_before_after

frontend_lib/src/component/TinyEditor/TinyEditor.jsx
```diff
@@ -0,0 +1,400 @@
...
+  const inputRef = useRef(null)
+
+  let defaultRoleList = []
+
+  const toolbar = props.isAdvancedEdition ? advancedToolBar : simpleToolBar
+  // NOTE - MP - 2023-01-10 - Changing the key allow reloading the Editor component this allows us
+  // to change the toolbar
+  const editorKey = props.isAdvancedEdition ? 'advanced_key' : 'simple_key'
+
+  useEffect(() => {
```
grignards, line 71: I'm unsure a `useEffect()` is helpful here, what is the goal to do this instead of directly computing `defaultRoleList` in the component? (`const defaultRoleList = props.roleList.map(...)`
If it is a question of optimization, the [useMemo()](https://reactjs.org/docs/hooks-reference.html#usememo) effect seems a better pick.

raphj:
Mega PR, @Haskaris is a serial fixer :sweat_smile:

All my annoying, wysiwyg editor-related tickets are getting closed xD

Haskaris:
Tracim 4.5.0 is definitely something 🚀

Issue #6231, opened 2023-09-07
✨ Feat: Implement SAML authentication flow 


Millefeuille42:
### Feature description

This issue consolidates components of https://github.com/tracim/tracim/issues/6209 and focuses on implementing the SAML authentication flow in tracim.

### How It Works

The SAML authentication flow operates as follows:

![SAML Flow Diagram](https://github.com/tracim/tracim/assets/42064302/540fe07b-d8cb-4eaa-972d-fe9b1317cd72)

- The Identity Provider (IdP) is an external service responsible for authenticating users at the request of the Service Provider (SP), which in this case is tracim.
- The IdP has its own independent authentication mechanism.
- Both the SP and IdP must provide metadata via a specific endpoint, which contains information about endpoints and authentication requirements.

The process unfolds as follows:

1. The SP (tracim) redirects the user to the IdP, initiating the Single Sign-On (SSO) request.
2. The IdP challenges the user for authentication.
3. After authentication, the IdP redirects the user back to the SP with a POST request, containing session information. This is known as the Assertion Consumer Service (ACS) request.

From this point onward, tracim should perform the following tasks:

- Analyze the session information (e.g., check for expired requests, validate IdP's signature, etc.).
- Extract user data from the session information, such as Email, Username, and unique ID, which are referred to as attributes.
- Check if the user exists in tracim; if not, create a new user.
- Create a tracim session for the user, effectively connecting them to tracim.

De-authentication of the user is a separate task covered in https://github.com/tracim/tracim/issues/6211

#### Tasks to Complete

- [x] Implement Metadata Endpoint
- [x] Create a Generic Single Sign-On (SSO) Endpoint to redirect users to the appropriate IdP SSO endpoint
- [x] Set up the ACS Endpoint for redirection as a POST request
- [x] Match IdP-provided attributes with tracim's user fields (e.g., mail, username, unique id)
- [x] Implement user creation if they do not exist
- [x] Implement user authentication

### Related PR

https://github.com/tracim/tracim/pull/6232

PR #6214, opened 2023-07-26
Refactor: Security policy

grignards:
See #6233

## Checkpoints

**For developers**

- [x] If relevant, manual tests have been done to ensure the stability of the whole application and that the involved feature works
- [x] The original issue is up to date w.r.t the latest discussions and contains a short summary of the implemented solution
- [x] Automated tests covering the feature or the fix, have been written, deemed irrelevant (give the reason), or an issue has been created to implement the test (give the link)
- [x] Make sure that:
  - if there are modifications in the Tracim configuration files (eg. `development.ini`), they are documented in `backend/doc/setting.md`
  - any migration process required for existing instances is documented
  - relevant people for these changes are notified
- [x] Original authors of the features included in a multi-feature branch (maintenance fixes -> develop, security fixes -> develop, …) should be part of the reviewers, especially if you encountered merge conflicts.

**For code reviewers**

- [x] The code is clear enough
- [x] If there are FIXMEs in the code, related issues are mentioned in the FIXME
- [x] If there are TODOs, NOTEs or HACKs in code, the date and the developer initials are present

**For testers**

- ~~[ ] Manual, quality tests have been done~~ => no manual test needed as this is a refactor and the automated test coverage is enough to catch errors.

Issue #6212, opened 2023-07-25
📝 Doc: Document SAML authentication configuration


grignards:
### Which file(s) on Tracim is affected?

backend doc

### Which part(s) of the file would you like to see updated?

SAML authentication support comes with new configuration possibilities which should be properly documented:

- [x] #6210 
- [x] #6209 

Also, a developer documentation should be integrated, even if not properly designed. For example, schemas like  the one in #6231 is interesting to be integrated.

Documentation ust be available in:

- [x] `development.ini.sample`
- [x] `doc/backend/setting.md`

3 things bellow are left to be done. They will be implemented in another issue, see comments
- #6206
- add doc
  - `doc/TESTING.md`
  - `/tools_docker/test_docker/README.md`

### Related PR

https://github.com/tracim/tracim/pull/6234
https://github.com/tracim/tracim/pull/6232

lebouquetin:
I moved back the issue to TODO in order @Skylsmoi test to deploy the service and validate/invalidate the documentation

The expected process is:
- @Skylsmoi tries to install/configure
- @Millefeuille42 is available for help (but no intervention)
- @Skylsmoi updates the documentation according to what @Millefeuille42 says.

Millefeuille42:
Documentation can be found in `doc/backend/setting.md` from line 169

Skylsmoi:
PR for improving doc: https://github.com/tracim/tracim/pull/6260

Note that documentation about tests isn't made yet

Issue #6211, opened 2023-07-25
✨ Feat: Allow SAML Single Logout


grignards:
### Feature description

SAML Single Logout is a technique which allows to disconnect from an IdP coming from a SP (Tracim here).

To support this logout mechanism Tracim must be able to redirect to the IdP's SLO url when the user clicks on the logout button.

### How it works

- [x] when auth_type is `SAML`, the logout button should be a link to a backend url.
- [x] when this url is reached the backend should redirect to the user's IdP SLO url.
- [x] when a `<LogoutResquest> from the IdP is received, the backend should end its user's session.

### Open questions

- Should this logout method the only one when SAML is used? Or should local logout AND Single Logout coexist?

### Related PR

#6237
https://github.com/tracim/tracim/pull/6232

Millefeuille42:
@lebouquetin

I believe that implementing the described approach may not be the most effective way. Instead, the logout method should be dynamically selected based on the user's login method when they access the logout endpoint.

In this particular scenario, the preferred action is to redirect the user to the Identity Provider's Single Log Out (SLO) page. Moreover, a dedicated route will be created to support the SAML SLO endpoint, ensuring compliance with standards. This route will handle logout requests from the Identity Provider. By combining both mechanisms, it results in no to little additional workload compared to using either approach exclusively, and in fact, it might even reduce the complexity.

This implementation is expected to be straightforward and is well-suited for long-term use, seamlessly integrating into the existing system while leaving room for potential integration of other logout mechanisms in the future.

lebouquetin:
I'm not sure of what you mean. Here is what I understand being the best way to implement logout:

- identify the logout method based on login method (are you able to identify it? how?)
- in case of "classical login" the logout keep being "classical"
- in case of SAML, the logout process is the one described by @grignards in the description (redirect to IdP logout process)

Is this hat you have in mind? @Millefeuille42

Millefeuille42:
Yeah this is somewhat what I was thinking, the changes are all done in backend on the logout route.

Considering that the `auth_typ`e for each user is stored in the DB, we could guess based on that (this or add per session `auth_type` in the user's session data)

![image](https://github.com/tracim/tracim/assets/42064302/655b528a-8907-4a6c-ae7a-0be5e52cbc35)

This would then permit the selection of a logout method depending on the `auth_type`, just like we're already doing for login.

```
        for auth_type in self._config.AUTH_TYPES:
            try:
                return self._authenticate(
                    login=login,
[...]

PR #6232, opened 2023-09-07
Feat: Implement SAML authentication flow

Millefeuille42:
https://github.com/tracim/tracim/issues/6206
https://github.com/tracim/tracim/issues/6209
https://github.com/tracim/tracim/issues/6210
https://github.com/tracim/tracim/issues/6211
https://github.com/tracim/tracim/issues/6212
https://github.com/tracim/tracim/issues/6231
https://github.com/tracim/tracim/issues/6246
https://github.com/tracim/tracim/issues/6248
https://github.com/tracim/tracim/issues/6258

~~This PR originates from https://github.com/tracim/tracim/pull/6214 this needs to be sorted out before merging (by merging the originating PR for example)~~ EDIT: Done

## Checkpoints

**For developers**

- [x] Make sure that:

**For code reviewers**

**For testers**


Millefeuille42:
What needs to be done

- [ ] Conformity with tracim's code (make sure their is no better way to do X and Y)
- [ ] Adapt code style to match styling rules
- [ ] Reading and matching attributes per IdP from a config file
- [ ] Storing authn_response data in session and checking for session lifetime or validity when validating the session
- [ ] Cleanup eventual leftovers
- [ ] (Maybe OOS) Integrate the JSON config file as part of development.ini ?

backend/tracim_backend/lib/utils/authentication.py
```diff
@@ -20,16 +28,195 @@
...
+            self.saml_config.allow_unknown_attributes = True
+            self.saml_config.attribute_map = attribute_map
+        self._metadata_response = Response(content_type='application/xml')
+        self._metadata_response.text = create_metadata_string(config=self.saml_config,
+                                                              configfile=None).decode()
+        self.saml_client = Saml2Client(config=self.saml_config)
+
+        for _, data in self.saml_client.config.vorg.items():
+            if data.common_identifier not in app_config.IDP_LIST:
+                app_config.IDP_LIST.append(data.common_identifier)
```
Millefeuille42, line 74: This is the process of adding the configuration from `backend/settings_saml2.json` (as well as adding mandatory parameters, such as `self.saml_config.allow_unknown_attributes = True`

`self.saml_config.attribute_map = attribute_map` and the definition of `attribute_map` might be unnecessary

Some preparation for the metadata endpoint is also done, it's done to compute the metadata response (and fetch metadata from the remote IdPs) once at startup, to prevent slowdowns of the request handling

The list of IdPs is also added to the `app_config` under the `IDP_LIST` attribute

backend/tracim_backend/lib/utils/authentication.py
```diff
@@ -20,16 +28,195 @@
...
+
+        if 'pyramid_saml' not in config.get_settings():
+            msg = 'Missing \'pyramid_saml\' configuration in settings'
+            raise AssertionError(msg)
+        else:
+            settings = config.get_settings().get('pyramid_saml')
+
+        if 'saml_path' not in settings:
+            msg = 'Missing \'saml_path\' in settings'
+            raise AssertionError(msg)
```
Millefeuille42, line 110: This function just tries to get the path of the JSON settings file from the config / the ENV, this could be adjusted to get the info from the `app_config` (INI file)

It basically just throws an error if the path if found nowhere

backend/tracim_backend/lib/utils/authentication.py
```diff
@@ -20,16 +28,195 @@
...
+        # The form should redirect to the chosen IdPs SSO endpoint when submitted.
+        # Maybe some JS will be needed here?
+        return Response(
+            f"""
+          <form action="{self.app_config.WEBSITE__BASE_URL}/saml/acs" method="POST">
+            <input name="saml_user_id" type="hidden" value="saml-user"/>
+            <input name="redirect_url" type="hidden" value="{request.path_url}" />
+            <button>Login</button>
+          </form>"""
+        )
```
Millefeuille42, line 169: This function is deemed useless at the time, but might me be useful someday ?

backend/tracim_backend/lib/utils/authentication.py
```diff
@@ -20,16 +28,195 @@
...
+        return user.user_id
+
+    def permits(self, request, context, permission):
+        # TODO - M.L - 2023/09/06 - Maybe check for session lifetime or validity
+        # if request.authenticated_userid is None \
+        #     and not request.path.startswith("/saml") \
+        #     and not request.path.startswith("/saml") \
+        #     and not request.path is "/ui/login":
+        #     return Denied("Nobody is allowed")
+        return Allowed("Allowed")
```
Millefeuille42, line 141: permits just, uh permits at the time. The initial blocking pattern was too exclusive and blocked access to too many ressources. A proper check of the session lifetime or validity must be done here.

backend/tracim_backend/lib/utils/authentication.py
```diff
@@ -20,16 +28,195 @@
...
+            return HTTPNotFound("This IdP doesn't exist")
+        _, info = self.saml_client.prepare_for_authenticate(
+            entityid=target,
+        )
+
+        redirect_url = None
+        for key, value in info['headers']:
+            if key == 'Location':
+                redirect_url = value
+        return HTTPFound(redirect_url)
```
Millefeuille42, line 211: This functions is the entrypoint for the SAML auth flow, it tries to get the IdP chosen name from the `target` query parameter and tries to match it with a `vorg` entry from the config (see details about the config below). If it finds a match, it redirects to it's sso endpoint.

backend/tracim_backend/lib/utils/authentication.py
```diff
@@ -20,16 +28,195 @@
...
+        if identity is None or "UserID" not in identity:
+            return HTTPBadRequest()
+
+        # TODO - M.L - 2023/09/06 - Associate response / expiry date to session
+        request.session["saml_user_id"] = ''.join(identity["UserID"])
+        request.session["saml_mail"] = ''.join(identity["EmailAddress"])
+        request.session[
+            "saml_name"] = f'{"".join(identity["FirstName"])} {"".join(identity["LastName"])}'
+
+        return HTTPFound("/")
```
Millefeuille42, line 190: This functions parses the response from the IdP (well, technically from the users, as the response is encoded in the POST data but well) does some little checks (is the IdP known, is the response expired, etc...) and matches hardcoded attributes to session storage. Authentication checks are then done with the `authenticated_userid` function

backend/tracim_backend/lib/utils/authentication.py
```diff
@@ -20,16 +28,195 @@
...
+
+        user = user_api.saml_authenticate(
+            user=None,
+            user_id=saml_user_id,
+            name=request.session.get("saml_name"),
+            mail=request.session.get("saml_mail"),
+        )
+        self.remember(request, user.user_id)
+        request.set_user(user)
+        return user.user_id
```
Millefeuille42, line 132: This function (which seems to be called each time to check the session), checks the session data and calls the `saml_authenticate` function, which will try to find the user with the `user_id` and if it does not find it, creates it.

backend/tracim_backend/lib/core/user.py
```diff
@@ -670,12 +681,66 @@ def authenticate(self, password: str, login: str, ldap_connector: "Connector" =
...
+
+        if user.is_deleted:
+            raise UserDoesNotExist("This user has been deleted")
+
+        if not user.is_active:
+            raise UserAuthenticatedIsNotActive("This user is not activated")
+
+        if user.auth_type == AuthType.UNKNOWN:
+            user.auth_type = auth_type
+        return user
```
Millefeuille42, line 734: This function, like other similar functions, will attempt to find a user matching the provided user_id. If it can't find one, it will create a user with the provided information.

Millefeuille42:
Borrowed from https://github.com/tracim/tracim/issues/6205

- `entityid`: Is the SP (us) metadata endpoint
- `metadata`: Corresponds to the metadata endpoint of each IdP, they are then queried and parsed to provide their entityId in saml_client.metadata.metadata
- `service`: Corresponds to the listing of endpoints the service provider (us) provides
- `virtual_organization`: Used to bind metadata URLs to common_identifiers, to provide human readable identifiers

```
{
        "entityid": "http://127.0.0.1:6543/saml2/metadata/",
        "metadata": {
            "local": [],
            "remote": [
                {
                    "url": "https://samltest.id/saml/idp"
                }
            ]
        },
        "service": {
            "sp": {
                "endpoints": {
                    "assertion_consumer_service": [
                        [
                            "http://127.0.0.1:6543/saml2/acs",
                            "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST"
                        ]
                    ],
                    "single_logout_service": [
                        [
                            "http://127.0.0.1:6543/saml2/slo/post",
                            "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect"
                        ],
                        [
                            "http://127.0.0.1:6543/saml2/slo/redirect",
                            "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST"
                        ]
                    ]
                },
                "allow_unsolicited": true,
                "authn_requests_signed": false,
                "logout_requests_signed": false,
                "want_assertions_signed": false,
                "want_response_signed": false,
                "name_id_format": "urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified",
                "name_id_format_allow_create": true,
                "want_name_id": true
            }
        },
        "key_file": "./sp.key",
        "cert_file": "./sp.crt",
        "xmlsec_binary": "/usr/bin/xmlsec1",
        "metadata_cache_duration": {
            "default": 86400
        },
        "virtual_organization": {
            "https://samltest.id/saml/idp": {
                "common_identifier": "idp_0"
            }
        }
    }
```

Millefeuille42:
Should the code be cleaned up after integration, creating follow up issues, or we want to merge clean code only ?

frontend/src/css/Login.styl
```diff
@@ -23,6 +23,12 @@
       margin-block-start 15%
       display flex
       flex-direction column
+      &__saml
+        &__link
+          cursor pointer
+          font-size met adataFontSize
```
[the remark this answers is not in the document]
Millefeuille42: fixed

backend/tracim_backend/lib/core/user.py
lebouquetin: @Millefeuille42 is there an external_id mechanism?

Millefeuille42: Not currently

backend/tracim_backend/lib/core/user.py
```diff
@@ -670,12 +681,66 @@ def authenticate(self, password: str, login: str, ldap_connector: "Connector" =
...
+        """
+        auth_type = AuthType.SAML
+
+        # TODO - M.L - 2023/09/06 - Change this to get saml response and isolate attributes
+
+        if not user:
+            try:
+                user = self.get_one_by_login(user_id)
+            except UserDoesNotExist:
+                _ = self.create_user(
```
lebouquetin, line 711: Here you need to check external_id in order to update an existing account or to create a new one. I suggest you to add a TODO so we can merge and find back where to fix #6258

lebouquetin: By the way, I'm afraid this wont work. The SAML IdP can't know our internal `user_id`.

backend/tracim_backend/lib/core/user.py
```diff
@@ -670,12 +681,66 @@ def authenticate(self, password: str, login: str, ldap_connector: "Connector" =
...
+        or UserAuthenticatedIsNotActive
+        :param user: user to check,, can be none if user not found, will try
+         to create new user if none but ldap auth succeed
+        :param user_id: user_id of the user
+        :param name: name extracted from saml attributes
+        :param mail: mail extracted from saml attributes
+        """
+        auth_type = AuthType.SAML
+
+        # TODO - M.L - 2023/09/06 - Change this to get saml response and isolate attributes
```
lebouquetin, line 705: What is this TODO @Millefeuille42 ? Is it still up-to-date?

backend/tracim_backend/lib/utils/authentication.py
```diff
@@ -32,41 +44,121 @@
 AUTH_TOKEN_QUERY_PARAMETER = "access_token"
 CLIENT_TOKEN_HEADER = "X-Tracim-ClientToken"
 
+SAML_IDP_DEFAULT_CONFIG = {
+    "default": {
+        "logo_url": "",
+        "displayed_name": "",
+        "attribute_map": {
+            "user_id": "${mail}",
```
lebouquetin, line 53: I'm not sure to understand. I'm pretty sure in tracim user_id is an integer. This can't work

lebouquetin (changes requested):

backend/tracim_backend/lib/utils/authentication.py
```diff
@@ -32,41 +44,121 @@
...
+
+        if "pyramid_saml" not in config.get_settings():
+            msg = "Missing 'pyramid_saml' configuration in settings"
+            raise AssertionError(msg)
+        else:
+            settings = config.get_settings().get("pyramid_saml")
+
+        if "saml_path" not in settings:
+            msg = "Missing 'saml_path' in settings"
+            raise AssertionError(msg)
```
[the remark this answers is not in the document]
Millefeuille42: It does exit when this variable is not set.

backend/tracim_backend/lib/core/user.py
Millefeuille42: Nope, I will remove it

backend/tracim_backend/lib/utils/authentication.py
Millefeuille42: I messed up the naming. User_Id actually binds to tracim's username. I will make it clearer with https://github.com/tracim/tracim/issues/6258

lebouquetin: If so @Millefeuille42 please add a checkbox in #6258 description so we are sure not to forget it.

backend/tracim_backend/lib/utils/authentication.py
```diff
@@ -32,41 +44,121 @@
...
 
+SAML_IDP_DEFAULT_CONFIG = {
+    "default": {
+        "logo_url": "",
+        "displayed_name": "",
+        "attribute_map": {
+            "user_id": "${mail}",
+            "email": "${eduPersonUniqueId}",
+            "username": "${commonName}",
+            "display_name": "${surName} ${givenName}",
```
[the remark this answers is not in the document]
lebouquetin: If you plan to change it at the same time as user_id, then add an extra checkbox in #6258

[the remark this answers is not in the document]
Millefeuille42: Will do

backend/tracim_backend/lib/utils/authentication.py
Millefeuille42: Will do

backend/tracim_backend/lib/core/user.py
Millefeuille42: Will remove it in #6258 (and will add a checkbox for it)

backend/tracim_backend/lib/utils/authentication.py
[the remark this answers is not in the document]
Skylsmoi: done in https://github.com/tracim/tracim/pull/6259

backend/tracim_backend/lib/core/user.py
Skylsmoi: done in https://github.com/tracim/tracim/pull/6259

backend/tracim_backend/lib/core/user.py
Skylsmoi: removed in https://github.com/tracim/tracim/pull/6259

backend/tracim_backend/lib/utils/authentication.py
Skylsmoi: done in https://github.com/tracim/tracim/pull/6259

Skylsmoi:
@Millefeuille42 I have found an issue during my tests:
- Have a user U1 created from a SAML auth id ID1 (I used the OpenSSO)
- Change the username of U1 in tracim
- Login again through SAML with auth id ID1
- backend throw error 500: "tracim_backend.exceptions.EmailAlreadyExists: Email given come.huguies@algoo.fr already exist, please choose something else"
- and in frontend we see a page with only:
"Internal Server Error

The server encountered an unexpected internal server error

(generated by waitress)"

lebouquetin:
@Skylsmoi do you get a JSON answer or a raw HTML page?

Skylsmoi:
raw html page

lebouquetin:
@Skylsmoi could you give me the associated route? I'll give it a look

Millefeuille42:
@Skylsmoi This is fixed by #6259. However I have to improve the error management to prevent 500 statuses.

lebouquetin:
> However I have to improve the error management to prevent 500 statuses.

actually, all frontend-targeting routes MUST NEVER return raw HTML. The use of hapic is/should be done this way. This does not prevent to have HTTP 500 but even HTTP 500 must be JSON formatted.

This is the reason why I asked for the route: it is probably missing a decorator

Millefeuille42:
@lebouquetin The issue is that the exception is from an overriden method of the auth policy. So it's not caught by a view function.

What needs to be done is to catch the exception (log it?) , and return `None` on exception.

I'm talking about this function:

```
    def authenticated_userid(
        self, request: TracimRequest
    ) -> typing.Optional[typing.Union[str, int]]:
        saml_user_id = request.session.get("saml_user_id")
        if saml_user_id is None:
            return None
        saml_expiry = request.session.get("saml_expiry")
        if saml_expiry is None or saml_expiry <= int(time.time()):
            return None

        user_api = UserApi(None, request.dbsession, self.app_config)

        user = user_api.saml_authenticate(
            user=None,
            external_id=saml_user_id,
            username=request.session.get("saml_username"),
            public_name=request.session.get("saml_public_name"),
            mail=request.session.get("saml_email"),
            profile=request.session.get("saml_user_profile"),
        )
        self.remember(request, user.user_id)
        request.set_user(user)
        return user.user_id
```

lebouquetin:
I understand @Millefeuille42 , thanks for the explanation.

Reading at the following code and according to the fact that authentication updates now the user in the database, does it mean that updating by hand the profile (on the client side) would allow to escalate privileges?
```
        user = user_api.saml_authenticate(
            user=None,
            external_id=saml_user_id,
            username=request.session.get("saml_username"),
            public_name=request.session.get("saml_public_name"),
            mail=request.session.get("saml_email"),
            profile=request.session.get("saml_user_profile"),
        )
```

Skylsmoi:
I reproduced the issue and it's not actually html, it's plain text

and the route is http://localhost:7999/

Millefeuille42:
@lebouquetin I mean it could, but it's session data, is it even possible ?

Millefeuille42:
> Also, creating this issue creates a cookie that will try to login with that same credentials and always trigger the error 500. Making it impossible to use tracim since we can't access the login page

Thanks for the insight. I will add that the current session is also deleted when an exception occurs.

Millefeuille42:
@lebouquetin @Skylsmoi I fixed the 500 errors, the login is now juste considered as failed and the error is logged in the console.

lebouquetin:
> @lebouquetin I mean it could, but it's session data. Is it even possible to edit it correctly client side ?

This is a good question. I'm pretty sure it's not actually: session data on client side is only the token, isn't it?

Millefeuille42:
> > @lebouquetin I mean it could, but it's session data. Is it even possible to edit it correctly client side ?
> 
> This is a good question. I'm pretty sure it's not actually: session data on client side is only the token, isn't it?

From what I know (in general, not about tracim. I must read doc about beaker for this) sessions are usually either JWTs in this case data is encoded in the token, and "encrypted" using a secret key, or sessions keys, linked to session data that is stored on the backend.

So unless it doesn't work like this in tracim, we can safely assume this is safe from client-side tampering (will make sure of this by reading documentation of course)

backend/tracim_backend/lib/utils/authentication.py
```diff
@@ -32,41 +46,148 @@
...
+    #  as few times as possible, meaning that if it meets a '}', it will not be matched
+    placeholders = re.findall(r"\${(.*?)}", pattern)
+    formatted_value = pattern
+    for placeholder in placeholders:
+        if placeholder in attributes:
+            formatted_value = formatted_value.replace(
+                f"${{{placeholder}}}", "".join(attributes[placeholder])
+            )
+    return formatted_value
+
```
buxx, line 82: @Millefeuille42 Probably more simple like this ?

```python
template = "Foo:{foo}, Bar:{bar}, Baz:{baz}"
attributes = {"foo": "42", "bar": "42", "baz": "42"}
print(template.format(**attributes))
```

Note : template.format will fail if key is absent from `attributes` (`KeyError: 'baz'`) and will not fail if `attributes` contain an extra key.

backend/tracim_backend/views/core_api/user_controller.py
```diff
@@ -162,6 +164,20 @@
...
+) -> bool:
+    """
+    Check if user can write provided field (is admin or field is writable)
+    :raise ReadOnlyFieldException if the field is not writable
+    """
+    read_only_fields = app_config.USER__READ_ONLY_FIELDS.get(request.candidate_user.auth_type)
+    if read_only_fields is not None:
+        if field in read_only_fields:
+            raise ReadOnlyFieldException(f"You can't modify the field {field.value}")
+    return True
```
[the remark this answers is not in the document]
buxx: And, as I see about `check_has_write_rights` usage, return value probably unnecessary

buxx (approved):

backend/tracim_backend/views/core_api/user_controller.py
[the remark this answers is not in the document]
Millefeuille42: @buxx iirc it was specifically asked to raise an exception if invalid. However it's open to discussion !

Millefeuille42:
Tests ok on local

Issue #6235, opened 2023-10-16
✨ Feat: add the ability to access configurators in plugins loaded hook


Millefeuille42:
### Feature description

Plugins should have access in a way to the `Configurator` in order to add routes and views (among other things) to Pyramid on their `on_plugins_loaded` hook.

### How it works

`Configurator` should be created and it's `Registry` should be pushed to the `Pyramid.threadlocal` stack  before loading plugins so plugins can access the `Registry` and the `Configurator` on their `on_plugins_loaded` hook.

PR #6236, opened 2023-10-16
Feat: add the ability to access configurator in plugins loaded hook

Millefeuille42:
#6235

## Checkpoints

**For developers**

- [x] Make sure that:

**For code reviewers**

**For testers**

### DIFF
--- a/backend/tracim_backend/__init__.py
+++ b/backend/tracim_backend/__init__.py
@@ -131,11 +131,13 @@ def web(global_config: OrderedDict, **local_settings) -> Router:
     app_config.configure_filedepot()
     settings["CFG"] = app_config
 
+    configurator = Configurator(settings=settings, autocommit=True)
+    configurator.begin()
+
     # Init plugin manager
     plugin_manager = init_plugin_manager(app_config)
-    settings["plugin_manager"] = plugin_manager
+    configurator.registry.settings["plugin_manager"] = plugin_manager
 
-    configurator = Configurator(settings=settings, autocommit=True)
     # Add beaker session cookie
     tracim_setting_for_beaker_session = sliced_dict(settings, beginning_key_string="session.")
     tracim_setting_for_beaker_session["session.data_dir"] = app_config.SESSION__DATA_DIR
@@ -333,6 +335,8 @@ def web(global_config: OrderedDict, **local_settings) -> Router:
     plugin_manager.hook.web_include(configurator=configurator, app_config=app_config)
 
     hapic.add_documentation_view("/api/doc", "Tracim API", "API of Tracim")
+
+    configurator.end()
     return configurator.make_wsgi_app()
 
 
--- a/doc/backend/hello_world_plugin.py
+++ b/doc/backend/hello_world_plugin.py
@@ -7,18 +7,18 @@ from tracim_backend.lib.utils.request import TracimContext
 from tracim_backend.models.auth import User
 
 
-@hookimpl
-def on_plugins_loaded(plugin_manager: PluginManager) -> None:
-    """Automatically loaded."""
-    print("Hello World !")
-
-
-class HookImpl:
+class HelloWorldPlugin:
     """Needs a registration using 'register_tracim_plugin' function."""
 
+    @hookimpl
+    def on_plugins_loaded(plugin_manager: PluginManager) -> None:
+        """Automatically loaded."""
+        print("Hello World !")
+
+    @hookimpl
     def on_user_created(self, user: User, context: TracimContext) -> None:
         print("created user {}".format(user.public_name))
 
 
 def register_tracim_plugin(plugin_manager: PluginManager):
-    plugin_manager.register(HookImpl())
+    plugin_manager.register(HelloWorldPlugin())
--- a/doc/backend/plugins.md
+++ b/doc/backend/plugins.md
@@ -74,7 +74,10 @@ The `register_tracim_plugin` function is used to register the plugin with the `P
 To register routes in a plugin, use the following snippet in the `on_plugins_loaded` hook:
 
 ```python
+from pyramid.config import Configurator
+from pluggy import PluginManager
 from pyramid.threadlocal import get_current_registry
+from tracim_backend.lib.core.plugins import hookimpl
 
 @hookimpl
 def on_plugins_loaded(self, plugin_manager: PluginManager) -> None:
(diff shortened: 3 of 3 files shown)


backend/tracim_backend/__init__.py
```diff
@@ -131,11 +131,13 @@ def web(global_config: OrderedDict, **local_settings) -> Router:
     app_config.configure_filedepot()
     settings["CFG"] = app_config
 
+    configurator = Configurator(settings=settings, autocommit=True)
+    configurator.begin()
+
     # Init plugin manager
     plugin_manager = init_plugin_manager(app_config)
-    settings["plugin_manager"] = plugin_manager
+    configurator.registry.settings["plugin_manager"] = plugin_manager
```
Millefeuille42, line 139: This creates the `Configurator` *before* loading the plugins, making the `Registry` and, therefore, the `Configurator` accessible on the `threadlocal` stack for plugins to register routes, views, modify settings...

`configurator.begin()` pushes the `Registry` on the `threadlocal` stack. It is later popped with `configurator.end()`

`Registry` and `Configurator` can be accessed in this manner within a plugin's `on_plugins_loaded` hook.

```python
    @hookimpl
    def on_plugins_loaded(self, plugin_manager: PluginManager) -> None:
        """
        This method is called when the plugin is loaded.
        """
        from pyramid.threadlocal import get_current_registry
        registry = get_current_registry()
        config = Configurator(registry=registry)
        config.add_route("my_route", "/plugins/my_route", request_method="GET")
        config.add_view(self.my_view, route_name="my_route")
 ```

File: backend/tracim_backend/__init__.py [hunk 2]
```diff
@@ -333,6 +335,8 @@ def adapt_auth_policy(authentication_policy) -> SecurityPolicyAdaptor:
     plugin_manager.hook.web_include(configurator=configurator, app_config=app_config)
 
     hapic.add_documentation_view("/api/doc", "Tracim API", "API of Tracim")
+
+    configurator.end()
```
Millefeuille42, line 339: This pops the `Registry` from the `threadlocal` stack, propagating the changes.

backend/tracim_backend/__init__.py
Millefeuille42: Since the hook is called before starting the server and (at least I think so) the whole operation is single threaded, this is safe for use and follows standard usage of Pyramid's `Configurator`

PhilippeAccorsi:
This modification can have impact on existing plugin ? Something change to use actual plugin ( cf https://github.com/tracim/tracim/tree/develop/backend/official_plugins ) or not ?

Millefeuille42:
> This modification can have impact on existing plugin ? Something change to use actual plugin ( cf https://github.com/tracim/tracim/tree/develop/backend/official_plugins ) or not ?

@PhilippeAccorsi This has no impact whatsoever; it solely involves enhancing plugin functionality and does not require any modification to existing plugins.

buxx:
This PR content looks okay !

buxx (approved):

buxx:
Doc is :+1:

Millefeuille42:
Due to the nature of the changes there is not much to test

- Make sure that configuration still works and that changes are applied
- Test the code snippet provided in comments to see if a route is indeed registered in a plugin

backend/tracim_backend/__init__.py
lebouquetin: > (at least I think so) [...] this is safe for use

Could you check this information and put it in the example?

You should also:
- put a link to this documentation in the code example
- move the example here to the #6235 issue description in order to make information available for future readers (nobody reads PR comments)

> and follows standard usage of Pyramid's Configurator

Could you link to the associated documentation? (in the code and in the issue description, not in the comments here)

backend/tracim_backend/__init__.py
lebouquetin: > propagating the changes.

To the entire app? If so, I'm afraid the complete system does not work:

- the configuration would be updated in each thread (threadlocal means the execution is local for a thread, not executed in a unique thread) then applied N times according to the number of threads?

I don't know how pyramid is designed for this stuff, I'm ok to merge the PR *in case you are sure of the behavior you expect* and that *it works like you/we imagine*.

lebouquetin:
> Test the code snippet provided in comments to see if a route is indeed registered in a plugin

Also, in a multi-threaded context test if the route is registered one time or multiple times, at least making us understand the behavior of the backend.

backend/tracim_backend/__init__.py
Millefeuille42: This behavior is done internally when providing the `autocommit=True` parameter to the constructor of the `Configurator`. Putting it manually is manually triggering this propagation since we manually popped the `Configurator` to the stack.

To make sure of what I say, I tested it locally. Changes done from the plugins or read from the config file are indeed applied to the app.

I am sure of the behavior and it works as I imagined it.

backend/tracim_backend/__init__.py
Millefeuille42: I will put the required information and links to the documentation in the issue and the code

Millefeuille42:
@lebouquetin I will test more thoroughly in a multithreaded context to make sure everything works as planned. However testers of this PR should also test in single and multithreaded context too.

backend/tracim_backend/__init__.py
Millefeuille42: Ok I understand. To make sure I tested with verifiable events and 4 threads, whether it is prints, requests to external elements etc.. It is only ran once. I tested with `uwsgi` and the following settings:

```
workers = 4
threads = 4
```

I have not tested SQL yet, but it seems configuration is resolved before threading anything or starting processes. I have confirmation that configuration code (and so plugins configuration) is only ran once no matter the numbers of workers/threads

lebouquetin: Nice work @Millefeuille42 . So I suggest to add this information to the documentation - most of developers won't ask for it but the ones who will will also have answer to their questions

PhilippeAccorsi (changes requested):
Just to make sure you have seen this documentation about plugin https://github.com/tracim/tracim/tree/develop/backend/official_plugins

I think, in the past, this documentation is not set in a good folder so we need to make only one documentation I think.

Maybe duplicated information ;)

Millefeuille42:
@PhilippeAccorsi I don't think both documentations are really related, are they ?
One is about how to code and enable plugins, the other is the documentation for the official plugins themself (and not the plugin system)

backend/tracim_backend/__init__.py
Millefeuille42: I added this information in the plugins doc in the "technical information" section. It is easier to find it here, since plugins dev *should* read the documentation (at least more that issues).

lebouquetin:
@Skylsmoi @buxx I'd be happy if this is merged with the today's version (and added to the release notes) - cc @Millefeuille42

Skylsmoi:
ok

@Millefeuille42 how to test this feature on branch?

Millefeuille42:
@Skylsmoi Simply running the backend and see if it runs correctly (both with pserve and uwsgi)

Skylsmoi:
tested on branch

Issue #6216, opened 2023-07-26
✨ Feat: Allow user properties to be read-only


grignards:
### Feature description

When Tracim creates users based on an external (LDAP/remote user/SAML/…) system, it also sets some properties which shouldn't not be editable in Tracim afterwards as this would cause a discrepancy between the external system and Tracim.

### How it works

- [ ] add `read_only` property in `User` class (needed since several authentication mechanism can be active and users from different authentication be present in the database).
- [ ] have a method on `UserApi` like `create_user_from_external_auth` which will then fill the `read_only` list.
- [ ] the `read_only` list must be visible for the frontend
- [ ] the frontend shall then make use of this list to disable edition of properties in the list.
- [ ] (optional) use this mechanism for the LDAP authentication (beware this will require special care in the migration script).

Millefeuille42:
After investigation, I have two solutions in mind.

The first would be to add a read_only field in the user_configs table's JSON.
This field will have a boolean value. This value will be set upon creation and will be evaluated when trying to modify the user.

If the user doing the request is not an admin and the target user is in read-only, block any modification besides the language.

This read-only property could also be evaluated in the frontend to prevent the user from modifying their data (currently the check is only done against the user's authentication type)

The second option is similar in most points, it would add a read-only field in the same spot, but this field will instead be a list of user attributes (username, email...).

Upon modification of the user, if the user doing the request is not an admin, if any of the modified fields are deemed read-only the request will be blocked.

In the frontend, the list of read-only attributes will be evaluated, and associated UI form fields will be disabled.

This logic is made to be evolutive. This will permits to evolve the first solution into the second one quite easily.

The first solution lacks granularity but is faster to implement, and still do not block evolution to the second solution (most of the code will be reused and no DB migration will be required). The second one is completed but slower to implement.

I'll let the decision up to you @lebouquetin

lebouquetin:
The two solutions you propose are on a user-level granularity. What about a more global configuration, related to auth method?

For example, some configuration like this:

```
read_only_fields["standard"] = ""
read_only_fields["ldap"] = "username, public_name, email_addresse"
read_only_fields["saml"] = "username, public_name, email_addresse"
read_only_fields["remote"] = "*"
```

This way, we could configure read-only fields according to the auth mechanism (ie the source of truth).

This could be upgraded to a user-level granularity in case some fields may be defined RO/RW at user level (in this case the configuration would become dynamic - something like a merge of auth-based configuration merged with user-defined configuration

[...]

PR #6251, opened 2023-11-02
feat: allow user properties to be readonly

Millefeuille42:
#6216 

## Checkpoints

**For developers**

- [x] Make sure that:

**For code reviewers**

**For testers**


backend/development.ini.sample
```diff
@@ -376,6 +376,12 @@ api.key = %(basic_setup.api_key)s
 ## TLS usage to communicate with your LDAP server
 ; ldap_tls = False
 
+## Read only fields per auth_type, parameter name is user.profile.read_only_fields.{auth_type}
+# value is a comma separated list of read-only fields.
+# example:
+# user.profile.read_only_fields.saml=username,email,public_name
+# user.profile.read_only_fields.ldap=username,password,email,public_name
+
```
PhilippeAccorsi, line 384: Missing default value (look in begining of sample file "Tracim Config File" about syntax).
Is it important to show default value set in tracim code.

backend/development.ini.sample
```diff
@@ -386,6 +392,8 @@ api.key = %(basic_setup.api_key)s
 ## The profile of those users will be user.default_profile
 ; user.self_registration.enabled = False
 
+
+
```
PhilippeAccorsi, line 396: double line is not in the standard syntax of this file (If i'm not wrong)

backend/development.ini.sample
```diff
@@ -376,6 +376,12 @@ api.key = %(basic_setup.api_key)s
 ## TLS usage to communicate with your LDAP server
 ; ldap_tls = False
 
+## Read only fields per auth_type, parameter name is user.profile.read_only_fields.{auth_type}
+# value is a comma separated list of read-only fields.
```
Skylsmoi, line 380: "fields" are database table's columns?
if yes, maybe it requires in line in the doc, if it's consistent with other parameter's doc

backend/development.ini.sample
PhilippeAccorsi: Maybe add a comment about why set this parameter is "important".

doc/backend/env_settings.md
```diff
@@ -21,6 +21,7 @@ To update the list below you will need to:
 | TRACIM_DEFAULT_LANG                                                       | default_lang                                                   | DEFAULT_LANG                                                       |
 | TRACIM_PREVIEW_CACHE_DIR                                                  | preview_cache_dir                                              | PREVIEW_CACHE_DIR                                                  |
 | TRACIM_AUTH_TYPES                                                         | auth_types                                                     | AUTH_TYPES                                                         |
+| TRACIM_USER__PROFILE__READ_ONLY_FIELDS_XYZ                                | user.profile.read_only_fields.xyz                              | USER__PROFILE__READ_ONLY_FIELDS__XYZ                               |
```
PhilippeAccorsi, line 24: I not understand _XYZ here because not same syntax in development.ini.sample
I think it is better to add one parameter for each supported type of auth_type to be more explicit. If not, I think the minimum is to show all supported value in development.ini.sample

What do you think about that @lebouquetin ?

doc/backend/env_settings.md
```diff
@@ -171,10 +172,6 @@ To update the list below you will need to:
...
 | TRACIM_CALDAV__PRE_FILLED_EVENT__DESCRIPTION_FILE_PATH                    | caldav.pre_filled_event.description_file_path                  | CALDAV__PRE_FILLED_EVENT__DESCRIPTION_FILE_PATH                    |
-| TRACIM_COLLABORATIVE_DOCUMENT_EDITION__SOFTWARE                           | collaborative_document_edition.software                        | COLLABORATIVE_DOCUMENT_EDITION__SOFTWARE                           |
-| TRACIM_COLLABORATIVE_DOCUMENT_EDITION__COLLABORA__BASE_URL                | collaborative_document_edition.collabora.base_url              | COLLABORATIVE_DOCUMENT_EDITION__COLLABORA__BASE_URL                |
-| TRACIM_COLLABORATIVE_DOCUMENT_EDITION__FILE_TEMPLATE_DIR                  | collaborative_document_edition.file_template_dir               | COLLABORATIVE_DOCUMENT_EDITION__FILE_TEMPLATE_DIR                  |
-| TRACIM_COLLABORATIVE_DOCUMENT_EDITION__ENABLED_EXTENSIONS                 | collaborative_document_edition.enabled_extensions              | COLLABORATIVE_DOCUMENT_EDITION__ENABLED_EXTENSIONS                 |
```
PhilippeAccorsi, line 177: Why remove this ?

PhilippeAccorsi (changes requested):
Need some improvement.

frontend/src/component/Account/PersonalData.jsx
```diff
@@ -24,6 +23,13 @@ export class PersonalData extends React.Component {
     }
   }
 
+  isFieldWritable = field => {
+    return !this.props.displayAdminInfo &&
+      this.props.system.config.user__read_only_fields &&
+      this.props.system.config.user__read_only_fields[this.props.userAuthType] &&
+      this.props.system.config.user__read_only_fields[this.props.userAuthType].includes(field)
+  }
+
```
Skylsmoi, line 32: it is more usual to name this kind of function "isEditable"
also, you could take advantage of [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) to reduce the "&&"

And lastly, I think you are using "props.displayAdminInfo" as a "isAdmin" kind of props. I would prefer not to.
I think, since the information about "isAdmin" comes from props, it is the parent component that should give a props for this specific usage. Or, simpler way, give a isAdmin props and update displayAdminInfo props

Skylsmoi (changes requested):

backend/tracim_backend/config.py
```diff
@@ -414,6 +415,21 @@ def _load_global_config(self) -> None:
...
+                separator=",",
+                cast_func=str,
+                do_strip=True,
+            )
+            # HACK - M.L - 2023-10-30 - This is to satisfy CFG._check_consistency despite being
+            #  poorly usable
+            self.__setattr__(
+                f"USER__PROFILE__READ_ONLY_FIELDS__{auth_type.value.upper()}", read_only
+            )
+            self.USER__READ_ONLY_FIELDS[auth_type.value] = read_only
```
lebouquetin, line 431: I suggest to put an assert here in order to be sure there will be a known value. Otherwise, the dynamic behavior will make it very difficult to investigate in case of error.

backend/tracim_backend/views/core_api/user_controller.py
```diff
@@ -162,6 +163,15 @@
 ]
 
 
+def has_write_rights(field: str, app_config: CFG, request: TracimRequest):
```
lebouquetin, line 166: return type is missing

backend/tracim_backend/views/core_api/user_controller.py
```diff
@@ -162,6 +163,15 @@
...
+    read_only_fields = app_config.USER__READ_ONLY_FIELDS.get(request.candidate_user.auth_type.value)
+    return not (
+        request.current_user.profile.name != "ADMIN"
+        and read_only_fields is not None
+        and field in read_only_fields
+    )
```
lebouquetin, line 172: this logical computation is too difficult to understand.

For something more readable:

if xxx:
    if yyy:
        if zzz:
            return True
return False

doc/backend/env_settings.md
Millefeuille42: I used the output from the command line on top of the file, will add these lines back.

doc/backend/env_settings.md
Millefeuille42: Since this is not limited to current auth_types, the list should be updated every time an auth_type is added. If this is not a problem, I will make an exhaustive list

backend/development.ini.sample
Millefeuille42: Currently, default value is empty. How should I specify this ?

backend/development.ini.sample
Millefeuille42: yes fields are database table's columns. But it is more the form fields you see on the UI when editing your (or another's user) profile. I don't really know how to name them in this case...

frontend/src/component/Account/PersonalData.jsx
Millefeuille42: Noted for the name and the operator (since I saw it first and only in TS, I thought this was exclusive to Typescript), will use it now !

From the usage of it, I seemed that `props.displayAdminInfo` is not really a `isAdmin`, but is only set to `true` if the user is an admin. Perhaps renaming this prop then ?

backend/tracim_backend/config.py
Millefeuille42: The dict parameter is to ease the usage of the said parameter. Since the use we make of it is typical to a `Dict` (Basically we are accessing a value depending on a string).

We could make it a `List` of `List`s, since `List` are used in the parameters, but I think that this is not optimal.

backend/tracim_backend/views/core_api/user_controller.py
Millefeuille42: Will fix this.

backend/development.ini.sample
Millefeuille42: Added a comment explaining why and when this parameter is important. Still don't know how to address blank default values

backend/tracim_backend/views/core_api/user_controller.py
Millefeuille42: resolved

backend/tracim_backend/views/core_api/user_controller.py
Millefeuille42: resolved

doc/backend/env_settings.md
Millefeuille42: resolved

backend/development.ini.sample
Skylsmoi: "fields" is fine. What I'm pointing at is how an admin will be able to know which values fields can take

frontend/src/component/Account/PersonalData.jsx
Skylsmoi: I think renaming the props is a good solution

About the optional chaining, I don't think we have already use it in tracim and it's a recent addition in js.
So you must check that it is properly transpilled by frontend build. If not, it's better not to use it before upgrading Babel

backend/tracim_backend/views/core_api/user_controller.py
```diff
@@ -1500,13 +1518,15 @@ def bind(self, configurator: Configurator) -> None:
...
+            "/users/{user_id:\d+}/workspaces/{workspace_id}/contents/{content_id}/read",
+            # noqa: W605
             request_method="PUT",
         )
         configurator.add_view(self.set_content_as_read, route_name="read_content")
         configurator.add_route(
             "unread_content",
-            "/users/{user_id:\d+}/workspaces/{workspace_id}/contents/{content_id}/unread",  # noqa: W605
+            "/users/{user_id:\d+}/workspaces/{workspace_id}/contents/{content_id}/unread",
+            # noqa: W605
```
lebouquetin, line 1532: pourquoi y-a-t-il ce changement @Millefeuille42 ?

doc/backend/env_settings.md
lebouquetin: @Millefeuille42 the documentation must be explicit and exhaustive

backend/tracim_backend/views/core_api/user_controller.py
Millefeuille42: It has been done when running the lint tools. Not sure of *why* these changes have been made by it though.

backend/development.ini.sample
Millefeuille42: Ok I see, I will make it more clear then

frontend/src/component/Account/PersonalData.jsx
```diff
@@ -78,7 +83,7 @@ export class PersonalData extends React.Component {
               placeholder={props.userPublicName}
               value={state.newPublicName}
               onChange={this.handleChangePublicName}
-              disabled={!shouldAllowEditionFromAuthType}
+              disabled={this.isFieldWritable('public_name')}
```
lebouquetin: disable=isFieldWritable(xxx) ?

I suggest a `isReadOnly()` function which is more straightforward to understand `disabled={this.isReadOnly('public_name')}`

doc/backend/env_settings.md
Millefeuille42: @lebouquetin The parameter is documented in the sample `.ini` file. As where the in-depth configuration documentation is AFAIK. Now from what I understand, it is required to list all parameters depending on the auth_type. I will de so

frontend/src/component/Account/PersonalData.jsx
lebouquetin: Actually, I do not understand: if the field is writable, then `isFieldWritable(xxx)` returns True, then disabled=true which is wrong

Millefeuille42: Indeed I mixed-up the naming and function... I will make it more clear with something like `isReadOnly`

Millefeuille42: Thinking about it. It really checks if it is editable, not just if it is read-only (an admin can write, despite the field being read-only).

I will adjust the naming and return value accordingly.

backend/development.ini.sample
Millefeuille42: resolved in last commit

backend/development.ini.sample
Millefeuille42: resolved in last commit

frontend/src/component/Account/PersonalData.jsx
Millefeuille42: I decided to make it more readable, without the operator. I think we already had this discussion, and the answer was to not use it

backend/tracim_backend/views/core_api/user_controller.py
Millefeuille42: resolved

frontend/src/component/Account/PersonalData.jsx
Millefeuille42: I changed it to `isDisabled`, this makes it more clear about what it returns and checks

doc/backend/env_settings.md
Millefeuille42: Listed relevant auth_types in the table and remove XYZ

backend/tracim_backend/views/core_api/user_controller.py
```diff
@@ -320,6 +330,8 @@ def set_user_email(self, context, request: TracimRequest, hapic_data=None):
         uapi = UserApi(
             current_user=request.current_user, session=request.dbsession, config=app_config  # User
         )
+        if not has_write_rights("email", app_config, request):
+            return HTTPBadRequest()
```
lebouquetin, line 337: You must return some explanation. Error cases must be conform to Tracim error code strategy: You must return a Tracim error code + details. See `error.py` + `class ErrorCode`

backend/tracim_backend/views/core_api/user_controller.py
```diff
@@ -345,6 +357,8 @@ def set_user_username(self, context, request: TracimRequest, hapic_data=None):
         uapi = UserApi(
             current_user=request.current_user, session=request.dbsession, config=app_config  # User
         )
+        if not has_write_rights("username", app_config, request):
+            return HTTPBadRequest()
```
lebouquetin, line 364: You must return some explanation. Error cases must be conform to Tracim error code strategy: You must return a Tracim error code + details. See `error.py` + `class ErrorCode`

backend/tracim_backend/views/core_api/user_controller.py
```diff
@@ -369,6 +383,8 @@ def set_user_password(self, context, request: TracimRequest, hapic_data=None):
         uapi = UserApi(
             current_user=request.current_user, session=request.dbsession, config=app_config  # User
         )
+        if not has_write_rights("password", app_config, request):
+            return HTTPBadRequest()
```
lebouquetin, line 390: You must return some explanation. Error cases must be conform to Tracim error code strategy: You must return a Tracim error code + details. See `error.py` + `class ErrorCode`

lebouquetin:
There are sevearl fixes expected:

- more clear documentation
- better error case management
- readability of some parts of code could/should/may be improved

backend/tracim_backend/views/core_api/user_controller.py
Millefeuille42: @lebouquetin Could you give more explanation about the *how* please ? I can't find usage in the code, all I get is `raise xxx` which triggers a HTTP 500, followed by an "unexpected error" in the frontend.

If I return a `HTTPBadRequest()` with error code + details in the body, frontend still displays a generic `Error while changing username` error.

If providing more details in the code and in the request body, then I know how to do it, but to send more info to the frontend, I not sure of how...

backend/tracim_backend/config.py
lebouquetin: We miss-understood each other.

The ease of use would be to have 4 statick config fields:

```
USER__READ_ONLY_FIELDS__DEFAULT
USER__READ_ONLY_FIELDS__LDAP
USER__READ_ONLY_FIELDS__REMOTE
USER__READ_ONLY_FIELDS__SAML
```

You feel that a dynamic configuration is better. You type the config field like this: `self.USER__READ_ONLY_FIELDS: typing.Dict[str, typing.List[str]] = {}` which does not really help because *string are evil*.

First, the typing should be something like `self.USER__READ_ONLY_FIELDS: typing.Dict[AuthType, typing.List[str]] = {}`

Then, what are the valid values for the strings? There are really no limit? 

I suggest to explicitly assert that identified read-only fields exist (raise an error during config avoid tracim to start and make it easier to be sure the configuration is ok for run)


By the way: auth types are explicitly defined in the `AuthType` enum class;  did you update `class AuthType()` with a new ENUM value SAML? (in `auth.py`)

backend/tracim_backend/views/core_api/user_controller.py
```diff
@@ -320,6 +333,10 @@ def set_user_email(self, context, request: TracimRequest, hapic_data=None):
         uapi = UserApi(
             current_user=request.current_user, session=request.dbsession, config=app_config  # User
         )
+        if not has_write_rights("email", app_config, request):
+            return HTTPBadRequest(
```
lebouquetin, line 337: You should raise a `ReadOnlyFieldException` (which you have to create) + associated error code.

For some example, search `WORKSPACE_NOT_FOUND` in `error.py` file. We can have a talk tomorrow morning

lebouquetin: The way to implement errors in the API is the following one:

```
uapi = UserApi(current_user=request.current_user, session=request.dbsession, config=app_config)
if has_write_rights("email", app_config, request):
    user = uapi.set_email(
        request.candidate_user,
        hapic_data.body.loggedin_user_password,
if not has_write_rights("username", app_config, request):
    [...]
```

and the implementation of the check should be the following one:

```
def has_write_rights(field: str, app_config: CFG, request: TracimRequest) -> bool:
    """
    Check if user can write provided field (is admin or field is writable)
    """
    read_only_fields = app_config.USER__READ_ONLY_FIELDS.get(request.candidate_user.auth_type.value)
    if request.current_user.profile.name != "ADMIN":
        if read_only_fields is not None:
            if field in read_only_fields:
                raise ReadOnlyFieldException(f"You can't modify the field {field}")
    return True
```

The explaination is the following one:

- the function will raise an exception
- the exception would be caught by hapic using the following decorator: `@hapic.handle_exception(ReadOnlyFieldException, HTTPStatus.BAD_REQUEST)`
- hapic will process the exception with associated code and message and builds the right JSON struct for frontend being able to process errors

lebouquetin: Look at the rout signature:

```
    @hapic.with_api_doc(tags=[SWAGGER_TAG__USER_ENDPOINTS])
    @hapic.handle_exception(WrongUserPassword, HTTPStatus.FORBIDDEN)
    @hapic.handle_exception(EmailAlreadyExists, HTTPStatus.BAD_REQUEST)
    @hapic.handle_exception(ExternalAuthUserEmailModificationDisallowed, HTTPStatus.BAD_REQUEST)
    @check_right(has_personal_access)
    @hapic.input_body(SetEmailSchema())
    @hapic.input_path(UserIdPathSchema())
    @hapic.output_body(UserSchema())
    def set_user_email(self, context, request: TracimRequest, hapic_data=None):
        [...]
```

The `WrongUserPassword`, `EmailAlreadyExists`, `ExternalAuthUserEmailModificationDisallowed` errors are catched and transform to HTTP Bad Request with tracim standard JSON error  

You may probably even implement something similar to `has_personal_access`

lebouquetin: The reason for this software design is:

- to keep function as readable as possible
- to manage errors a standard way
- to manage errors out-of-code-path
- to ensure we'll always return valid JSON (every HTTP answer is processed using hapic

Code is easier to read and not less robust

backend/tracim_backend/config.py
```diff
@@ -414,6 +415,21 @@ def _load_global_config(self) -> None:
             cast_func=AuthType,
             do_strip=True,
         )
+        self.USER__READ_ONLY_FIELDS: typing.Dict[str, typing.List[str]] = {}
+        for auth_type in self.AUTH_TYPES:
+            read_only = string_to_unique_item_list(
```
lebouquetin, line 420: `read_only` is a list of strings, right? So you should better name it `readonly_field_list`

backend/tracim_backend/views/core_api/user_controller.py
```diff
@@ -162,6 +163,18 @@
...
 
+def has_write_rights(field: str, app_config: CFG, request: TracimRequest) -> bool:
+    """
+    Check if user can write provided field (is admin or field is writable)
+    """
+    read_only_fields = app_config.USER__READ_ONLY_FIELDS.get(request.candidate_user.auth_type.value)
+    if request.current_user.profile.name != "ADMIN":
+        if read_only_fields is not None:
+            if field in read_only_fields:
+                return False
```
lebouquetin, line 174: raise ReadOnlyFieldException

backend/tracim_backend/views/core_api/user_controller.py
```diff
@@ -162,6 +163,18 @@
 ]
 
 
+def has_write_rights(field: str, app_config: CFG, request: TracimRequest) -> bool:
+    """
+    Check if user can write provided field (is admin or field is writable)
+    """
+    read_only_fields = app_config.USER__READ_ONLY_FIELDS.get(request.candidate_user.auth_type.value)
+    if request.current_user.profile.name != "ADMIN":
```
lebouquetin, line 171: - I suggest not to allow admin to edit readonly fields
- If we allow (let me know if you see a very good reason for this), you *must not* compare with a string `ADMIN`. In this case, you should use `request.current_user.profile != Profile.ADMIN:`

Strings are evil.

doc/backend/env_settings.md
lebouquetin: @Millefeuille42 tracim may be use without a `.ini` file. This may seem like duplication of information but:

- documentation here is targeting all users. This is the reference, the source of truth
- documentation in the `.ini` file is the fastest way to help sysadmins to configure their tracim server.

These are different usages, information may be a bit duplicated but it's required

backend/tracim_backend/config.py
Millefeuille42: Ok I understand better. The reason I was not limiting anything was that I wanted to make it as evolutive as possible (as it is, this part doesn't have to be edited if anything changes). But I understand how it also *is* an issue. I will fix this

Millefeuille42: And yes I added an auth type properly as defined in the `AuthType` enum and the `AuthType` class

backend/tracim_backend/views/core_api/user_controller.py
Millefeuille42: Ok I see. Indeed, my bad for matching with strings and not profile itself.

As for the admins, I'll let @PhilippeAccorsi have a word about this since he was the first to bring this subject.

lebouquetin: @Millefeuille42 there are two types of administrators:

- technical administrators who are in charge of runtime. They access to CLI and have the spirit of "not touching things that may break stuff".
- functionnal administrators. The are in charge of functionnal parameterization of softwares. The use a UI and consider that software can't be break.

Most of the time, things available through graphical UI should not break stuff while CLI opeartions may break stuff.

@PhilippeAccorsi what do you think?

frontend/src/component/Account/PersonalData.jsx
Skylsmoi: I think too, sadly :<

Skylsmoi (approved):
Approving only the frontend changes
Haven't checked the backend ones

backend/tracim_backend/views/core_api/user_controller.py
Millefeuille42: I get your point. I do agree then. I'll still let Philippe answer before doing anything else

backend/tracim_backend/views/core_api/user_controller.py
Millefeuille42: Resolved in last commit

backend/tracim_backend/config.py
Millefeuille42: Resolved in last commit

backend/development.ini.sample
```diff
@@ -376,6 +376,20 @@ api.key = %(basic_setup.api_key)s
...
+# available fields are the following: "public_name", "username", "email" and "password"
+# example:
+# user.profile.read_only_fields.saml=username,email,public_name
+# user.profile.read_only_fields.ldap=username,password,email,public_name
+# By default, read_only_fields for all auth_types are empty
+
+; user.profile.read_only_fields.saml=
+; user.profile.read_only_fields.ldap=
+; user.profile.read_only_fields.internal=
+
```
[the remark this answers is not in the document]
Millefeuille42: Resolved in last commit

backend/tracim_backend/views/core_api/user_controller.py
Millefeuille42: Resolved in last commit. Thanks for the explanation ! It is indeed way better to read / code

backend/tracim_backend/views/core_api/user_controller.py
PhilippeAccorsi: @Millefeuille42 @lebouquetin

PhilippeAccorsi (approved):

backend/tracim_backend/views/core_api/user_controller.py
Millefeuille42: Read only status is only config wise (not "set", it is evaluated depending on config). It still should be editable through SQL and CLI. Additionally, the read-only status is not considered if data is changed on the external system.

Millefeuille42: removed rights for admin to edit read-only fields in last commit

backend/tracim_backend/views/core_api/user_controller.py
```diff
@@ -387,6 +409,10 @@ def set_user_infos(self, context, request: TracimRequest, hapic_data=None):
         Set user info data
         """
         app_config = request.registry.settings["CFG"]  # type: CFG
+        try:
+            check_has_write_rights(UserReadOnlyFields.PUBLIC_NAME, app_config, request)
+        except ReadOnlyFieldException:
```
lebouquetin, line 414: why do you try/except in this case? I would expect using a ` @hapic.handle_exception(ReadOnlyFieldException, HTTPStatus.BAD_REQUEST)`

Is it a miss? If not, then an explanation in a comment is a good idea, I think

Millefeuille42: @lebouquetin Since `public_name` is always sent for this request (even when changing language) it would always throw an error. Instead we make sure the `public_name` doesn't change but we do not prevent other changes. I will put a comment

Millefeuille42: Added the comment

lebouquetin:
I did not test ; we are targetting to test it on branchdev with real shibb configuration

Issue #6258, opened 2023-11-08
✨ Feat: Add an external_id field to user accounts


lebouquetin:
### Feature description

Currently:

When using a remote auth mechanism (HTTP_AUTH, LDAP or SAML), we need to map internal username or email address with remote field.

- this is the only way we can provision new accounts on the fly.
- this may result in duplicated user account if the remote username or email address change
- our username constraints may be different from remote ones, and in this case we can't map user accounts with efficiency

Expected:

**AT LEAST FOR SAML AUTHENTICATION** 

- Add an `external_id` free string field associated to tracim user account. This field should be varchar (1024) in order to allow virtually any id - we don't know the rules of remote user management system.
- fill this field with remote value (use the mapping mechanism for this, the one used for email, display_name, etc)
- when auto-provisionning an account for a given username or email address, check if the user is already existing (with this external_id):
  - if not -> create the new account
  - if the external_id already exist -> update the account with all new values (display_name, username, email_address)

bonus:
- add a documentation with a SQL query allowing sysadmin to identify the case where an external_id appears several times in the database (do not force uniqueness for now, but allow to identify a potential problem

Extra-work:

- create an issue for integration of the external_id in the APIs + admin screens
- create an issue in order to share the mechanism with LDAP & REMOTE HTTP authentication protocols

Fixes:

- [x] Remove obsolete TODOs
- [x] Change display_name to public_name
- [x] Change user_id to username

### How it works

The explanation seems enough for me

### Related PR
https://github.com/tracim/tracim/pull/6232
https://github.com/tracim/tracim/pull/6259

lebouquetin:
@Millefeuille42 there are some explanation about external_id here: https://github.com/tracim/tracim/issues/6209 . Can you check the cases of what is implemented and what is not ? It looks like you missed some fields like external_id, lang, (see: The possible properties must include email, username, display_name, profile, lang and external_id.)

lebouquetin:
@Millefeuille42 you told me there is a downgrade issue with the current state of development. Could you give the error traceback here?

Millefeuille42:
@lebouquetin here is the traceback from the CI

```
self = <alembic.operations.batch.ApplyBatchImpl object at 0x7fd820492d90>, const = UniqueConstraint()


    def drop_constraint(self, const):

        if not const.name:

            raise ValueError("Constraint must have a name")

        try:

            if const.name in self.col_named_constraints:

                col, const = self.col_named_constraints.pop(const.name)

    

[...]

PR #6259, opened 2023-11-08
Feat: add external id field to user accounts

Millefeuille42:
https://github.com/tracim/tracim/issues/6258

## Checkpoints

**For developers**

- [x] Make sure that:

**For code reviewers**

**For testers**


backend/tracim_backend/migration/versions/8a9d929b44b9_add_external_id_to_user.py
```diff
@@ -0,0 +1,29 @@
...
+# revision identifiers, used by Alembic.
+revision = "8a9d929b44b9"
+down_revision = "f4d030ffd43e"
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table("users") as batch_op:
+        batch_op.add_column(sa.Column("external_id", sa.VARCHAR(length=255), nullable=True))
+        batch_op.create_unique_constraint(op.f("uq__users__external_id"), ["external_id"])
```
lebouquetin, line 20: In the issue description (in bonus part) I suggested *(do not force uniqueness for now, but allow to identify a potential problem)*

Are you sure we can have uniqueness + several lines with a NULL value? (for mysql, sqlite and postgresql)? I have a vague memory of some issue with this kind of situation (maybe it's not an issue today)

lebouquetin (changes requested):

backend/tracim_backend/migration/versions/8a9d929b44b9_add_external_id_to_user.py
lebouquetin: according to your answer we could keep uniqueness

backend/tracim_backend/lib/core/user.py
```diff
@@ -695,34 +706,37 @@ def saml_authenticate(
...
                 )
-                transaction.commit()
-                user = self.get_one_by_login(user_id)
+                # transaction.commit()
+                user = self.get_one_by_external_id(external_id)
         if user and user.auth_type not in [auth_type, AuthType.UNKNOWN]:
             raise WrongAuthTypeForUser(
                 'User "{}" auth_type is {} not {}'.format(
-                    user_id, user.auth_type.value, auth_type.value
+                    username, user.auth_type.value, auth_type.value
```
lebouquetin, line 739: Username is not the right discriminant foir this case:

- if you are in the saml context, then the external_id is the right discriminant. In this case, the error must explain that the check is done on external id. Something like this: `'User with external_id "{}" auth_type is {} not {}'`
- if you are in the global login context, the tracim user_id (int) is the right discriminant

backend/tracim_backend/lib/core/user.py
```diff
@@ -958,6 +972,7 @@ def _check_username_correctness(self, username: str) -> bool:
     def update(
         self,
         user: User,
+        external_id: typing.Optional[str] = None,
```
lebouquetin, line 975: I think you should put this new field at the end of parameters if you want to be sure not to break the interface.

Example: if somewhere in the code stuff is written like `obj.update(user, "new name", "new email")` inserting the external_id breaks the behavior

lebouquetin (changes requested):

backend/tracim_backend/migration/versions/8a9d929b44b9_add_external_id_to_user.py
Millefeuille42: I'm not sure that's the issue :/ I know that it is ok for postgres but not the rest.
I suggested to fill the values with the users's actual user_id but we agree that it could conflict with real external_ids . I'm not sure of what to do from now on

Millefeuille42: We should remove uniqueness then ?

backend/tracim_backend/lib/core/user.py
Millefeuille42: Fixed in last commit

backend/tracim_backend/lib/core/user.py
Millefeuille42: Fixed in last commit

backend/tracim_backend/migration/versions/8a9d929b44b9_add_external_id_to_user.py
```diff
@@ -0,0 +1,29 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "8a9d929b44b9"
+down_revision = "f4d030ffd43e"
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    with op.batch_alter_table("users") as batch_op:
+        batch_op.add_column(sa.Column("external_id", sa.VARCHAR(length=255), nullable=True))
```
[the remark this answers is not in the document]
Millefeuille42: Fixed in last commit

Millefeuille42:
@Skylsmoi test issues ***should*** be resolved, I think you can start to test this PR.

To do so:

- Migrate the DB using `alembic -c development.ini upgrade f4d030ffd43e && alembic -c development.ini upgrade 8a9d929b44b9` (or just use a fresh sqlite DB)
- Modify your `settings_saml2.json` to match the updated attribute map format (see below)
- Add a user in the database that has an identical unique column and an identical `external_id` as the user created by logging in with SAML and SSOCircle. Set the `public_name` different data
- Try to log in with SSOCircle, it should work
- There should not be an additional user in the DB, and the `public_name` should have been replaced by the one provided by SSOCircle


NOTE: `user_id` in the JSON is `external_id` in the DB and `public_name` in the JSON is `display_name` in the DB.

Example:
Assuming you have this attribute mapping for SSOCircle:

```json
      "attribute_map": {
        "username": "${UserID}",
        "user_id": "${UserID}",
        "email": "${EmailAddress}",
        "public_name": "${FirstName} ${LastName}"
      },
```

And that the user created by logging in with SSOCircle should have their field to these values

- username: millefeuille
- external_id: millefeuille
- email: xxx.xxx@example.com
- display_name: Millefeuille xxx

Manually create in the DB a user with these values: 

- username: aaaaaa
- external_id: millefeuille
- email: xxx.xxx@example.com
- display_name: aaaaaa bbbbb

When you log in, the user should be modified to these values:

- username: millefeuille
- external_id: millefeuille
- email: xxx.xxx@example.com
- display_name: Millefeuille xxx

lebouquetin:
C'est ok pour moi.

Skylsmoi:
> It's ok for me

You mean the code review or have you tested on branch?

Skylsmoi:
@Millefeuille42 On your branch, before connecting though saml, I tried to create a user manually from admin panel

On a brand new db (tracimcli db init), I can't connect to admin@admin.admin account:

```
  File "/home/come/repo/tracim/backend/tracim_backend/views/core_api/session_controller.py", line 49, in login
    user = uapi.authenticate(
  File "/home/come/repo/tracim/backend/tracim_backend/lib/core/user.py", line 679, in authenticate
    return self._authenticate(
  File "/home/come/repo/tracim/backend/tracim_backend/lib/core/user.py", line 786, in _authenticate
    return self.saml_authenticate(user, login, saml_name, saml_mail)
TypeError: saml_authenticate() missing 1 required positional argument: 'mail'
```

Millefeuille42:
@Skylsmoi Had this during development too but I had it fixed :/, I will take a look

Skylsmoi:
I was up to date with feat/6258__add_external_id_field_to_user_accounts

maybe it was fixed on another branch?

Millefeuille42:
@Skylsmoi I worked when I tested for no known reason (I mean it shouldn't have) :/ Still fixed it, it should work now !

Skylsmoi:
@Millefeuille42 What I did:
1. connect as admin@admin.admin
2. create a user from admin profile in tracim
3. manually set my UserID from SSO Circle (algoo.Skylsmoi-test1) in the field external_id in my database
4. disconnect from tracim
5. connect through SSO Circle SAML auth

It did not create a new user. Which is good.
It did update the display_name. Which is good.
It did NOT update neither the username nor the email. Which I think is wrong

Millefeuille42:
@Skylsmoi Indeed there is an issue with username and email (not with public_name)

It is weird because the update method supposedly updates the user fields, but they are not updated only for these fields :/ 


@lebouquetin @buxx Anything about this ?

Millefeuille42:
Edit: Tried a lot of things. Pretty much sure alembic "is" the culprit here. But I can't get my finger on what is actually wrong :/

buxx:
## Summary of flush error fix

**Context** : When saml authentication was processed by tracim, some `User` update was not done (display name, username). A session flush error happened because the session was already flushed (saml authentication code is called regularly through Pyramid `Request.authenticated_userid` property).

**Fix** : Add a `Session.no_autoflush` over the saml authenticate function to prevent autoflush.

**Fixme** : To architecturally fix the problem, we must move the `User` update (display name, username) in other place. The currently chosen place is `SAMLSecurityPolicy._acs`

@lebouquetin

Issue #6366, opened 2024-02-20
✨ Feat: Add support of Redis cluster


Millefeuille42:
### Feature description

Add support of Redis cluster for tracim. Currently it is required to connect to a cluster through a proxy, which is far from optimal.

### How it works

- Upgrade dependency `redis` to `redis==4.60`.
`4.6.0` because it's the latest version that is designed to support python 3.7, however this will not be a problem anymore when #6360 will be resolved.

- Upgrade dependency `rq` to `rq>=1.15.0`
Because it's the first version that explicitly needs `redis>=4.0.0`, while this doesn't seem really necessary, it would be better to upgrade `rq` too, to prevent any future issue.

Millefeuille42:
Turns out upgrading `rq` causes at leat an error during the tests.
It's a quite weird error that is poorly documented on internet.

> [rq enqueue function: fails with TypeError:function_xxx() takes 2 positional arguments but 3 were given](https://stackoverflow.com/questions/45890670/rq-enqueue-function-fails-with-typeerrorfunction-xxx-takes-2-positional-argu)

Considering that I'm not sure about how to solve this, I'll leave it as it is.
This leaves us with two options:
- Try to solve the issue
- Do not upgrade `rq` and leave it for "upgrading dependencies" related developments

Millefeuille42:
[lebouquetin](https://github.com/lebouquetin) commented [Feb 21, 2024](https://github.com/tracim/tracim/pull/6367#issuecomment-1956366757)

@Millefeuille42 we need to support redis cluster for sessions. Async tasks (the target of RQ) do not require to support redis cluster . We deploy full tracim nodes (with rq + local redis) in scalable architecture, so rq can be used with classic / standalone redis


[Millefeuille42](https://github.com/Millefeuille42) commented [Feb 21, 2024](https://github.com/tracim/tracim/pull/6367#issuecomment-1956372646) •

@lebouquetin I hear that, but since rq depends on the redis-py driver we are using, we need to make sure it's version supports the version of the driver. There is no signs this doesn't support it, but we need to be cautious. And not just rely on the "queuing" != "session" assomption, otherwise this could bring a lot of issues in the future if we are not cautious about that.

lebouquetin:
@Millefeuille42 I do not understand how changing only dependencies can activate redis cluster support.

What is done with redis proxy for example, is to run the proxy with the configuration of all redis nodes. There is nothing in tracim configuration which allow to setup several redis nodes

[...]

PR #6367, opened 2024-02-20
feat: add support of redis cluster

Millefeuille42:
#6366

## Checkpoints

**For developers**

- [x] Make sure that:

**For code reviewers**

**For testers**


buxx (approved):

Millefeuille42:
Turns out upgrading `rq` causes at leat an error during the tests.
It's a quite weird error that is poorly documented on internet.

> [rq enqueue function: fails with TypeError:function_xxx() takes 2 positional arguments but 3 were given](https://stackoverflow.com/questions/45890670/rq-enqueue-function-fails-with-typeerrorfunction-xxx-takes-2-positional-argu)

Considering that I'm not sure about how to solve this, I'll leave it as it is.
This leaves us with two options:
- Try to solve the issue
- Do not upgrade `rq` and leave it for "upgrading dependencies" related developments

buxx:
I suggest "Do not upgrade rq". Unless if its required for https://github.com/tracim/tracim/issues/6366.

Millefeuille42:
@buxx It is not required per se, and our current version is *supposed* to be compatible with latest redis-py. I do think we should go with that too.

lebouquetin:
@Millefeuille42 we need to support redis cluster for sessions. Async tasks (the target of RQ) do not require to support redis cluster . We deploy full tracim nodes (with rq + local redis) in scalable architecture, so rq can be used with classic / standalone redis

Millefeuille42:
@lebouquetin I hear that, but since rq depends on the `redis-py` driver we are using, we need to make sure it's version supports the version of the driver. There is no signs this doesn't support it, but we need to be cautious. And just no rely on the "queuing" != "session" assomption otherwise this could bring a lot of issues in the future if we are not cautious about that.

lebouquetin:
By the way, this kind of comments should better be in the issue (not in the PR) : if we come back on reasons for decisions, we will go to the issues, the PR will not be investigated.

Millefeuille42:
Still in progress due to the fact that beaker (and redis-py version) is the main culprit. see #6366 for more details.

On hold for now

backend/setup.py
```diff
@@ -39,7 +39,7 @@
     # mail-notifier
     "mako",
     "lxml>=4.8.0",
-    "redis>=3.5.3",
```
PhilippeAccorsi, line 42: Question: do we have same version like at present? or 4.6.0 is not the latest but necessary version.

backend/tracim_backend/config.py
```diff
@@ -1154,6 +1155,12 @@ def _check_global_config_validity(self) -> None:
             self.check_directory_path_param(
                 "SESSION__DATA_DIR", self.SESSION__DATA_DIR, writable=True
             )
+        elif self.SESSION__TYPE in ["ext:rediscluster"]:
+            self.check_mandatory_param(
+                "SESSION__URLS",
+                self.SESSION__URLS,
+                when_str="if session type is {}".format(self.SESSION__TYPE),
+            )
```
PhilippeAccorsi, line 1163: can we check SESSION_URL is empty (if it is mandatory when SESSION_URLS is set)

Millefeuille42: SESSION_URL doesn't have to be empty. I juste mimiced actual behavior and description.

backend/setup.py
Millefeuille42: 4.6.0 is the latest update of the highest major version that supports python 3.7, This will be updated once we drop python 3.7 support

buxx (approved):

Millefeuille42:
For the tester:
1. Change `development.ini` to have these parameters
```
## "Redis cluster" storage config sample:
session.type = ext:rediscluster
session.urls = redis://<node-1>:<port>,redis://<node-2>:<port>...
```
For the urls, use an available redis cluster, provide as much urls as desired, they are considered to be entrypoints; if all the entrypoints fails tracim has no access to the cluster.

2. Connect to tracim
3. It should work as usual

Skylsmoi:
Issue during test:
I couldn't make it to work. I got rejected because of auth, I couldn't find a way to give the redis server password.
I tried:
- giving it in url in development.ini session.urls
- to give it by env var REDISCLI_AUTH (found it in redis doc)
- to tweak the code in beaker/ext/redisclusternm.py from my venv

Millefeuille42:
@Skylsmoi This should work now, specify the password with this format `redis://:[password]@[host]:[port]`

The full URL format is `redis://[username]:[password]@[host]:[port]` but since we use legacy auth in the cluster we are running, username is not necessary

Millefeuille42:
Note: the change is on the beaker dep, which is currently fetched from github repo algoo/beaker

To update (when in virtualenv):
- `pip uninstall beaker`
- `pip install -r requirements.txt`

Millefeuille42:
New version (1.13.0) including our fix is live on pypi https://pypi.org/project/Beaker/ and github https://github.com/bbangert/beaker/releases/tag/1.13.0

There is a few extras commits ahead of our fork, most commits concerns doc or tests.

Issue #6355, opened 2024-02-14
🐛 Bug: When using ldap,internal auth_type, internal auth doesn't work


Skylsmoi:
### Tracim version

4.7.1

### Issue description

When using ldap,internal auth_type, internal auth doesn't work.

Using a local tracim account doesn't work for login.

### How to reproduce

1. Have a LDAP connected tracim with "auth_type: ldap, internal" in development.ini
2. create an account using tracimcli
3. try to connect to that account

Millefeuille42:
Turns out this behaviour only happens for `UNKNOWN` `auth_type` due to how the current flow is implemented:

1. Iterate over the allowed `auth_types`
2. Go to LDAP auth function
3. LDAP auth flow is continued if user has `auth_type` of `LDAP` or `UNKNOWN`
4. LDAP auth flow raises `WrongLDAPCredentials` exception
5. global auth flow catches this exception as a "fatal" exception (by that I mean user can't login)

We should change the exception raised on step 4 if the user's `auth_type` is `UNKNOWN` to `WrongAuthTypeForUser`, then the auth flow will continue with the other allowed `auth_types`

All of the provided code is from `backend/tracim_backend/lib/core/user.py`
Step 1 and 5 (line 756): 
```python
        for auth_type in self._config.AUTH_TYPES:
            if auth_type == AuthType.SAML:
                continue
            try:
                return self._authenticate(
                    login=login,
                    password=password,
                    ldap_connector=ldap_connector,
                    auth_type=auth_type,
                )
            except AuthenticationFailed as exc:
                raise exc
            except WrongAuthTypeForUser:
                pass
```

Step 2 (line ):
```
        try:
            if auth_type == AuthType.LDAP:
                if ldap_connector:
                    return self._ldap_authenticate(user, login, password, ldap_connector)
                raise MissingLDAPConnector()
            elif auth_type == AuthType.INTERNAL:
                return self._internal_db_authenticate(user, login, password)
            else:
                raise UnknownAuthType()
        except (
            EmailRequired,
            WrongUserPassword,
            WrongLDAPCredentials,
            UserDoesNotExist,
            UserAuthenticatedIsDeleted,
            UserAuthenticatedIsNotActive,
            TracimValidationFailed,
        ) as exc:
            raise AuthenticationFailed('User "{}" authentication failed'.format(login)) from exc
```

Step 3 (line ):
```
        if user and user.auth_type not in [auth_type, AuthType.UNKNOWN]:
[...]

PR #6369, opened 2024-02-21
fix(backend): LDAP auth flow doesn't compromise auth for "internal as unknown" users anymore

Millefeuille42:
#6355 

## Checkpoints

**For developers**

- [x] Make sure that:

**For code reviewers**

**For testers**

### DIFF
--- a/backend/tracim_backend/lib/core/user.py
+++ b/backend/tracim_backend/lib/core/user.py
@@ -561,6 +561,9 @@ need to be in every workspace you include."
         data = ldap_connector.authenticate(login, password)
         if not data:
             if user.auth_type == AuthType.UNKNOWN:
+                # INFO - M.L - 2024-02-22 - In the event the user is of "UNKNOWN" auth type,
+                #  we can't be sure if the user is not allowed to authenticate with the other
+                #  auth types. This permits to try to authenticate with the other auth types.
                 raise WrongAuthTypeForUser(
                     'User "{}" auth_type is {} and cannot authenticate with {}'.format(
                         login, user.auth_type.value, auth_type.value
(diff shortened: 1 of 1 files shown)


buxx (approved):

backend/tracim_backend/lib/core/user.py
```diff
@@ -560,6 +560,12 @@ def _ldap_authenticate(
         # INFO - G.M - 2018-11-22 - LDAP Auth
         data = ldap_connector.authenticate(login, password)
         if not data:
+            if user.auth_type == AuthType.UNKNOWN:
```
lebouquetin, line 563: I suggest you put a comment with reference to the issue. Otherwise, a maintainer will not understand why this special case is here

Millefeuille42: done

Millefeuille42:
How to test:

In `.../backend`

1. Remove current venv
2. Recreate it and activate it `python3 -m venv env; . env/bin/activate`
3a. Install dependencies `pip install -r requirements.txt`
3b. Install other dependencies if required (like postgres)
4. Run tracim with SAML enabled ([see documentation](https://github.com/tracim/tracim/blob/develop/doc/backend/setting.md#saml-authentication))
5. If it starts it means it works, but it is recommended to test the full auth flow too

Skylsmoi:
tested on branch

Only issue is that when user already exists in db and is logged through ldap (because it has the proper id:password in local), the auth_type of the user isn't updated and is still UNKNOWN.
The issue doesn't exists for users created through the ldap auth process, these users have auth_type LDAP.

Since this process isn't a usual one and probably should never exists because we should not create a user locally with its password from ldap server, I consider this PR to be valid

Skylsmoi:
Issue created related to my last comment: https://github.com/tracim/tracim/issues/6380

Issue #6381, opened 2024-02-27
✨ Feat: force tracimcli execution to be executed from tracim folder


PhilippeAccorsi:
### Feature description

This fix is necessary for example to prevent message "fatal: not a git......." visible if you execute tracimcli command directly from the root of the docker container.

Some of admin-sys think that command have some error after seen this kind of message. Ex:
```
root@*******:/# tracimcli user update -c /etc/tracim/development.ini -u philippeaccorsi -l {user_email}
```
Result
```
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
2024-02-27 10:52:43 WARNI [tracim:47][MainThread] [CFG] DEPOT_STORAGE_DIR parameter is deprecated. 
2024-02-27 10:52:43 WARNI [tracim:47][MainThread] [CFG] DEPOT_STORAGE_NAME parameter is deprecated. 
2024-02-27 10:52:43 WARNI [tracim:47][MainThread] [CFG] EMAIL__NOTIFICATION__SMTP__USE_IMPLICIT_SSL parameter is deprecated. Use EMAIL__NOTIFICATION__SMTP__CONNECT_METHOD parameter instead.
2024-02-27 10:52:43 WARNI [tracim:47][MainThread] [PluginManager] Cannot find plugin entry point 'register_tracim_plugin' in 'montracim' plugin
```

If we execute same command from `/tracim` folder:
```
release_04.08.00
2024-02-27 10:55:00 WARNI [tracim:47][MainThread] [CFG] DEPOT_STORAGE_DIR parameter is deprecated. 
2024-02-27 10:55:00 WARNI [tracim:47][MainThread] [CFG] DEPOT_STORAGE_NAME parameter is deprecated. 
2024-02-27 10:55:00 WARNI [tracim:47][MainThread] [CFG] EMAIL__NOTIFICATION__SMTP__USE_IMPLICIT_SSL parameter is deprecated. Use EMAIL__NOTIFICATION__SMTP__CONNECT_METHOD parameter instead.
2024-02-27 10:55:00 WARNI [tracim:47][MainThread] [PluginManager] Cannot find plugin entry point 'register_tracim_plugin' in 'montracim' plugin
User updated
```

### How it works

User can execute tracimcli command from anywhere just if command is executed automatically from tracim repository.

Or

[...]

Issue #6312, opened 2023-12-05
✨ Feat: Allow tracimcli to be used in docker context without having to specify the path to config file


Skylsmoi:
### Feature description

Allow tracimcli to be used in docker context without having to specify the path to config file.

### How it works

in `backend/tracim_backend/lib/utils/utils.py` , the value of `DEFAULT_TRACIM_CONFIG_FILE` should come from an env variable

### Related PR

#6418

PhilippeAccorsi:
This fix is necessary for example to prevent message "fatal: not a git......." visible if you execute tracimcli command directly from the root of the docker container.

Some of admin-sys think that command have some error after seen this kind of message. Ex:
```
root@*******:/# tracimcli user update -c /etc/tracim/development.ini -u philippeaccorsi -l {user_email}
```
Result
```
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
2024-02-27 10:52:43 WARNI [tracim:47][MainThread] [CFG] DEPOT_STORAGE_DIR parameter is deprecated. 
2024-02-27 10:52:43 WARNI [tracim:47][MainThread] [CFG] DEPOT_STORAGE_NAME parameter is deprecated. 
2024-02-27 10:52:43 WARNI [tracim:47][MainThread] [CFG] EMAIL__NOTIFICATION__SMTP__USE_IMPLICIT_SSL parameter is deprecated. Use EMAIL__NOTIFICATION__SMTP__CONNECT_METHOD parameter instead.
2024-02-27 10:52:43 WARNI [tracim:47][MainThread] [PluginManager] Cannot find plugin entry point 'register_tracim_plugin' in 'montracim' plugin
```

If we execute same command from `/tracim` folder:
```
release_04.08.00
2024-02-27 10:55:00 WARNI [tracim:47][MainThread] [CFG] DEPOT_STORAGE_DIR parameter is deprecated. 
2024-02-27 10:55:00 WARNI [tracim:47][MainThread] [CFG] DEPOT_STORAGE_NAME parameter is deprecated. 
2024-02-27 10:55:00 WARNI [tracim:47][MainThread] [CFG] EMAIL__NOTIFICATION__SMTP__USE_IMPLICIT_SSL parameter is deprecated. Use EMAIL__NOTIFICATION__SMTP__CONNECT_METHOD parameter instead.
[...]

PR #6417, opened 2024-03-26
feat(back): change directory before executing tracimcli command

floreal15:
https://github.com/tracim/tracim/issues/6381

## Checkpoints

**For developers**

- [ ] Make sure that:
- [ ] Original authors of the features included in a multi-feature branch (maintenance fixes -> develop, security fixes -> develop, …) should be part of the reviewers, especially if you encountered merge conflicts.

**For code reviewers**

**For testers**

### DIFF
--- a/backend/tracim_backend/command/__init__.py
+++ b/backend/tracim_backend/command/__init__.py
@@ -5,6 +5,7 @@ from cliff.app import App
 from cliff.command import Command
 from cliff.commandmanager import CommandManager
 import logging
+import os
 from pyramid.paster import bootstrap
 from pyramid.paster import setup_logging
 import sys
@@ -56,6 +57,7 @@ class AppContextCommand(Command):
     def take_action(self, parsed_args: argparse.Namespace) -> None:
         try:
             super(AppContextCommand, self).take_action(parsed_args)
+            self.changeDirectory()
             self._setup_logging(parsed_args)
             if self.auto_setup_context:
                 with bootstrap(parsed_args.config_file) as app_context:
@@ -97,3 +99,15 @@ class AppContextCommand(Command):
             default=False,
         )
         return parser
+
+    def changeDirectory(self) -> None:
+        filePath = os.path.dirname(__file__)
+        listDirectory = filePath.split("/")
+        if len(listDirectory) > 3 and listDirectory[-3] == "backend":
+            os.chdir(os.path.dirname(__file__))
+            os.chdir("../..")
+        else:
+            try:
+                os.chdir("/etc/tracim")
+            except:
+                pass


backend/tracim_backend/command/__init__.py
```diff
@@ -97,3 +99,14 @@ def get_parser(self, prog_name: str) -> argparse.ArgumentParser:
             default=False,
         )
         return parser
+
+    def changeDirectory(self) -> None:
+        directoryPath = os.path.dirname(__file__)
+        if len(directoryPath) > 30 and directoryPath[-30:] == "backend/tracim_backend/command":
```
PhilippeAccorsi, line 105: What is the name of the file used ? If I use prod.ini, is it still working ?

PhilippeAccorsi: and/or where we can set config file name ?

backend/tracim_backend/command/__init__.py
```diff
@@ -97,3 +99,14 @@ def get_parser(self, prog_name: str) -> argparse.ArgumentParser:
...
+            os.chdir(os.path.dirname(__file__))
+            os.chdir("../..")
+        else:
+            try:
+                os.chdir("/etc/tracim")
```
PhilippeAccorsi, line 110: this is the folder where find config file, but for example, git is not present in this folder (git used when execute tracimcli command to show actual version of tracim).

PhilippeAccorsi (commented):
Question :)

backend/tracim_backend/command/__init__.py
floreal15: ok, so in with folder should we be to execute a tracimcli command on docker ?

Millefeuille42:
To be fair, I don't think there should be warnings about being "not in a git repository".
Changing the `cwd` to the **source code** is not a solution IMHO.

Next thing we know tracimcli could be installed as a system/local program and then used outside of the "source code" context. Even in the source code, there is no guarantee that a `.git` folder might be present.

backend/tracim_backend/command/__init__.py
floreal15: the __file__ is the path to the file executed, backend/tracim_backend/command/__init__.py here, the config file name isn't link to that

floreal15: the config file name can be set with the -c option ( in that case a relative path from another place that the on we will move to execute tracimcli will not work)
or use whet we be developed here : https://github.com/tracim/tracim/issues/6312

floreal15:
> there is no guarantee that a .git folder might be present.
tracimcli will return an error and won't work without it, I think yes there is a guarantee else it won't work

Millefeuille42:
@floreal15 unsure of how the `.git` folder is used for the cli ? 
For all I know it's used to set the `BUILD_VERSION` settings which is not used in tracimcli.

The whole point of this issue was to prevent irrelevant warning about git, this error is solely noise when using tracimcli and will not affect behavior of tracimcli. So I don't think hacks and risky path gymnastics are a solution for this.

floreal15:
Yes, I agree that removing useless warning should be enough

buxx:
I entirely agree with @Millefeuille42 , command must be agnostic about .git or source code place. Probably this development should be done in pair programming with experienced developer ?

backend/tracim_backend/command/__init__.py
PhilippeAccorsi: ok

backend/tracim_backend/command/__init__.py
PhilippeAccorsi: I think it is in `/tracim` or `/tracim/backend`

PhilippeAccorsi:
This can be a good idea to just have something like `tracimcli --version` and result is `tracim version with commit_id` if git is available or something from `setup.cfg` if git not available. Or more simple, all the time the value of `version` available in setup.cfg.

@floreal15 @Millefeuille42 @buxx

Skylsmoi (approved):

Millefeuille42 (approved):

Issue #6485, opened 2024-05-06
Prepare release 4.9.1


Millefeuille42:
- [x] update backend version (backend/setup.cfg)
- [x] update frontend version (frontend/src/version.json)
- [x] update changelog (CHANGELOG.md)

PR #6486, opened 2024-05-06
Prepare new release 4.9.1

Millefeuille42:
#6485


Millefeuille42:
Since there is no change since latest build, see https://github.com/tracim/tracim/pull/6484/files for the changelog edits

Issue #6398, opened 2024-03-05
🐛 Bug: Mention autocomplete can shows deactivated members of a space


Skylsmoi:
**FIRST ACTION IS TO UPDATE DESCRIPTIOn AND TITLE OF THE ISSUE**

### Tracim version

4.8.0

### Issue description

When creating a mention, the autocomplete popin sometimes show deactivated users.

It appears endpoint GET /api/users/234/workspaces/all/settings returns deactivated users, it should do so when passed a path parameter and the default behavior should be to do not.

### How to reproduce

The issue happens intermittently.

1. Open a space
2. Open a content of the space
3. In the comment area, type @
4. The autocomplete popin sometimes shows deactivated users

### Related PR

https://github.com/tracim/tracim/pull/6501

floreal15:
At the current state (tracim 4.8.0) when using a mention, tracim will use either know member list ( for the recent activities and dashboard) or the current workspace member list to find the list of available user for mention,
removing the deactivate user from know members will result in them not being in the list of user we could mention but also their name won't be display on content they have create before because they will be unknown by the front end;
I see few possibilities here:
- do not show the deactivate user, even what they did before being deactivated will appear as create by unknown.
- separate know member list in tow list in front end, one to store user that could be mentioned and  another to store the full list of user to get the full list of user to be able to show who create a content even if deactivate
- know member list return also if each user is deactivated or not and the front filter by depending if deactivate user should be show or not

lebouquetin:
@Millefeuille42 could you comment here what solution you are implementing? In last comment, there are 3 options, the first thing to do is to comment here what you plan to do.

lebouquetin:
I agree this is the best solution :+1:

PR #6501, opened 2024-05-29
fix: remove inactive users from autocomplete in the recent activities page

Millefeuille42:
https://github.com/tracim/tracim/issues/6398

## Checkpoints

**For developers**

- [x] Make sure that:

**For code reviewers**

**For testers**

### DIFF
--- a/backend/tracim_backend/views/core_api/schemas.py
+++ b/backend/tracim_backend/views/core_api/schemas.py
@@ -408,13 +408,6 @@ class UserDigestSchema(marshmallow.Schema):
         example="My-Power_User99", required=False, default=None, allow_none=True
     )
     workspace_ids = marshmallow.fields.List(marshmallow.fields.Int(example=3))
-
-
-class KnownMembersUserDigestSchema(UserDigestSchema):
-    """
-    UserDigestSchema with an "is_active" boolean added
-    """
-
     is_active = marshmallow.fields.Bool()
 
 
--- a/backend/tracim_backend/views/core_api/schemas.py
+++ b/backend/tracim_backend/views/core_api/schemas.py
@@ -410,6 +410,14 @@ class UserDigestSchema(marshmallow.Schema):
     workspace_ids = marshmallow.fields.List(marshmallow.fields.Int(example=3))
 
 
+class KnownMembersUserDigestSchema(UserDigestSchema):
+    """
+    UserDigestSchema with an "is_active" boolean added
+    """
+
+    is_active = marshmallow.fields.Bool()
+
+
 class AppCustomActionsDigestSchema(marshmallow.Schema):
     icon_text = marshmallow.fields.String(
         required=True,
--- a/backend/tracim_backend/lib/core/user.py
+++ b/backend/tracim_backend/lib/core/user.py
@@ -248,7 +248,7 @@ class UserApi(object):
 
     def get_users_in_common_with_user_workspace(
         self, user_id: int
-    ) -> List[Tuple[int, str, str, int]]:
+    ) -> List[Tuple[int, str, str, int, bool]]:
         """
         Returns a list of found user data in common with the provided user as a tuple
         :param user_id: id of the user to get data from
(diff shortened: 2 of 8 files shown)


backend/tracim_backend/views/core_api/schemas.py
```diff
@@ -410,6 +410,14 @@ class UserDigestSchema(marshmallow.Schema):
     workspace_ids = marshmallow.fields.List(marshmallow.fields.Int(example=3))
 
 
+class KnownMembersUserDigestSchema(UserDigestSchema):
+    """
+    UserDigestSchema with an "is_active" boolean added
+    """
+
+    is_active = marshmallow.fields.Bool()
```
lebouquetin, line 418: Why not complete the UserDigestSchema with the `is_active` field? This would lead to less code to maintain (so less bug to fix)... or there is a good reason to split into 2 models?

lebouquetin: @Millefeuille42

Millefeuille42: This is to prevent modifying routes that use this Schema. Modifying it would either:
- Need to edit all the routes using this Schema to provide the value for `is_active`
- Make it optional and have inconsistent results between all routes using this endpoint

While this generates more "noise", this permits to 
1. Contain the changes in the scope of the issue (do not engage in a refactor)
2. Have consistent, clear and defined response schemas on endpoints (and thus in the swagger docs)

Skylsmoi: I'm assuming this would make every endpoint already using the UserDigestSchema to now return a new property "is_active: true" which would not be used
I don't know if it's an issue but we might expect changes in unit tests and api documentation

lebouquetin: > Require us to edit all the routes using this Schema to provide the value for is_active

This is straight without any dev work, I think.

> Require us to make it optional and have inconsistent results between all routes using this endpoint

Why do you want to make it optional ?

> using the UserDigestSchema to now return a new property "is_active: true" which would not be used

Of course this is not used (it is not available;)

Doing like this increase complexity. I can understand "you dont want to break anything; this is the reason for unit tests existance".

lebouquetin: I understand the point "do not break existing code". I'm pretty confident we can do it without to break anything. @Millefeuille42 let's try to add the property to UserDigestSchema and we talk about what it breaks

Millefeuille42: done

Skylsmoi:
tested on branch
not validated, it is missing the is_active information in TLM user.modified
That TLM should deactivate the user in redux

Millefeuille42:
@Skylsmoi after investigation, the TLM **does** deactivate the user in Redux. The issue is that the TinyMCE editor is not updated when the known member list changes.

The change IS propagated to `frontend_lib/src/component/TinyEditor/TinyEditor.jsx` but the `<Editor>` is not rendered, at least it's what I think.

The `userList` present in the callback function we pass to TinyMCE's `init.setup` field is set to the state it was when the component was first rendered. Meaning that it is not updated until we unmount and re-mount the editor. This applies to any of the props used in `init.setup` actually (new user joining the workspace for example)

So I think this is out of the scope of this issue.

Millefeuille42:
Just opened the editor and tried to mention, both lists are matching (top one is the one `TinyEditor.jsx` gets, the other one is the one in TinyMCE's `init.setup`) (look at the `isActive` field)
image

If I change the user status without closing the editor, top list is up to date (confirming that the TLM actually works), but bottom list is not. (look at the `isActive` field)
image

Millefeuille42:
And ofc, If I close the editor and re-open it, the mention list is now up to date

buxx (approved):

Skylsmoi:
as seen together, I agree that the TLM is handled and update redux

The known limitation is when the user is deactivated while we are editing something in tinymce, tinymce will still show the deactivated user until tinymce is closed and reopened.

Skylsmoi (approved):

PR #6505, opened 2024-05-30
Fix/6502  fix horizontal scrollbar in app kanban

Skylsmoi:
This branch has been merged into maintenance/4.9
It will be in release 4.9.2

Now merging it into develop


Skylsmoi:
This is not the right process. Closing the PR

Issue #6707, opened 2025-03-24
✨ Feat: Enable Elasticsearch with HTTPS and Basic Auth authorization


floreal15:
### Feature description

We should allow to connect tracim to an Elastic search using HTTPS or a basic auth authentication

### How it works

create parameters : 
search.elasticsearch.ssl = Boolean to enable HTTPS connection
search.elasticsearch.auth.username = String to set the connection username, if set enable the basic auth connection
search.elasticsearch.auth.password = string, basic auth password

PR #6717, opened 2025-03-26
feat(back): enable elastic search https and basic auth connection

floreal15:
https://github.com/tracim/tracim/issues/6707

## Checkpoints

**For developers**

- [x] Make sure that:

**For code reviewers**

**For testers**


backend/development.ini.sample
```diff
@@ -857,6 +857,11 @@ caldav.radicale.headers.Access-Control-Expose-Headers = Etag
 # to connect to a cluster, it is possible to provided multiple hosts/ports as a comma separated list
 ; search.elasticsearch.host = localhost,
 ; search.elasticsearch.port = 9200,
+# elasticsearch security configuration
+; search.elasticsearch.ssl = False
+; search.elasticsearch.ssl.ca_certs =
+; search.elasticsearch.auth.username =
+; search.elasticsearch.auth.password =
```
PhilippeAccorsi: Can we put this 5 lines just before example ?

I think `auth` can be naming `basic_auth`
What kind of file do we need to put in `search.elasticsearch.ssl.ca_certs` ? Path with file or just filename ?

Can you add also an example in example area, to present which kind of value are accepted ?

PhilippeAccorsi: maybe replace `search.elasticsearch.ssl` by `search.elasticsearch.ssl.activated` (like we have for some parameters with value True or False).

Is it possible to use `search.elasticsearch.ssl=False` (or true) without auth username and password ?

PhilippeAccorsi: What is going on if I have a cluster of elasticsearch ? Each node receive same auth or just one in the list of node ?

PhilippeAccorsi (commented):
some questions :)

backend/development.ini.sample
lebouquetin: > Is it possible to use search.elasticsearch.ssl=False (or true) without auth username and password ?

It's required: these are two separate features:
- use basic_auth or no auth
- use a SSL based connection

@floreal15 please add a thumb up if it is working like this.

Also I like the renaming proposals `ssl.activated` and `basic_auth` instead of `auth`

For the certs file path, we should have something explicit about the expected content of the file. Is it a SSL chain (cert + intermediates ?) a cert only? Also giving an example with a full path should be the right way to help users

lebouquetin: > What is going on if I have a cluster of elasticsearch ? Each node receive same auth or just one in the list of node ?

A line of documentation should explain this (expected behavior: apply the strategy to all configured nodes - @floreal15 thumb up if this is what you implemented)

lebouquetin: @floreal15 please also report all behavior decided here in the issue body. This way, future maintainance will be easier for developers (they will have the full specification without to serarch in merge requests discussions)

floreal15: I rename the config parameter and improve the comments

lebouquetin:
@floreal15 I'm about to validate this PR. Before that I would like to be sure you tested it with the client environment in following cases:

- upload a file
- search for the file
- execute the tracimcli command that re-index the entire search index.

If you did not, could you do these tests and add a thumb to my comment? Then we'll merge it

floreal15:
yes, I tested those cases

lebouquetin:
So lets merge it. Good job :+1:

PR #6742, opened 2025-04-28
Feat/6744  merge revision in timeline

Skylsmoi:
https://github.com/tracim/tracim/issues/6744

## Checkpoints

**For developers**

- [x] Make sure that:

**For code reviewers**

**For testers**


floreal15 (approved):

floreal15:
1-When a content get the revision merged, it's not possible to see the other revision, the link doesn't work anymore

2-When you have a file with a lot a revision (a kanban for example), you will have the show more button, but if there is no comment in the list of following revision to be shown, the revision will be grouped with the other (which is good) but if there are hundreds of revisions before the next comment, I will have to click to show more dozens of times.
Maybe the show more should increase the timeline element (and continue to fetch element until the timeline has grown in case the revision are all grouped

3-There is a bug (doesn't seem to be a regression), when a content has a lot of revision, if you click on show more revision a few times in a short time, the order of some revision is wrong, the first few revision just under the show more button are wrong and get different css (no hover)

The first point is mandatory fix before merge, the second is to think and the 3 is not a regression so it's not a problem to keep

lebouquetin:
@Skylsmoi could you implement the updates I suggested in thie issue? Also, could you add up-to-date screenshots before/after when implemented?

- [x] change wording _from / to_ to _between xx / yy_
- [x] show only one date when revisions are the same day
- [x] add up-to-date screenshots in a new comment in the issue

Skylsmoi:
I will yes

Skylsmoi:
> 1-When a content get the revision merged, it's not possible to see the other revision, the link doesn't work anymore

Will fix

lebouquetin:
@Skylsmoi I think it will be easier to maintain translations with templated strings.

Instead of 

```
  const titleHtml = firstRevisionDate === lastRevisionDate
    ? lastRevisionDate
    : [
      props.t('Between'),
      firstRevisionDate,
      props.t('and'),
      lastRevisionDate
    ].join(' ')
```

I suggest something like that:

```
  const titleHtml = firstRevisionDate === lastRevisionDate
    ? lastRevisionDate
    : `Between ${firstRevisionDate} and ${lastRevisionDate}`
```

Skylsmoi:
We cannot use template literals for the translation function.
Using 
```jsx
props.t(`Between ${firstRevisionDate} and ${lastRevisionDate}`)
```
wont work with the generation of the translation files.

We would have to write:
```jsx
props.t(
  'Between {{firstRevisionDate}} and {{lastRevisionDate}}',
  { firstRevisionDate: firstRevisionDate, lastRevisionDate: lastRevisionDate }
)
```

floreal15:
I got a crash in js and the app crashed when I try to open a kanban with a lot of revision : 
`Uncaught RangeError: Invalid time value`

floreal15:
tested on branch

Issue #6642, opened 2024-12-23
🐛 Bug: Edition in Kanban can be slow


Skylsmoi:
### Tracim version

4.11+

### Issue description

When a kanban has many cards (~150+), editing the kanban by any way (edit card, move card, edit column...) is slow.

This PR https://github.com/tracim/tracim/pull/6634 has already improved the frontend performance issue.

There is still a backend performance issue, the PUT api call to edit the Kanban is processing for ~30s in backend.

### How to reproduce

1. Open a kanban with many cards (150+)
2. Edit anything
The kanban is designed to block edition until the save is complete. So the kanban is unusable until the save is complete.

floreal15:
I have done some investigation: 

## Probable cause 

One of the main things slowing down the edition in the Kanban is the mark_read function that set the read date of all the revision of the content (and it's sub content) with the new date.
This update is done separately for each revision of the content (and for each of the sub-content), which can make the time really long for content with a lot of revision like the Kanban;

## Solution 

### Idea

To improve this function, I see two possibilities:

- Improving the algorithm to use less sql request

- Change the way to mark a content as read (maybe we should only set the current revision to read and not every revision of the content since if we consider that read is for the user is viewing the revision, then he only sees the current one, not the past one)

### Test of a solution

I tested to improve the algorithm and keep the same use for the read, the endpoint to update the kanban got it's duration reduced by 85% (goes from around 3.8s to 580 ms for a kanban with 2000 revision and without comment)
By reducing the time of the update endpoint, the other following request is also improved greatly.



@lebouquetin what did you think we should implement?

lebouquetin:
@floreal15 if I understand correctly, a content with 200 revisions generates 200 "mark read" SQL queries?

lebouquetin:
How works your fix which reduce by 85% the duration?

floreal15:
> if I understand correctly, a content with 200 revisions generates 200 "mark read" SQL queries? 

in the mark read function, there will be 200 get read request and 200 update/insert read request

> How works your fix which reduce by 85% the duration?

It recuperates the list of revision of the content, 
then update all read for the revision that already exists,
then get the revision that isn't read yet
then insert those read how weren't on the table

my change is only for the reading of  revision, I didn't change the handling of sub-content, that is done recursively and it should also be possible to improve it

floreal15:
Follow up investigation:

[...]

PR #6743, opened 2025-04-28
fix(back): improve kanban perf 

floreal15:
https://github.com/tracim/tracim/issues/6642

## Checkpoints

**For developers**

- [x] Make sure that:

**For code reviewers**

**For testers**


backend/tracim_backend/applications/content_file/controller.py
```diff
@@ -265,8 +266,16 @@ def download_revisions_file(self, context, request: TracimRequest, hapic_data=No
...
             config=app_config,
         )
-        content = api.get_one(hapic_data.path.content_id, content_type=ContentTypeSlug.ANY.value)
-        revision = api.get_one_revision(revision_id=hapic_data.path.revision_id, content=content)
+        try:
+            revision = api.get_one_revision(
+                revision_id=hapic_data.path.revision_id, content_id=hapic_data.path.content_id
+            )
+        except ContentRevisionNotFound:
+            raise TracimFileNotFound(
```
lebouquetin: You transform an exception into another which may be false.

The right way to do this here would be to try to get the content in order to give a better information to the API user:

```
        except ContentRevisionNotFound as e:
            try:
                api.get_one(hapic_data.path.content_id, content_type=ContentTypeSlug.ANY.value)
                raise e  # the content exist but the revision does not -> raise Revision eror
            except ContentNotFound as ee:
                raise ee  # the revision does not exist BUT the content itself does not exist -> raise Content Error
```

backend/tracim_backend/applications/content_file/controller.py
```diff
@@ -385,7 +394,9 @@ def preview_pdf_full_revision(self, context, request: TracimRequest, hapic_data=
         content = api.get_one(hapic_data.path.content_id, content_type=ContentTypeSlug.ANY.value)
```
lebouquetin, line 391: Do you have access to the content file extension from the API? If so, you could event avoid the api.get_one()

backend/tracim_backend/applications/content_file/controller.py
```diff
@@ -421,7 +432,9 @@ def preview_pdf_revision(self, context, request: TracimRequest, hapic_data=None)
...
         if content.file_extension in app_config.PREVIEW__SKIPLIST:
```
lebouquetin, line 430: Same remark here - I'm not sure you already have access to the file extension but if you have, you may skip the get_one()

backend/tracim_backend/applications/content_file/controller.py
```diff
@@ -541,7 +554,9 @@ def sized_preview_jpg_revision(self, context, request: TracimRequest, hapic_data
...
```
lebouquetin, line 552: same remark

backend/tracim_backend/applications/content_html_document/controller.py
```diff
@@ -225,8 +225,9 @@ def get_html_document_revision(
             session=request.dbsession,
             config=app_config,
         )
-        content = api.get_one(hapic_data.path.content_id, content_type=ContentTypeSlug.ANY.value)
-        revision = api.get_one_revision(revision_id=hapic_data.path.revision_id, content=content)
+        revision = api.get_one_revision(
```
lebouquetin, line 232: you should catch File / Revision not found exception in the decorators

backend/tracim_backend/applications/content_html_document/controller.py
```diff
@@ -339,8 +340,9 @@ def full_pdf_revision_preview(self, context, request: TracimRequest, hapic_data=
...
```
lebouquetin, line 349: you should catch the file / revision not found exceptions in the decorators

backend/tracim_backend/lib/utils/authorization.py
```diff
@@ -258,8 +260,16 @@ def __init__(self, allowed_content_type_list: typing.List[str]):
...
-        current_content_type_slug = content_type_list.get_one_by_slug(content.type).slug
+        content_api = ContentApi(
+            config=tracim_context.app_config,
+            session=tracim_context.dbsession,
+            current_user=tracim_context.safe_current_user(),
+        )
+        content_id = tracim_context._get_content_id_in_request()
+        content_type = content_api.get_content_property(
+            property_list=[ContentRevisionRO.type], content_id=content_id
+        )[0]
```
lebouquetin: It is almost forbidden to use integer indexes: this is not maintainable. Is it really slower to map this to a dict?

backend/tracim_backend/lib/core/content.py
```diff
@@ -2490,3 +2503,55 @@ def set_content_namespace(
         )
 
         return content
+
+    def get_content_property(
+        self,
+        property_list: typing.List[QueryableAttribute],
+        content_id: int = None,
```
lebouquetin, line 2559: if values are required, do not set default values.

backend/tracim_backend/lib/core/content.py
```diff
@@ -2490,3 +2503,55 @@ def set_content_namespace(
...
+
+        Returns:
+            tuple: Values of requested properties in same order as property_list
+
+        Raises:
+            ValueError: If neither content_id nor revision_id is provided
+            ContentRevisionNotFound: If the specified revision is not found
+            ContentNotFound: If the specified content is not found
+        """
+        if revision_id is None and content_id is None:
```
lebouquetin: I suggest to use assertion here: We dont want to manage clearly a case that is not a use case.

```
assert content_id or revision_id, "Either content_id or revision_id must be provided"
```

backend/tracim_backend/lib/core/content.py
```diff
@@ -2490,3 +2503,55 @@ def set_content_namespace(
...
+    def get_content_property(
+        self,
+        property_list: typing.List[QueryableAttribute],
+        content_id: int = None,
+        revision_id: int = None,
+    ) -> typing.Tuple:
+        """
+        Return the specified properties of a revision.
+        - If revision_id is provided, return the properties for that revision.
+        - If only content_id is provided, return the properties of the last revision for that content.
```
lebouquetin, line 2566: Is there a use case where we really not to get data without to give revision_id?

backend/tracim_backend/lib/utils/authorization.py
lebouquetin: I suggest to move this to an internal method of content_api.

```
content_type = content_api.get_content_type(content_id)
```

and in content_api:

```
def _get_content_property()  # this method is private because it is too tricky for public usage
    [...]

def get_content_type(content_id):  # this is a public method ready to use from exterior
    return self._get_content_property(
            property_list=[ContentRevisionRO.type], content_id=content_id
        )[0]
```

backend/tracim_backend/views/core_api/workspace_controller.py
```diff
@@ -690,12 +691,13 @@ def get_content(self, context, request: TracimRequest, hapic_data=None) -> None:
             session=request.dbsession,
             config=app_config,
         )
-        content = api.get_one(
+
+        content = api.get_content_property(
```
lebouquetin: This is not a content, so the name is not good.

lebouquetin (changes requested):
I think all view methods should catch both revision not found and content not found exceptions

```
@hapic.handle_exception(ContentRevisionNotFound, HTTPStatus.BAD_REQUEST)
@hapic.handle_exception(ContentNotFound, HTTPStatus.BAD_REQUEST)
```

backend/tracim_backend/applications/content_file/controller.py
floreal15: I don't think we have this information in the request

backend/tracim_backend/applications/content_file/controller.py
floreal15: I don't think we have this information in the request

backend/tracim_backend/applications/content_file/controller.py
floreal15: I don't think we have this information in the request

backend/tracim_backend/lib/core/content.py
floreal15: For now it's not the case but I made this function to be also usable when we only have the revision_d

backend/tracim_backend/lib/core/content.py
floreal15: For now it's not the case, I made this function to be also usable when we only have the revision_id

backend/tracim_backend/lib/core/content.py
floreal15: done

backend/tracim_backend/views/core_api/workspace_controller.py
floreal15: done

backend/tracim_backend/lib/utils/authorization.py
floreal15: I create one get_content_property witch using _get_content_property and get_content_property get the property list as a list of string and send back a dict with the string and their value

backend/tracim_backend/applications/content_file/controller.py
floreal15: done

backend/tracim_backend/applications/content_html_document/controller.py
floreal15: done

backend/tracim_backend/applications/content_html_document/controller.py
floreal15: done

backend/tracim_backend/applications/content_file/controller.py
```diff
@@ -265,8 +268,25 @@ def download_revisions_file(self, context, request: TracimRequest, hapic_data=No
...
-        content = api.get_one(hapic_data.path.content_id, content_type=ContentTypeSlug.ANY.value)
-        revision = api.get_one_revision(revision_id=hapic_data.path.revision_id, content=content)
+        try:
+            revision = api.get_one_revision(
+                revision_id=hapic_data.path.revision_id, content_id=hapic_data.path.content_id
+            )
+        except ContentRevisionNotFound:
+            try:
+                api.get_one(hapic_data.path.content_id, content_type=ContentTypeSlug.ANY.value)
+                raise ContentRevisionNotFound(
```
lebouquetin: Is there a requirement to rebuild the exception message? What is the content of the original exception?

backend/tracim_backend/applications/content_file/controller.py
```diff
@@ -265,8 +268,25 @@ def download_revisions_file(self, context, request: TracimRequest, hapic_data=No
...
+                    "the content {content_id} has no revision {revision_id}".format(
+                        content_id=hapic_data.path.content_id,
+                        revision_id=hapic_data.path.revision_id,
+                    )
+                )
+            except ContentNotFound:
+                raise ContentNotFound(
```
lebouquetin: same remark: you catch a ContentNotFound exception, is this required to overwrite the original text?

backend/tracim_backend/applications/content_file/controller.py
lebouquetin: ok

backend/tracim_backend/applications/content_file/controller.py
lebouquetin: ok

backend/tracim_backend/lib/core/content.py
```diff
@@ -2490,3 +2503,98 @@ def set_content_namespace(
...
+            ContentNotFound: If the specified content is not found
+        """
+        assert content_id or revision_id, "Either content_id or revision_id must be provided"
+
+        list_query_attribute: typing.List[QueryableAttribute] = []
+        for property in property_list:
+            if property == "type":
+                list_query_attribute.append(ContentRevisionRO.type)
+            elif property == "workspace_id":
+                list_query_attribute.append(ContentRevisionRO.workspace_id)
```
lebouquetin: You could better write this:

```
        properties = {
              "type": ContentRevisionRO.type,
              "workspace_id": ContentRevisionRO.workspace_id
        }
        for property_name in property_list:
           list_query_attribute.append(properties[property_name])   # will raise a keyError or catch it
               
```

backend/tracim_backend/lib/core/content.py
```diff
@@ -2490,3 +2503,98 @@ def set_content_namespace(
...
+
+        result = self._get_content_property(
+            property_list=list_query_attribute, content_id=content_id, revision_id=revision_id
+        )
+
+        result_dict = {}
+
+        i = 0
+        for property in property_list:
```
lebouquetin: Write this with a zip:

```
for name, value in zip(property_list, result:
    result_dict[name] = value
```

Example if you want to test:

>>> a = ["type", "worksapce_id"]
>>> b = ["file", 4]
>>> for x, y in zip(a, b):
...     print(f"{x} = {y}")

backend/tracim_backend/applications/content_file/controller.py
floreal15: you wanted to make this check to inform the user if the content exist or not

backend/tracim_backend/applications/content_file/controller.py
floreal15: you wanted to make this check to inform the user if the content exist or not

backend/tracim_backend/lib/core/content.py
```diff
@@ -2490,3 +2503,98 @@ def set_content_namespace(
...
+            ContentNotFound: If the specified content is not found
+        """
+        assert content_id or revision_id, "Either content_id or revision_id must be provided"
+
+        query = self._session.query(*property_list)
+
+        if revision_id is not None:
+            query = query.filter(ContentRevisionRO.revision_id == revision_id)
+        else:
+            query = query.order_by(ContentRevisionRO.revision_id.desc())
```
lebouquetin: Why do you set this if revision_id is None?

backend/tracim_backend/lib/core/content.py
```diff
@@ -2490,3 +2503,98 @@ def set_content_namespace(
...
+
+        if content_id is not None:
+            query = query.filter(ContentRevisionRO.content_id == content_id)
+
+        result = query.first()
+
+        if not result:
```
lebouquetin, line 2598: you do not need to raise specific errors, here.

The easiest way would be to add a PropertyNotFound("Property {property} not found for revision {revision_id} of content {content_id}"

backend/tracim_backend/lib/utils/authorization.py
```diff
@@ -258,8 +259,16 @@ def __init__(self, allowed_content_type_list: typing.List[str]):
...
 
     def check(self, tracim_context: TracimContext) -> bool:
-        content = tracim_context.current_content
-        current_content_type_slug = content_type_list.get_one_by_slug(content.type).slug
+        content_api = ContentApi(
+            config=tracim_context.app_config,
+            session=tracim_context.dbsession,
+            current_user=tracim_context.safe_current_user(),
+        )
+        content_id = tracim_context._get_content_id_in_request()
```
lebouquetin: this is a private method, it should not be called from outside.

backend/tracim_backend/lib/utils/authorization.py
```diff
@@ -258,8 +259,16 @@ def __init__(self, allowed_content_type_list: typing.List[str]):
...
+        content_type = content_api.get_content_property(
+            property_list=["type"], content_id=content_id
```
lebouquetin: do not set keys when calling a method if they are in the right order: this will limit maintainance in case properties are renamed (and it does not help)

backend/tracim_backend/lib/core/content.py
floreal15: It's to get the last revision and it's unnecessary when we filter on the revision, but I will remove it from the else to that it's called all time and improve the code reading

backend/tracim_backend/lib/core/content.py
floreal15: done

backend/tracim_backend/lib/utils/authorization.py
floreal15: done

backend/tracim_backend/lib/utils/authorization.py
floreal15: done

backend/tracim_backend/lib/core/content.py
floreal15: done

backend/tracim_backend/lib/core/content.py
floreal15: done

backend/tracim_backend/applications/content_file/controller.py
floreal15: those exception handling have been remove since they were useless

backend/tracim_backend/applications/content_file/controller.py
floreal15: those exception handling have been remove since they were useless

backend/tracim_backend/exceptions.py
```diff
@@ -694,3 +694,7 @@ class UserCallNotFound(NotFound):
 
 class InvalidMention(TracimException):
     error_code = ErrorCode.INVALID_MENTION
+
+
+class PropertyNotFound(TracimException):
+    error_code = ErrorCode.CONTENT_NOT_FOUND
```
lebouquetin: I suggest to create a dedicated error code

backend/tracim_backend/lib/core/content.py
```diff
@@ -2490,3 +2504,101 @@ def set_content_namespace(
...
+        workspace_id=None,
+    ) -> dict:
+        """
+        Return the specified properties of a content or revision.
+        - If revision_id is provided, return the properties for that revision.
+        - If only content_id is provided, return the properties of the last revision for that content.
+        - If both are provided, ensure the revision belongs to the content.
+
+        Args:
+            property_list (list[str]): List of property names to retrieve
```
lebouquetin: Note: add the information that currently, only type and workspace_id are supported

backend/tracim_backend/lib/core/content.py
```diff
@@ -2490,3 +2504,101 @@ def set_content_namespace(
...
+            content_id (int, optional): ID of the content
+            revision_id (int, optional): ID of the specific revision
+
+        Returns:
+            dict: Dictionary mapping property names to their values
+
+        Raises:
+            ValueError: If neither content_id nor revision_id is provided
```
lebouquetin: This is wrong, I think: the raised exception would be AssertionError

backend/tracim_backend/lib/core/content.py
```diff
@@ -2490,3 +2504,101 @@ def set_content_namespace(
...
+        Args:
+            property_list (List[QueryableAttribute]): List of SQLAlchemy column attributes to retrieve
+            content_id (int, optional): ID of the content
+            revision_id (int, optional): ID of the specific revision
+
+        Returns:
+            tuple: Values of requested properties in same order as property_list
+
+        Raises:
+            ValueError: If neither content_id nor revision_id is provided
```
lebouquetin: This is probably wrong: would raise AssertionError

backend/tracim_backend/lib/core/content.py
```diff
@@ -2490,3 +2504,101 @@ def set_content_namespace(
...
+            ContentRevisionNotFound: If the specified revision is not found
```
lebouquetin: Both ContentRevisionNotFound and ContentNotFound are wrong no: you will raise PropertyNotFound

backend/tracim_backend/lib/core/content.py
```diff
@@ -2490,3 +2504,101 @@ def set_content_namespace(
...
+            content_id (int, optional): ID of the content
+            revision_id (int, optional): ID of the specific revision
+
+        Returns:
+            dict: Dictionary mapping property names to their values
+
+        Raises:
+            ValueError: If neither content_id nor revision_id is provided
+            ContentRevisionNotFound: If the specified revision is not found
+            ContentNotFound: If the specified content is not found
```
lebouquetin: Both ContentRevisionNotFound and ContentNotFound are wrong no: you will raise PropertyNotFound

backend/tracim_backend/lib/utils/authorization.py
```diff
@@ -258,8 +261,23 @@ def __init__(self, allowed_content_type_list: typing.List[str]):
...
+        workspace = tracim_context.current_workspace
+        if workspace:
+            workspace_id = workspace.workspace_id
+        content_id = tracim_context.get_content_id_in_request()
+        try:
+            content_property = content_api.get_content_property(
+                ["type"], content_id, workspace_id=workspace_id
+            )
+        except PropertyNotFound:
+            raise ContentNotFound(f"Content {content_id} not found")
```
lebouquetin, line 279: I'm not sure we should raise something else thant ContentTypeNotAllowed here.

backend/tracim_backend/lib/utils/request.py
```diff
@@ -263,6 +263,15 @@ def _get_content_id_in_request(self) -> typing.Optional[int]:
         except ContentNotFoundInTracimRequest:
             return None
 
+    def get_content_id_in_request(self) -> typing.Optional[int]:
+        """
+        Return content_id if exist, return None if content_id doesn't exist
+        This differ from _get_content_reaction_id as it does not raise exception
+        but None in case content_id doesn't exist
```
lebouquetin: IF the behavior is different, then the comment should explain *why* you implemented it a different way

lebouquetin (changes requested):
Several minor remarks/questions to fix then it will be ok

backend/tracim_backend/lib/utils/request.py
floreal15: removed the copied comment since it creates confusion

backend/tracim_backend/lib/utils/authorization.py
floreal15: it's to be consistent with the old mechanism that raised ContentNotFound when the content wasn't found (witch is the case when PropertyNotFound happend)

backend/tracim_backend/lib/core/content.py
floreal15: fixed

backend/tracim_backend/lib/core/content.py
floreal15: fixed

backend/tracim_backend/lib/core/content.py
floreal15: fixed

backend/tracim_backend/lib/core/content.py
floreal15: fixed

backend/tracim_backend/lib/core/content.py
floreal15: done

backend/tracim_backend/exceptions.py
floreal15: ok, done

lebouquetin (approved):
Good job @floreal15 :+1:

lebouquetin:
@floreal15 I let you check the manual test checkbox and merge the PR.

PR #6841, opened 2026-03-23
Feat/6834 link workspace settings fix2

### LINKED ISSUE #6834: ✨ Feat: direct access to workspace settings from sidebar
### Currently

Workspace settings are accessible from the right menu when showing the dashboard screen. This requires several clicks:

1. click on the left sidebar on the space
2. then go tothe right
3. click on "space settings"

### Expected

Allow direct access to the space settings screen from the left sidebar.

There is a `...` button which show a series of entries (see below)

Image

### How it works

- Add an entry `Settings` between the `Gallery` and `Leave space` entries
- Only show this entry if the user is allowed (ie Space administrator or super admin of the instance)

### Related PR

#6844

### What was implemented

Image

piRGoif:
Fixes #6834

### DIFF
--- a/frontend/src/component/Sidebar/SidebarSpaceItem.jsx
+++ b/frontend/src/component/Sidebar/SidebarSpaceItem.jsx
@@ -19,7 +19,6 @@ import { DRAG_AND_DROP, NO_ACTIVE_SPACE_ID } from '../../util/helper.js'
 import { putNotificationListAsRead } from '../../action-creator.async.js'
 import { newFlashMessage, readNotificationList } from '../../action-creator.sync.js'
 import { LOCK_TOGGLE_SIDEBAR_WHEN_OPENED_ON_MOBILE } from '../../container/Sidebar.jsx'
-import { CUSTOM_EVENT } from 'tracim_frontend_lib/src/customEvent'
 import appFactory from '../../util/appFactory'
 
 const qs = require('query-string')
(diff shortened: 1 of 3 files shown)


piRGoif:
In this PR, the link is added but it has a bug...

## Bug
**Initial conditions:**
Use a Tracim instance with 2 spaces, "Space1Admin" and "Space2Contrib"
Connected user "test" has :
- "Space manager" access on "Space1Admin"
- Contributor access on "Space2Contrib"

**To reproduce:**
1. Log in to Tracim with the test user
2. Click on the "Space1Admin" space link in the sidebar => this opens up the space dashboard
3. Open the "Space2Contrib" contextual menu, select the "Space information" link

**What happens:**
The "Space2Contrib" space settings opens, and the agenda link is the one for "Space1Admin"

**What should have happen:**
We should have the space information instead, as we are only contributor to the "Space2Contrib" space. 
Also the agenda URL should be the "Space2Contrib" one.


## What I understand so far
The link is opened though the OpenWorkspaceAdvanced component. Its openWorkspaceAdvanced() method is using `props.currentWorkspace` : this object comes from Redux (see mapStateToProps at the end of the file in `component/Workspace/OpenWorkspaceAdvanced.jsx`), and has correct id and label properties, but wrong memberList and agendaUrl :(

The existing space link opens up the space "dashboard" tab. it is apparently done though the WorkspacePage component, which has a updateCurrentWorkspace() method that sends updates to `props.currentWorkspace`. For testing purpose, I tried to replicate this code in OpenWorkspaceAdvanced, the update is done but has no effect on openWorkspaceAdvanced :(

piRGoif:
Closing as I've created another PR on another branch: #6834

PR #6845, opened 2026-04-08
Fix: Display template badge in search results

piRGoif:
Fixes #6840

PR #6847, opened 2026-04-09
development.ini.sample: proposition to avoid mistakes with trailing commas

piRGoif:
Having trailing commas cause a crash when loading config (ValueError: invalid literal for int() with base 10: '')

### DIFF
--- a/backend/development.ini.sample
+++ b/backend/development.ini.sample
@@ -99,23 +99,26 @@ basic_setup.session_secret = change_this_value_please!
 # control_uri for GRIP protocol reverse-proxy (PushPin) as a comma separated list
 # additional Pushpin instances can be provided
 # this is the base url used for publishing new events to reverse-proxy.
-; live_messages.control_zmq_uri = tcp://localhost:5563,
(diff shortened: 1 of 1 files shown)


piRGoif:
@lebouquetin I had issues with trailing commas when trying to configure Elastic on my dev instance (#6845). Here is my proposition to make extra clear there should be no trailing commas on single values. What do you think ?

lebouquetin:
@piRGoif please report this in an issue including the stacktrace of the error.

Documentation may help but the right way to fix it would be to make the code resilient to this problem - the stacktrace will help to see where is the problem

piRGoif:
> @piRGoif please report this in an issue including the stacktrace of the error.

Done : #6850

piRGoif:
The bug will be fixed, so I'm closing this PR.

Very important thing to notice : as mentioned in the file header, there are 2 types of comments
- "#" : examples to explain
- ";" : correspond to default values in the code
I did modify ";" comments, so this wasn't a good idea.

Issue #6074, opened 2023-01-04
🐛 Bug: missing TLMs


TheWathis:
### Tracim version

v4.4.3 build_097

### Issue description

There is some missing TLMs (to update user interface)  
So far, I've spotted creating a `share content` (see #3178) and a missing TLM (or just not handled) for models

### How to reproduce

1. Create content
2. Put the content as a model
3. Remove the content as a model

Alternatively

1. Create a content
2. Share it using the share option

### What browsers are you seeing the problem on?

Chrome, Firefox

### What devices are you seeing the problem on?

Desktop

grignards:
This old issue already mentions share contents TLMs: https://github.com/tracim/tracim/issues/3178

grignards:
For models I think that this is a modification of the content/file (`is_template` property) so I think this is "only" the frontend handling which is missing.

lebouquetin:
I think we may create a new issue agregating all missing TLMs. What do you think @piRGoif ?

piRGoif:
Yes I totally agree !

piRGoif:
Model : it is ok now in Tracim 2026.07.00

piRGoif:
I've updated the description to compile references to issues around TLMs (missing or wrongly handled)

PR #6852, opened 2026-06-11
Add favorite icon in contents list

### LINKED ISSUE #6835: ✨ Feat: Allow to easily favorite folders (and other contents)
### Feature description

Currently, contents can be added to favorites through a click on the star button in the content header

Image

This is also possible for folders but the `Edit` folder menu is counter intuitive.

### Expected

Allow to easily set a folder as a favorite. It would be interesting to be able to set a folder as favourite through the content list menu.

By the way, making this genericwould be easy add powerfull

### How it works

Maybe add a generic entry to `...` button on the content list?

- Add a `Set favorite`  entry between `Edit` and `Delete` entries?

Examples:

- folder `…` menu:

Image

- content `…` menu:

Image

### Related PR

_No response_

piRGoif:
Fixes #6835

Note that I first tried to add a contextual menu item
But as the menu isn't displayed for Reader access users, I switched (latest commit) to the same icon as in content details

Test plan : 
- ✅ content list
  - Firefox tab 1 user admin content list
  - Firefox tab 2 user admin folder app
  - Firefox tab 3 user test content list
  - Chrome tab 4 user admin content list
  - on / off favorite, check state update on tabs
- apps : Firefox tab 1 user admin content list, Firefox tab 2 user admin app, on / off favorite and check state update
  - ✅ app folder
  - ✅ app logbook
  - ✅ app topic
  - ✅ app file
  - ✅ app note
  - ✅ app kanban
- ❗️ "my favourites" screen : Firefox tab 1 user admin content list, Firefox tab 2 user admin my favourites, on / off favorite and check state update
  - favorite is added but "content is not available" error displayed


frontend/src/container/WorkspaceContent.jsx
```diff
@@ -149,6 +157,32 @@ export class WorkspaceContent extends React.Component {
...
+  handleRemoveContentFromFavoriteList = data => {
+    if (data.source === 'shell') return
+    const { props } = this
+    const favorite = props.favoriteList.find(f => f.contentId === data.content_id)
+    if (!favorite) return
+    props.dispatch(removeFavorite(favorite))
+  }
+
+  /**
+   * WORKAROUND - PG - 2026-06-11 - Favorite status is not returned by the content list API,
```
lebouquetin, line 176: For information: this is expected in order to align with RESTful philosophy: the API returns the same content for all users. Favorites are related both to content and user, so in the case we would return the favorite status in the content list endpoint, the result would differ from a user to another one.

We try to keep this behavior in the API in order to make debugging and understanding easier. 

Example:

- hey, I do not get the right information when getting data from API route /xxx
- You sure? I do exactly the same request and I get the right information

In the same philosophy, routes like `/api/users/me` are always HTTP 302 redirections to `/api/users/<user_id>`

frontend/src/action-creator.async.js
```diff
@@ -1388,7 +1388,7 @@ export const postContentToFavoriteList = (userId, contentId) => async dispatch =
         ...FETCH_CONFIG.headers
       },
       method: 'POST',
-      body: { content_id: contentId }
+      body: JSON.stringify({ content_id: contentId })
```
lebouquetin: What is the interest of this change?

frontend_lib/src/appContentFactory.js
```diff
@@ -906,21 +928,27 @@ export function appContentFactory (WrappedComponent) {
...
+      }))
+      GLOBAL_dispatchEvent({ type: CUSTOM_EVENT.REMOVE_CONTENT_FROM_FAVORITE_LIST, data: { content_id: content.content_id, source: 'app' } })
     }
 
     loadFavoriteContentList = async (loggedUser, setState) => {
       const response = await handleFetchResult(await getFavoriteContentList(this.apiUrl, loggedUser.userId))
       if (!response.ok) {
         sendGlobalFlashMessage(i18n.t('Error while getting favorites'))
         setState({ favoriteList: [] })
+        this.setState({ favoriteList: [] })
```
lebouquetin: why do you have both

```
setState({ favoriteList: [] })
this.setState({ favoriteList: [] })
```

Isn't it the same thing?

frontend_lib/src/appContentFactory.js
```diff
@@ -906,21 +928,27 @@ export function appContentFactory (WrappedComponent) {
...
         return
       }
       const favorites = response.body
       setState({ favoriteList: favorites.items })
+      this.setState({ favoriteList: favorites.items })
```
lebouquetin: why do you have both

```
setState({ favoriteList: [] })
this.setState({ favoriteList: [] })
```

Isn't it the same thing?

frontend_lib/src/appContentFactory.js
```diff
@@ -906,21 +928,27 @@ export function appContentFactory (WrappedComponent) {
...
     }
 
-    isContentInFavoriteList = (content, state) => {
-      return state.favoriteList && state.favoriteList.some(
+    isContentInFavoriteList = (content) => {
```
lebouquetin: What is the interest of this refactor?

frontend/src/component/Workspace/ContentItem.jsx
```diff
@@ -105,8 +109,16 @@ class ContentItem extends React.Component {
             }}
           />
 
-          {props.userRoleIdInWorkspace >= ROLE.contributor.id && (
-            <div className='d-none d-md-block content__actions' title={props.t('Actions')}>
+          <div className='d-none d-md-flex content__actions'>
```
lebouquetin: Two questions here:

- why do you replace `d-md-block` with `d-md-flex` class? 
- you remove the title of the item, why?

piRGoif: Good catch thanks, sorry I missed that one :/ I fixed it in ccdc0ade

frontend/src/action-creator.async.js
piRGoif: I struggled a lot so that contents list and content details (in a distinct app) are kept in sync, and this was part of the many modifications. I've just pushed 8e13c5ce to factorize the code used when clicking on both favorite icons. This function was removed by that commit.

frontend_lib/src/appContentFactory.js
piRGoif: The appContentFactory component is the HOC that embeds the content details apps. We need to update both local component state and appContentFactory own state so that both places (contents list and content details) stays in sync. 

Trying to factorize state would mean modifying all apps, so I was reluctant to do that ? I propose we have a face to face chat about that, I'll send you a private message

In the meantime I've added a comment in 44e6740a

frontend_lib/src/appContentFactory.js
piRGoif: As we have 2 states, the consumer no longer needs to pass its own, and we are relying on the appContentFactory state.

frontend_lib/src/appContentFactory.js
piRGoif: Closing this thread as it is the same issue discussed in the one above

frontend_lib/src/appContentFactory.js
piRGoif: Discussed with Damien : we should use the TLM instead...

piRGoif:
Just made a test both on my local dev Tracim instance and Algoo's instance: 

- having 3 tabs: 
  1. content A detail 
  2. content A detail again
  3. "my favourites" screen
- clicking the favourite icon in screen 1
- checking what changed in tabs 2 and 3

=> No update on tabs 2 and 3 :( In tab 3 update appears after a manual tab reloading though, but it doesn't in tab 2 :(

Didn't find an issue covering this exact use case, but there are multiple missing TLM issues existing, like #6074.

piRGoif:
Before starting implementing TLM for favorites, I've squashed (but kept changelog commit as is) and rebased the branch

backend/tracim_backend/lib/core/content.py
```diff
@@ -2427,10 +2427,8 @@ def add_favorite(self, content: Content, do_save: bool = True) -> FavoriteConten
         return favorite
 
     def remove_favorite(self, content_id: int, do_save: bool = True) -> None:
-        self._session.query(FavoriteContent).filter(
-            FavoriteContent.user_id == self._user.user_id,
-            FavoriteContent.content_id == content_id,
-        ).delete()
+        favorite = self.get_one_user_favorite_content(self._user.user_id, content_id)
+        self._session.delete(favorite)
```
lebouquetin, line 2431: What is the purpose of this refactor?

backend/tracim_backend/lib/core/event.py
```diff
@@ -104,6 +106,7 @@ class EventApi:
     workspace_subscription_schema = WorkspaceSubscriptionSchema()
     user_call_schema = UserCallSchema()
     user_config_schema = UserConfigSchema()
+    user_favorite_schema = UserFavoriteSchema()
```
lebouquetin: I do not see the interest to prefix user_x for favorites. 

For call and config it make sense:

- config_schema would not be very explicit
- call_schema is too generic because there are user_calls and group_calls are planned

I would have named this favorite_schema - like it is for reactions.

Do you think there are real interest in using a user_ prefix?

backend/tracim_backend/lib/core/event.py
```diff
@@ -884,6 +887,36 @@ def on_user_call_deleted(self, user_call: UserCall, context: TracimContext) -> N
     def on_user_config_modified(self, user_config: UserConfig, context: TracimContext) -> None:
         self._create_user_config_event(OperationType.MODIFIED, user_config, context)
 
+    @hookimpl
+    def on_favorite_content_created(
```
lebouquetin: FavoriteContent looks like a bad naming (not your fault). I suggest to name events only on the favorite base: `on_favorite_created`, `on_favorite_deleted` and `_create_favorite_event()`

Events are targetted to be used by external developers who don't care with consistency with tracim core source code, so this is the right place to rename things if they need to

backend/tracim_backend/lib/core/event.py
```diff
@@ -884,6 +887,36 @@ def on_user_call_deleted(self, user_call: UserCall, context: TracimContext) -> N
...
+    ) -> None:
+        self._create_user_favorite_event(OperationType.CREATED, favorite_content, context)
+
+    @hookimpl
+    def on_favorite_content_deleted(
+        self, favorite_content: FavoriteContent, context: TracimContext
+    ) -> None:
+        self._create_user_favorite_event(OperationType.DELETED, favorite_content, context)
+
+    def _create_user_favorite_event(
```
lebouquetin: suggestion: rename to `_create_favorite_event`

backend/tracim_backend/lib/core/event.py
```diff
@@ -1199,6 +1232,15 @@ def _get_user_config_event_receiver_ids(
     return {event.fields["user"]["user_id"]}
 
 
+def _get_user_favorite_event_receiver_ids(
```
lebouquetin: suggest to rename to `_get_favorite_event_receiver_ids`

backend/tracim_backend/lib/core/event.py
```diff
@@ -1214,6 +1256,7 @@ class BaseLiveMessageBuilder(abc.ABC):
         EntityType.TAG: _get_workspace_event_receiver_ids,
         EntityType.USER_CALL: _get_user_call_event_receiver_ids,
         EntityType.USER_CONFIG: _get_user_config_event_receiver_ids,
+        EntityType.USER_FAVORITE: _get_user_favorite_event_receiver_ids,
```
lebouquetin: I'm not very happy with the USER_ prefix … I can understand the fact that favorites are private - but maybe we would show statistics at a moment like "this content is 19x favorited".

What do you think?

backend/tracim_backend/lib/crud_hook/hookspec.py
```diff
@@ -172,3 +173,15 @@ def on_user_config_modified(self, user_config: UserConfig, context: TracimContex
     @hookspec
     def on_user_config_deleted(self, user_config: UserConfig, context: TracimContext) -> None:
         ...
+
+    @hookspec
+    def on_favorite_content_created(
```
lebouquetin: I'm definitly not happy with favorite_content naming. and this is the right place to rename favorite_content to favorite

backend/tracim_backend/lib/crud_hook/hookspec.py
```diff
@@ -172,3 +173,15 @@ def on_user_config_modified(self, user_config: UserConfig, context: TracimContex
...
+        self, favorite_content: FavoriteContent, context: TracimContext
+    ) -> None:
+        ...
+
+    @hookspec
+    def on_favorite_content_deleted(
```
lebouquetin: same suggestion for the naming

backend/tracim_backend/tests/functional/test_live_messages.py
```diff
@@ -512,3 +512,50 @@ def test_api__user_live_messages_endpoint_with_GRIP_proxy__ok__content_tag__asyn
...
+        assert favorite_created_event.event == "message"
+        result = json.loads(favorite_created_event.data)
+        assert result["event_type"] == "user_favorite.created"
+        assert result["read"] is None
+        assert result["fields"]
+        assert result["created"]
+        assert result["event_id"]
+        assert result["fields"]["user"]
+        assert result["fields"]["user"]["user_id"] == 1
+        assert result["fields"]["user_favorite"]
```
lebouquetin: Typically, here I would naturally expect a favorite (common name for all people) and not user_favorite

backend/tracim_backend/views/core_api/schemas.py
```diff
@@ -1944,6 +1946,15 @@ class PaginatedFavoriteContentSchema(BasePaginatedSchemaPage):
     items = marshmallow.fields.Nested(FavoriteContentSchema, many=True)
 
 
+class UserFavoriteSchema(marshmallow.Schema):
```
lebouquetin: rename to `FavoriteSchema`?

frontend/src/component/Workspace/ContentItem.jsx
```diff
@@ -11,6 +11,8 @@ import {
 import BtnExtendedAction from './BtnExtendedAction.jsx'
 import {
   APP_CUSTOM_ACTION_LOCATION_OBJECT,
+  FAVORITE_STATE,
+  FavoriteButton,
```
lebouquetin, line 15: Rename to FavoriteToggleButton ?

frontend/src/component/Workspace/Folder.jsx
```diff
@@ -8,6 +8,8 @@ import BtnExtendedAction from './BtnExtendedAction.jsx'
 import ContentItem from './ContentItem.jsx'
 import {
   DropdownMenu,
+  FAVORITE_STATE,
+  FavoriteButton,
```
lebouquetin, line 12: Rename to FavoriteToggleButton

frontend/src/container/WorkspaceContent.jsx
```diff
@@ -149,6 +159,32 @@ export class WorkspaceContent extends React.Component {
     if (!this.state.appOpenedType) this.setHeadTitle(this.getFilterName(qs.parse(this.props.location.search).type))
   }
 
+  handleExternalAddContentToFavoriteList = data => {
```
lebouquetin: Suggestion (if I understand correctly): rename `handleExternalAddContentToFavoriteList` to `handleAddContentToFavoriteList`

frontend/src/container/WorkspaceContent.jsx
```diff
@@ -149,6 +159,32 @@ export class WorkspaceContent extends React.Component {
...
+    if (data.source === 'shell') return
+    this.props.dispatch(addFavorite(data))
+  }
+
+  handleExternalRemoveContentFromFavoriteList = data => {
```
lebouquetin: Suggestion (if I understand correctly): rename `handleExternalRemoveContentFromFavoriteList` to `handleRemoveContentFromFavoriteList`

frontend_lib/src/index.js
```diff
@@ -254,6 +255,7 @@ import {
   getUsernameAvailability,
   getWorkspaceContentList,
   getWorkspaceDetail,
+  postContentToFavoriteList,
```
lebouquetin, line 258: why `postContentToFavoriteList` and not `postContent`?

frontend_lib/src/index.js
```diff
@@ -461,6 +463,7 @@ export {
   darkenColor,
   defaultApiContent,
   defaultDebug,
+  deleteContentFromFavoriteList,
```
lebouquetin, line 466: why `deleteContentFromFavoriteList` and not `deleteFavorite`?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### ✨ New Features
 
 - Show the template badge in search results [#6840](https://github.com/tracim/tracim/issues/6840)
+- Show the favorite icon in the content list with real-time sync across tabs (new favorite TLM) [#6835](https://github.com/tracim/tracim/issues/6835)
```
lebouquetin: Put two lines for this:

One is in the new features:

- Show the favorite icon in the content list

The other is a bugfix:

- implement favorite TLM for live propagation

Note:

- we do not put issue reference in the new features - most of the time new features are implemented through a series of tickets. 
- we always set an issue reference for bugfixes

lebouquetin (commented):
Several renaming may be required. In my point of view:

- `user_favorite` is redundant: `favorite` seems to be enough
- `favorite_content` is a naming error - we should avoid to propagate it and this is the right moment to avoid this propagation through TLM, hooks, etc.

I'm also not sure you really implemented a unique way of user interface refresh when toggling favorite state: the presence of `handleExternalAddContentToFavoriteList` and `handleExternalRemoveContentFromFavoriteList` …

naming may be simplier, like `ADD_CONTENT_TO_FAVORITE_LIST` or `handleClickAddToFavoriteList` (why not `ADD_FAVORITE` and `handleClickAddFavorite`?)

These naming may be discussed - the important part for the naming are public interfaces which cant be changed easily after being published

lebouquetin (commented):
Another comment: could you share up-to-date screenshots (better in the issue) in order I can review the CSS?

piRGoif:
> Another comment: could you share up-to-date screenshots (better in the issue) in order I can review the CSS?

The issue description contains the latest rendering ("what was implemented" title)

piRGoif:
TLM info learned during dev:
- they are working just right using serdev
- no log in the devtools network tab, either in Firefox or Chrome
- when a message is generated a log appears in the backend server log as DEBUG for SyncLiveMessageBuilder with strings "publish event" or "Sending event"
- the events are saved server side in the events table
- also a log appears in the devtools console, as info, with string ".:. TLM received" (only for master tab) and "GLOBAL_dispatchLiveMessage"
- ReduxTlmDispatcher contains the client SSE subscriptions

piRGoif:
Current "my favourites" screen error:
image

frontend_lib/src/appContentFactory.js
piRGoif: As TLM impl is now working, I'm closing this conversation

frontend_lib/src/appContentFactory.js
piRGoif: With the TLM impl the code was refactored again, so closing conversation

backend/tracim_backend/lib/core/event.py
piRGoif: Well, we don't have that need now, and favorites were added a long time ago... So I would recommend renaming to FAVORITE, and use a more specific name when adding something else (like FAVORITE_STATS for your example ?)

frontend/src/component/Workspace/ContentItem.jsx
piRGoif: The FavoriteButton component was created 5 years ago... why renaming it now ?

frontend/src/component/Workspace/Folder.jsx
piRGoif: Closing this conversation as it is the same suggestion as in ContentItem.jsx

CHANGELOG.md
piRGoif: Line split done in 697d4fd3

About issue reference for enhancements, can we have a chat about it ?

backend/tracim_backend/lib/core/event.py
piRGoif: Done in c8ab1d87

backend/tracim_backend/lib/core/event.py
piRGoif: Done in c8ab1d87

backend/tracim_backend/lib/core/event.py
piRGoif: Done in c8ab1d87

backend/tracim_backend/lib/core/event.py
piRGoif: Done in c8ab1d87

backend/tracim_backend/lib/crud_hook/hookspec.py
piRGoif: Done in c8ab1d87

backend/tracim_backend/lib/crud_hook/hookspec.py
piRGoif: Done in c8ab1d87

backend/tracim_backend/tests/functional/test_live_messages.py
piRGoif: Done in c8ab1d87

backend/tracim_backend/views/core_api/schemas.py
piRGoif: Done in c8ab1d87

frontend/src/container/WorkspaceContent.jsx
piRGoif: Done in c8ab1d87

frontend/src/container/WorkspaceContent.jsx
piRGoif: Done in c8ab1d87

frontend/src/container/WorkspaceContent.jsx
piRGoif: Note : Claude explained he used that prefix because : 

> In Tracim's micro-app architecture, the shell (frontend/) passes callback functions down to micro-apps. I likely used the External prefix to distinguish two categories of handlers in WorkspaceContent.jsx:
>
> 1. Internal handlers — functions the component calls in response to its own user interactions (e.g. handleClickFavorite)
> 2. "External" handlers — functions passed to a micro-app as props/callbacks, meaning they get called by the app when it wants to tell the shell something happened
>
> The idea was: from the micro-app's perspective, these functions come from "outside" (the shell), hence handleExternal*.

I didn't find any no other existing function with this prefix, so we can remove it.

backend/tracim_backend/lib/core/content.py
piRGoif: Before we had .query().filter().delete(), which did a bulk delete.
Claude did change to loading the object first in the session then calls delete in order to trigger the ORM lifecycle hooks (session.delete call).

Do you think this is a bad move, or there is a better way to do it ?

frontend_lib/src/index.js
piRGoif: Closing conversation as I will continue in previous one (postContentToFavoriteList)

frontend_lib/src/index.js
piRGoif: In action.async.js we have 3 favorite functions to call the API : 
- getFavoriteContentList
- postContentToFavoriteList
- deleteContentFromFavoriteList

These functions were created in 2021 (commit 01e284d) when introducing the favorites. 

I think we should keep both favorite and content in post and delete functions, do you agree ? Do you have better names in mind ?

frontend/src/component/Workspace/ContentItem.jsx
lebouquetin: Right - I thought it was a new component

backend/tracim_backend/lib/core/content.py
lebouquetin: > Do you think this is a bad move, or there is a better way to do it ?

I have no idea; The only difference I see is the bulk delete would allow to delete several items while the explicit get_one/delete would limit to delete a unique instance.

Let's keep the refactor.

piRGoif:
> Current "my favourites" screen error

I've just push 2 commits that fixes all the problems I saw in the "my favourites" screen.

PR is ready for review !

frontend_lib/src/index.js
piRGoif: Discussed with Damien: ok to keep it as is as it is legacy so out of scope of this PR

CHANGELOG.md
piRGoif: After a discussion with Damien, I've updated internal doc and updated changelog once again in 01f68661

piRGoif:
Rebased on develop + force push (there was 1 commit ahead)

piRGoif:
I had doubts about the SSE payload sent when adding a favorite.

Turns out that the SSE contains multiple properties but not the content body itself : 
- in backend/tracim_backend/views/core_api/schemas.py
- in FavoriteContentSchema.content definition we have `content = Nested(ContentDigestSchema)`
- in ContentDigestSchema we have all the properties
- DB record is get using tracim_backend.lib.core.content.ContentApi.get_one_user_favorite_content_in_context
- serialization is done in tracim_backend.lib.core.event.EventBuilder._create_favorite_event : we are dumping a FavoriteContentInContext to a FavoriteContentSchema

When loading the "my favourites" screen we are:
- Favorites.jsx : call to loadFavoriteList()
- API call to GET /api/users/{userId}/favorite-contents
- for each favorite an extra API call to GET /api/contents/{contentId}/path

With the TLM update, when a favorite created TLM is received, the "my favourites" screen does:
- ReduxTlmDispatcher : handleUserFavoriteCreated call the reducer
- Favorites.jsx componentDidUpdate detects favoriteList changed, finds new favorites by contentId diff
- for each new favorite with non-null content, fires GET /api/contents/{id}/path, then dispatches updateFavoriteList with the populated breadcrumbs
- component re-render

=> we have a similar behavior between component first load and favorite added with TLM.


TO me the PR is now ready for review !

lebouquetin:
What I understand from your last comment and investigations:

- the backend dispatch favorites as a content list - so favorites are actually serialized as contents or something similar)
- this behavior is reproduced in TLM - which is consistant

The drawback is now we load twice contents when showing list of contents:

- we already load contents
- in order to show the favorite / not favorite state, we load the favorites of the user - which also includes the associated content properties

Does this match the current behavior? (no need to change or fix this, its just to make things clear)

piRGoif:
> The drawback is now we load twice contents when showing list of contents

Discussed with Damien. I've added a note about this in the code (commit b238b47e) as an issue might be too much (we won't have the problem in our installed instances...)

frontend/src/action-creator.sync.js
```diff
@@ -323,6 +323,6 @@ export const setFavoriteList = (favoriteList) => ({ type: SET_FAVORITE_LIST, fav
 export const UPDATE_FAVORITE_LIST = `${UPDATE}/${FAVORITE_LIST}`
 export const updateFavoriteList = (favoriteList) => ({ type: UPDATE_FAVORITE_LIST, favoriteList })
 export const removeFavorite = (favorite) => ({ type: `${REMOVE}/${FAVORITE}`, favorite })
-export const addFavorite = (favorite) => ({ type: `${ADD}/${FAVORITE}`, favorite })
+export const addFavorite = (favorite) => ({ type: `${ADD}/${FAVORITE_LIST}`, favorite })
```
lebouquetin: Why do you rename this to FAVORITE_LIST?

backend/tracim_backend/lib/core/event.py
```diff
@@ -884,6 +887,40 @@ def on_user_call_deleted(self, user_call: UserCall, context: TracimContext) -> N
     def on_user_config_modified(self, user_config: UserConfig, context: TracimContext) -> None:
         self._create_user_config_event(OperationType.MODIFIED, user_config, context)
 
+    @hookimpl
+    def on_favorite_created(
+        self, favorite_content: FavoriteContent, context: TracimContext
```
lebouquetin: I think `favorite` would be a better name than `favorite_content` as this is a public API.

backend/tracim_backend/lib/core/event.py
```diff
@@ -884,6 +887,40 @@ def on_user_call_deleted(self, user_call: UserCall, context: TracimContext) -> N
...
+    ) -> None:
+        self._create_favorite_event(OperationType.CREATED, favorite_content, context)
+
+    @hookimpl
+    def on_favorite_deleted(
+        self, favorite_content: FavoriteContent, context: TracimContext
```
lebouquetin: I think `favorite` would be a better name than `favorite_content` as this is a public API.

backend/tracim_backend/lib/crud_hook/caller.py
```diff
@@ -65,6 +66,10 @@ def _call_hooks(
                 self._plugin_manager.hook.on_user_call_created(
                     user_call=obj, context=session.context
                 )
+            elif isinstance(obj, FavoriteContent):
+                self._plugin_manager.hook.on_favorite_created(
+                    favorite_content=obj, context=session.context
```
lebouquetin: Also rename here `favorite` instead of `favorite_content=obj`

backend/tracim_backend/lib/crud_hook/caller.py
```diff
@@ -146,6 +151,10 @@ def _call_hooks(
                 self._plugin_manager.hook.on_user_call_deleted(
                     user_call=obj, context=session.context
                 )
+            elif isinstance(obj, FavoriteContent):
+                self._plugin_manager.hook.on_favorite_deleted(
+                    favorite_content=obj, context=session.context
```
lebouquetin: Also rename here `favorite` instead of `favorite_content=obj`

backend/tracim_backend/lib/crud_hook/hookspec.py
```diff
@@ -172,3 +174,15 @@ def on_user_config_modified(self, user_config: UserConfig, context: TracimContex
     @hookspec
     def on_user_config_deleted(self, user_config: UserConfig, context: TracimContext) -> None:
         ...
+
+    @hookspec
+    def on_favorite_created(
+        self, favorite_content: FavoriteContent, context: TracimContext
```
lebouquetin: I think `favorite` would be a better name than `favorite_content` as this is a public API.

backend/tracim_backend/lib/crud_hook/hookspec.py
```diff
@@ -172,3 +174,15 @@ def on_user_config_modified(self, user_config: UserConfig, context: TracimContex
...
+    ) -> None:
+        ...
+
+    @hookspec
+    def on_favorite_deleted(
+        self, favorite_content: FavoriteContent, context: TracimContext
```
lebouquetin: I think `favorite` would be a better name than `favorite_content` as this is a public API.

lebouquetin (commented):
I suggest to rename `favorite_content`  to `favorite` in the public APIs (for the parameters).

frontend/src/action-creator.sync.js
piRGoif: It was a bug fix, and it disappeared from the commit message when I squashed :/ I'll had it in the final version !

To summarize: 
- addFavorite in frontend/src/action-creator.sync.js is called using dispatch, for example in frontend/src/container/WorkspaceContent.jsx
- when dispatch is called React search for a reducer available to handle the action
- the corresponding catch is in frontend/src/reducer/favoriteList.js
- both the action function and the reducer were added in 2021 (commit 01e284d8), but there was a mistake on the action

lebouquetin: I did not understand. What do you mean by "when dispatch is called React a reducer available to handle the action" ?

lebouquetin: What was the bug (what is the behavior without this fix, what is the behavior with this fix?)

piRGoif: Woops sorry, I fixed my initial message
Call me if it is still not very clear ?

piRGoif: The bug was that the value wasn't added in Redux. But this might not be a problem in content apps as they were relying on an internal state in appContentFactory

I've just checked the 4 actions and updated the code accordingly : 
- setFavoriteList and updateFavoriteList are receiving lists therefore their type is FAVORITE_LIST
- removeFavorite and addFavorite are receiving a single item therefore they should both use FAVORITE and not FAVORITE_LIST

piRGoif:
@lebouquetin about renaming favorite_content to favorite: there are 2 legacy places I left untouched, and I think we should keep them that way:
- tracim_backend.lib.core.content.ContentApi.get_one_user_favorite_content_in_context: function name reflects perfectly what it does
- FavoriteContentInContext: we also have a FavoriteContent class. The name reflects its usage also

lebouquetin:
> about renaming favorite_content to favorite: there are 2 legacy places I left untouched, and I think we should keep them that way:

Renaming those part is not required - they are internal and must be as consistant as possible. So I agree not to touch them

piRGoif:
Damien is ok to merge. Let's do a final squash to reduce the number of commits

lebouquetin:
> Let's do a final squash to reduce the number of commits

@piRGoif I think we keep only one commit per issue on the main branch (edit: you can squash & merge directly)

piRGoif:
> > Let's do a final squash to reduce the number of commits
> 
> @piRGoif I think we keep only one commit per issue on the main branch (edit: you can squash & merge directly)

My idea was to have a commit for the first implementation using state and another for the TLM, and merge fast-forward ("rebase and merge" github action). Are you ok with that ?

lebouquetin:
> My idea was to have a commit for the first implementation using state and another for the TLM, and merge fast-forward ("rebase and merge" github action). Are you ok with that ?

No, it's over-engineering. I expect to merge this development fast and to go forward on next issues. Explanation shared on a private discussion.

Issue #6711, opened 2025-03-25
⚙️ Misc: Python 3.12 will break SETUPTOOLS_USE_DISTUTILS=stdlib


Skylsmoi:
In the Dockerfile files we set:
```bash
ENV SETUPTOOLS_USE_DISTUTILS=stdlib
```

This is a workaround for setuptools 60+ (see https://github.com/tracim/tracim/issues/5183).

This issue in setuptools repo: https://github.com/pypa/setuptools/issues/3625 tells that it won't be usable on python 3.12+.

We either need to upgrade tracim's environment or to fix setuptools version when building Docker.

Tracim's backend dependencies set setuptools to 57.
See: https://github.com/tracim/tracim/blob/develop/backend/requirements-build.txt

But a dependency from Dockerfile install the latest version.

Probably in https://github.com/tracim/tracim/blob/develop/system_packages/debian/build_backend_packages.list (not confirmed).

Skylsmoi:
Investigation made during the build of the build_158 which broke because of the release of setuptools v78 which made our setup.cfg syntax outdated.

Fixed in: https://github.com/tracim/tracim/commit/dcbb24a6bfb29c8a48753e70390044a74db47822

piRGoif:
Closing this issue as it isn't relevant anymore for Python >= 3.12 and Setuptools >= 60.0.0

Explanations : 

Setuptools 60.0.0 (Dec 2021) bundled its own reimplementation of distutils and made it the default, replacing Python's stdlib distutils. This broke some packages that relied on stdlib distutils behaviour. The env var forced setuptools to use Python's own distutils instead of its bundled version.
=> source for distutils now bundled in setuptools : [Porting from Distutils - setuptools 83.0.0 documentation](https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html)

Since Python 3.12, distutils was removed from the stdlib entirely, so setuptools' bundled version is now the only distutils that exists.
=> source https://docs.python.org/3/whatsnew/3.12.html "the distutils package has been removed from the standard library"

Issue #1216, opened 2018-11-30
sidebar + dashborad: refresh problem when creating new shared space


PhilippeAccorsi:
At present:

If I'm in one shared space with some content in recent activity, when I create new shared space I can see in recent activity of this new shared space the list of content that we have in previously sharedspace.

Expected

- [ ] dashboard just show content of the associated sharedspace

tracim:
@Skylsmoi we must not have this kind of problem. All content stuff must be in a workspace context, so that we can't have content in wrong workspace. Contents are a sub-items of a workspace, it must be the same in frontend architecture.

When switching from a workspace to another, all stuff related to the workspace must be refreshed automatically (I mean... you must not code "refresh recent activity, refresh content list, refresh x or y: you code refresh workspace and everything must be contextualized so that this problem cant happen)

Skylsmoi:
@lebouquetin It already is like that. The problem was a handler of custom event that was declared in two places and caused a double refresh of data with, for one the them, the previous workspace id. The data displayed was always the longest to respond.

Skylsmoi:
https://github.com/tracim/tracim/commit/91e58ae4f1074a263e888c629cf3e04603e23c77

tracim:
ok, thanks for the explanation

PhilippeAccorsi:
Verified.

Issue #608, opened 2018-06-28
Rename "owner" to "author"


inkhey:
Rename owner to author in lib and db for content and content revision + do the necessary to handle migration.
https://github.com/tracim/tracim/issues/606

lebouquetin:
still existing @lebouquetin

PR #6860, opened 2026-07-03
support Python 3.13

### LINKED ISSUE #6856: Update to Python 3.13
Some tests were conducted with Debian 13 and turns out we have multiple problems when trying to run with Python 3.13 (see https://github.com/tracim/tracim/issues/6766#issuecomment-2929714203 for the original message)

- requirements-build.txt: setuptools, wheel and pip versions aren't compatible with Python 3.13
- requirements.txt: requires dependencies that seem to use APIs that were removed in python 3.13
- errors on db init (?)

Note there is an existing ticket about Python 3.12 issue with SetupTools: #6711

### LINKED ISSUE #6859: ✨ Feat: Support current Debian stable version (Trixie 13)
### Current status

Currently,

- Tracim only supports Python 3.9 to 3.11
- Docker image rely on Debian 11

Also, python 3.12+ is not supported:

- Python 3.12 support see #6711
- Python 3.13 support see #6856

Also we detected issues with the setup when running on Debian 13 with Python 3.0 or 3.11: see #6766

### Feature description

It is expected to upgrade the building blocks on which Tracim relies:

- Use Debian Trixie (13) as the docker root image
- Support Python 3.13 (Python 3.12 is not in the target)
- Use Pushpin from the project .deb repository: https://pushpin.org/docs/install/

There are multiple related issues:

- FIX 
  - Python version support
    - #6856
  - issues when running on Debian 13
    - #6766
    - #6864
    - #6865
- To do next
  - #6866
- WONTFIX 
  - #6711: goal is to support Python 3.13 and not lower versions
  - #6863: discovered during work on Python 3.13, but this is unrelated and out of scope

### How it works

- No issue when running setup
- No issue when operating the software

piRGoif:
Fixes #6856 and refs #6859

TODO:
- setup_default_backend.sh with Python 3.13
  - [x] solve errors when launching setup_default_backend.sh
    - [x] document modifications done
    - [x] document problems out of pr scope => #6861
- hapic update
  - [x] hapic switch to develop
     - [x] update marshmallow dependency
     - [x] update marshmallow usages in Tracim code
- dev setup
  - [x] pip install pytest pytest-testinfra python-dotenv
  - pip install -r requirements-dev.txt => ok with no modification
    - ~~requirements.txt~~ => already validated with setup_default_backend.sh
    - ~~requirements-test.txt~~ => differed in #6878
    - [x] requirements-devtool.txt
  - [x] pre-commit install
- code run
  - ~~auto test~~ => differed in #6878
    - [x] #6864
  - [x] local run
    - [x] cannot login ("Unknown field" in http response)
  - Docker
    - ~~Concourse~~ => differed in #6878
    - [x] Build => WIP Pierre
      - [x] local pushpin : Debian 13 + pusphin from Pushpin APT repo
      - [x] update dependencies + base os
      - [x] Node 16 based on Trixie
    - [x] Run
      - [x] Startup log
      - [x] Test : login, create user, 2 tab with 2 users, create space, create note, react to new note
- remaining tasks
  - [x] hapic new Python 3.13 compatible package 
  - ~~black~~ => in pyproject.toml, `[tool.black]` section there are no `target-version` set
- document
   - [x] report information added in this PR to the corresponding issue
   - [x] update Tracim documentation (see previous 61b29f1a)
   - [x] `pip-compile setup.py`
   - [x] changelog
   - ~~steps to upgrade to latest lib versions~~ => will be handled in #6866


piRGoif:
Other issues I saw:
(updated after a chat with Damien)

- ❗️ rawpy needed to be updated, is preview-generator still compatible ? 
- ❗️ hapic : for now the error with `import cgi` is solved as the webob which have the same issue imports cgi-legacy. But we need to update to latest develop HEAD which contains fixes for Python 3.13 - and this causes marshmallow issues (this lib being used in Tracim)
- 🤔 "parser error : Document is empty" when populating DB: this was already there with Python 3.11, but there aren't any existing issue :/ => non blocking, but what is the impact ?
- ❌ npm WARN deprecated [lodash.get@4.4.2](mailto:lodash.get@4.4.2) => not Python related, warning and non blocking, was already existing, so disregarding for now

lebouquetin:
>  rawpy needed to be updated, is preview-generator still compatible ?

@piRGoif what issue did you face? I installed it without problem on trixie / python 3.13.5

piRGoif:
> > rawpy needed to be updated, is preview-generator still compatible ?
> 
> @piRGoif what issue did you face? I installed it without problem on trixie / python 3.13.5

When running `./setup_default_backend.sh` I get : 

```
[10:26:18] $ Installing all preview dependencies from requirements-full-preview-generator.txt
Requirement already satisfied: vtk==9.4.0 in ./env/lib/python3.13/site-packages (from -r requirements-full-preview-generator.txt (line 2)) (9.4.0)
Requirement already satisfied: xvfbwrapper==0.2.9 in ./env/lib/python3.13/site-packages (from -r requirements-full-preview-generator.txt (line 4)) (0.2.9)
Requirement already satisfied: ffmpeg-python==0.2.0 in ./env/lib/python3.13/site-packages (from -r requirements-full-preview-generator.txt (line 16)) (0.2.0)
Requirement already satisfied: future==0.18.2 in ./env/lib/python3.13/site-packages (from -r requirements-full-preview-generator.txt (line 17)) (0.18.2)
ERROR: Could not find a version that satisfies the requirement rawpy==0.17.3 (from versions: 0.24.0, 0.25.0, 0.25.1, 0.26.0, 0.26.1, 0.27.0)
ERROR: No matching distribution found for rawpy==0.17.3
```

piRGoif:
Problems solved with dependencies : 


## requirements-build.txt

### pkgutil.ImpImporter removed in Python 3.13
Updated faulty libs to the first version without error under Python 3.13: pip, setuptools

## requirements.txt
### pillow locals() + exec() incompatibility in Python 3.13 as PEP 667 changed how exec writes to locals()
pillow 11.1.0 has official support see https://github.com/python-pillow/Pillow/releases/tag/11.0.0

#### weasyprint 63.0 requires Pillow < 11
In version 63.1 the upper bound was removed see https://github.com/Kozea/WeasyPrint/releases/tag/v63.1

### Two build failures, as they uses the private C API `_PyErr_WriteUnraisableMsg` which was removed in Python 3.13
Updated:
  - cffi https://github.com/python-cffi/cffi/releases/tag/v1.17.0 "Add support for Python 3.13"
  - lxml https://pypi.org/project/lxml/5.3.0/#files cp313 wheels present (CPython 3.13)

### `multidict==5.2.0` uses Py_TRASHCAN_SAFE_BEGIN/Py_TRASHCAN_SAFE_END which were removed in Python 3.13
https://github.com/aio-libs/multidict/releases/tag/v6.1.0 "Added support for Python 3.13"

### import cgi : module was removed in Python 3.13 (PEP 594)
  - webob : https://github.com/Pylons/webob/blob/1.8-branch/CHANGES.txt 1.8.9 added Python 3.13 support via the legacy-cgi package, 1.8.10 additionally has a security fix (open redirect in Python ≥3.10).
  - hapic : fixed by new webob version import (but as the version is incompatible with Python 3.13 we need more work !)

### zope.interface.exceptions.InvalidInterface: Concrete attribute, __firstlineno__. Python 3.13 added __firstlineno__ as an automatic class attribute, and old zope.interface treats it as an invalid concrete attribute on interface classes
zope-interface https://pypi.org/project/zope.interface/6.4.post2/ "Adjust for incompatible changes in Python 3.13b1"

### ConfigParser.readfp was removed in Python 3.12. paste.deploy at line 64 uses it as the getattr default — which Python evaluates eagerly, crashing even though read_file exists
pastedeploy https://docs.pylonsproject.org/projects/pastedeploy/en/latest/news.html "Fix a broken compatibility shim that would cause the ConfigParser to fail on Python 3.12 when ConfigParser.readfp is removed" + "Support Python 3.11 and 3.12"

### 'EntryPoints' object has no attribute 'items'
stevedore https://docs.openstack.org/stevedore/latest/user/history.html "Declare Python 3.12 support"

### matplotlib 3.11.0 requires pyparsing>=3, but you have pyparsing 2.4.7 which is incompatible.
transitive dep from preview-generator → vtk → matplotlib
cliff==3.8.0 requires only pyparsing>=2.1.0, no upper bound.
https://github.com/pyparsing/pyparsing/releases/tag/3.1.2 "Support for Python 3.13."

## requirements-full-preview-generator.txt
### no matching distribution error
  - vtk https://pypi.org/project/vtk/9.4.0/#files cp313 wheels
  - rawpy https://github.com/letmaik/rawpy/releases/tag/v0.24.0 "Add Python 3.13 wheels"
  - numpy https://github.com/numpy/numpy/releases/tag/v2.1.0 "Support for Python 3.13"

#### rawpy update
> Looking at all releases from 0.17.3 to 0.24.0, the only thing that could qualify as breaking is:
> - v0.22.0: "Add numpy 2 support (older versions are not supported anymore)" — dropped numpy 1.x
> - v0.23.0: reversed that — "Restore support for numpy 1.x, in addition to 2.x"
> 
> So with `rawpy==0.24.0`, both numpy 1.x and 2.x work. And since we're bumping numpy to 2.1.0 anyway, there's no conflict.
> 
> The APIs that preview-generator uses — rawpy.imread(), raw.postprocess() — are completely unchanged across all these versions. There are no breaking API changes relevant to Tracim.

piRGoif:
There is a WARN log remaining, but it is also happening with Python 3.11 so it is out of scope of this PR. I've added #6861

piRGoif:
Hapic needs to be updated to develop head (Damien worked on Python 3.13 compatibility with https://github.com/algoo/hapic/pull/219). 

But this version adds a new requirement: marshmallow 3.26.2. The dependency was already imported in Tracim with version 2.21.0, and used in Tracim backend code. The new version brings some depreciations / removals...

List of marshmallow modifications:

| Source                                                                                                                                                              | Version   | Description                                                                                            | Tracim status |
| ------------------------------------------------------------------------------------------------------------------------------------------------------------------- | --------- | ------------------------------------------------------------------------------------------------------ | ------------- |
| [Upgrading guide](https://marshmallow.readthedocs.io/en/3.x-line/upgrading.html#schemas-are-always-strict)                                                          | 2.x → 3.0 | `.dump().data` / `.load().errors` removed; methods now return data directly or raise `ValidationError` | **Fixed**     |
| [Email and URL fields should not validate on serialization · Issue #608 · marshmallow-code/marshmallow](https://github.com/marshmallow-code/marshmallow/issues/608) | 2.x → 3.0 | `ValidatedField` / `_validated` method removed                                                         | **Fixed**     |
| [3.x changelog](https://marshmallow.readthedocs.io/en/3.x-line/changelog.html#id59)                                                                                 | 3.0.0     | `Field.fail()` deprecated → `Field.make_error()`                                                       | _Not used_    |
| [2.x changelog](https://marshmallow.readthedocs.io/en/2.x-line/changelog.html#id14)                                                                                 | 2.19.0    | `fields.FormattedString` deprecated → `fields.Method` or `fields.Function`                             | _Not used_    |
| [3.x changelog](https://marshmallow.readthedocs.io/en/3.x-line/changelog.html#id47)                                                                                 | 3.3.0     | `fields.Nested('self', ...)` deprecated → use a callable instead                                       | _Not used_    |
| [3.x changelog](https://marshmallow.readthedocs.io/en/3.x-line/changelog.html#id40)                                                                                 | 3.7.0     | `marshmallow.pprint` deprecated                                                                        | _Not used_    |
| [3.x changelog](https://marshmallow.readthedocs.io/en/3.x-line/changelog.html#id35)                                                                                 | 3.10.0    | Passing metadata via kwargs (`example=`, `description=`) deprecated → `metadata={...}`                 | **Fixed**     |
| [3.x changelog](https://marshmallow.readthedocs.io/en/3.x-line/changelog.html#id29)                                                                                 | 3.13.0    | `missing=` / `default=` deprecated → `load_default=` / `dump_default=`                                 | **Fixed**     |
| [3.x changelog](https://marshmallow.readthedocs.io/en/3.x-line/changelog.html#id17)                                                                                 | 3.21.0    | `marshmallow.__version__` / `__parsed_version__` / `__version_info__` removed                          | _Not used_    |
| [3.x changelog](https://marshmallow.readthedocs.io/en/3.x-line/changelog.html#id8)                                                                                  | 3.24.0    | Validators returning `False` instead of raising `ValidationError`                                      | Not affected  |
| [3.x changelog](https://marshmallow.readthedocs.io/en/3.x-line/changelog.html#id8)                                                                                  | 3.24.0    | `context` parameter of Schema deprecated → `contextvars.ContextVar`                                    | _Not used_    |
| [3.x changelog](https://marshmallow.readthedocs.io/en/3.x-line/changelog.html#id8)                                                                                  | 3.24.0    | `Field`, `Mapping`, `Number` used directly as fields deprecated                                        | _Not used_    |
| [3.x changelog](https://marshmallow.readthedocs.io/en/3.x-line/changelog.html#id3)                                                                                  | 3.26.0    | `ordered` class Meta option deprecated                                                                 | _Not used_    |


About ValidatedField removal: 2 Tracim classes concerned: RFCEmail and TracimEmail      
The _validated methods weren't needed anymore:
- they were ensuring None value wouldn't trigger a validator error
- in marshmallow 3.0 the None value won't be checked based on the allow_none option (see ‎Field.deserialize)
- in all existing calls to these 2 classes without allow_none we have required=True

piRGoif:
Etude sur les lib dans requirements-test.txt : 

| Library | Pinned version | First Python 3.13 compatible version | Source |
|---|---|---|---|
| WebTest | 2.0.35 | Not mentioned in changelog | [changelog](https://docs.pylonsproject.org/projects/webtest/en/latest/changelog.html) |
| pytest | 7.2.2 | **8.2.1** (2024-05-19) — "Support for Python 3.13 (beta1 at the time of writing)" | [releases](https://github.com/pytest-dev/pytest/releases) |
| pytest-dotenv | 0.4.0 | Not mentioned — project archived Jul 2025 | [repo](https://github.com/quiqua/pytest-dotenv) |
| pytest-services | 2.0.1 | Not mentioned in changelog | [CHANGES.rst](https://github.com/pytest-dev/pytest-services/blob/main/CHANGES.rst) |
| pytest-testinfra | 6.8.0 | **10.2.0** (2025-03-29) — "Also run lint for py 3.12 and 3.13" | [CHANGELOG.rst](https://github.com/pytest-dev/pytest-testinfra/blob/main/CHANGELOG.rst) |
| pytest-timeout | 1.4.2 | **2.4.0** (2025-05-05) — "Add python 3.13" | [tags](https://github.com/pytest-dev/pytest-timeout/tags) |
| pytest-forked | 1.3.0 | Not mentioned — last explicit Python support: 3.11 (v1.5.0) | [CHANGELOG.rst](https://github.com/pytest-dev/pytest-forked/blob/master/CHANGELOG.rst) |
| responses | 0.10.11 | **0.25.7** (2024-03-11) — "Added support for python 3.13" | [releases](https://github.com/getsentry/responses/releases) |
| mock | 3.0.5 | Not mentioned in changelog | [CHANGELOG.rst](https://github.com/testing-cabal/mock/blob/master/CHANGELOG.rst) |
| freezegun | 0.3.12 | **1.5.2** — "Explicitly marks support for Python 3.13" | [CHANGELOG](https://github.com/spulec/freezegun/blob/master/CHANGELOG) |
| sseclient-py | 1.7 | Not mentioned in any release notes | [releases](https://github.com/mpetazzoni/sseclient/releases) |


=> as there are multiple big gaps, let's keep it as is and advised based on the auto test results


Idem pour requirements-devtool.txt :

| Library | Pinned version | First Python 3.13 compatible version | Source |
|---|---|---|---|
| gnureadline | *(unpinned)* | **8.2.13** (Oct 2024) — "Supports Python 3.13 (including free-threaded version)" | [tags](https://github.com/ludwigschwardt/python-gnureadline/tags) |
| flake8 | 6.0.0 | Not explicitly mentioned in any release note (7.3.0 added 3.14, 3.13 was never called out) | [release notes](https://flake8.pycqa.org/en/latest/release-notes/index.html) |
| isort | v5.10.1 | **6.0.0** (2024-01-27) — "Python 3.13 support" | [releases](https://github.com/PyCQA/isort/releases) |
| mypy | 1.9.0 | **1.12** (2024-10-14) — "Basic Support for Python 3.13 — This release adds partial support for Python 3.13 features and compiled binaries for Python 3.13." | [CHANGELOG.md](https://github.com/python/mypy/blob/master/CHANGELOG.md) |
| pre-commit | 2.18.1 | Not explicitly mentioned in changelog | [CHANGELOG.md](https://github.com/pre-commit/pre-commit/blob/main/CHANGELOG.md) |
| black | 23.3.0 | **24.10.0** (Oct 2024) — "officially tested with Python 3.13 and provides Python 3.13 mypyc-compiled wheels" | [CHANGES.md](https://github.com/psf/black/blob/main/CHANGES.md) |
| flake8-black | 0.3.6 | Not explicitly mentioned (latest: v4.0.0 Sep 2025, was dormant Dec 2022 → Sep 2025) | [tags](https://github.com/peterjc/flake8-black/tags) |
| flake8-isort | 6.0.0 | **6.1.2** (2025-01-29) — "Drop python 3.8, and add python 3.13" | [CHANGES.rst](https://github.com/gforcada/flake8-isort/blob/main/CHANGES.rst) |
| pip-tools | 7.4.1 | **7.5.2** (2025-11-11) — "Test and declare Python 3.13 support" | [CHANGELOG.md](https://github.com/jazzband/pip-tools/blob/main/CHANGELOG.md) |


=> doing the update, only local dev tools

- flake8 : update to latest 6.* (BC in 7.0.0)
- isort : seems like there aren't no impacting BC
- MyPy : changelog adds 3.13 mention up to 1.18 so updating to that version
- flake8-black : mention [in the README](https://github.com/peterjc/flake8-black#version-history) Python 3.13 support from v0.3.7
- pip-tools : BC in 7.5.0 : https://github.com/jazzband/pip-tools/blob/main/CHANGELOG.md#removals-and-backward-incompatible-breaking-changes-1

piRGoif:
Launching the test gives multiple lib errors:

## marshmallow 3 parameters keyword only

This is a change added in 3.0.0rc7

https://marshmallow.readthedocs.io/en/3.x-line/changelog.html#rc7-2019-06-15
_Backwards-incompatible_: Use keyword-only arguments ([[#1216](https://github.com/marshmallow-code/marshmallow/issues/1216)](https://github.com/marshmallow-code/marshmallow/issues/1216)).

Fix par maj backend/tracim_backend/views/core_api/schemas.py

## numpy => `np.float_` was removed in the NumPy 2.0 release. Use `np.float64` instead

Caused by a call from the elasticsearch lib

https://numpy.org/doc/stable/release/2.0.0-notes.html#numpy-2-0-python-api-removals

https://github.com/elastic/elasticsearch-py/releases/tag/v7.17.12
"Fix compatibility with NumPy 2.0"

## markdown => AttributeError: 'EntryPoints' object has no attribute 'get'

https://python-markdown.github.io/changelog/#342-2023-03-22
No mention to Python 3.13 in the changelog but only of 3.11 and 3.14
Python 3.11 => lib 3.4.2
Python 3.14 => lib 3.10.0

Tracim uses only markdown.markdown() — the top-level conversion function — with no extensions, no internal API, no deprecated objects. None of the breaking changes between 3.3.4 and 3.10.0 affect Tracim.


## freezegun => AttributeError: module 'uuid' has no attribute '_load_system_functions'

https://github.com/spulec/freezegun/blob/master/CHANGELOG
version 1.5.2

No hard breaking changes that would affect Tracim

piRGoif:
After fixing the blocking dependencies errors, launching the test gives : 

672 failed, 545 passed, 3 skipped, 10 xfailed, 572 errors in 804.23s (0:13:24)

lebouquetin:
@piRGoif reminder: I expect you to document the issue instead of the PR (nobody will look at the PR)

piRGoif:
> @piRGoif reminder: I expect you to document the issue instead of the PR (nobody will look at the PR)

Yes, I will report consolidated documentation to the issue at the dev end. I'm adding this step in the todo list

lebouquetin:
@piRGoif I pushed 2 commits on the branch:

- one which fix #6864 (we don't need a dedicated issue+branch+PR for this)
- one for API documentation improvment

lebouquetin:
@piRGoif I'm not sure … it looks like you removed zurl - https://github.com/tracim/tracim/commit/cf5bf4bbf17554d172c7e50e8fa622bba3e34ebb ? Is it intended? The commit message speak of uwsgi, not zurl (which is used for TLM and pushpin communication)

piRGoif:
> it looks like you removed zurl - [cf5bf4b](https://github.com/tracim/tracim/commit/cf5bf4bbf17554d172c7e50e8fa622bba3e34ebb) ? Is it intended?

This commit lack research to check for each modification validity, I'm working on it right now.

piRGoif:
Pushpin : there is no package in Debian 13 repo:
- https://tracker.debian.org/pkg/pushpin
- https://pushpin.org/docs/install/#debian-and-ubuntu

There are 3 ways to install: 
1. using the pushpin repo (meant for Ubuntu and not Debian)
2. downloading the .deb file
3. Debian SID

We chose option 1. 

As it cannot be installed simply using apt now, package was removed from `system_packages/debian/server_packages.list` and a note added in `system_packages/README.md`

piRGoif:
For TLM to work we also needed to update pubcontrol : we were requesting for 3.2.0, but in version 3.5.0 a fix was made to be compatible with Python >= 3.10

References:
- https://docs.python.org/3/whatsnew/changelog.html#changelog, Python 3.9.0 alpha 4: "bpo-39674: Revert “Do not expose abstract collection classes in the collections module” change (bpo-25988). Aliases to ABC like collections.Mapping are kept in Python 3.9 to ease transition from Python 2.7, but will be removed in Python 3.10."
- pubcontrol commit https://github.com/fanout/pypubcontrol/commit/85307323

piRGoif:
About the base images : 

- base debian:trixie VS debian:trixie-slim: 
  - slim get rid of man pages, documentation, has C locale only, some packages (apt-utils, readline-common, bsdutils), ...
  - just above 20MB gain => our image is far larger so no rush, keeping non slim version
- build_frontend node:16-bookworm: with Trixie we only have Node 22, 24, 26. So only updating to bookworm for now, and this souldn't affect anything as build_frontend is not used after build is done

piRGoif:
libmagic1 update: 
- package was there for:
  - bullseye (11): https://packages.debian.org/search?suite=bullseye&searchon=names&keywords=libmagic1
  - bookworm (12): https://packages.debian.org/search?suite=bookworm&searchon=names&keywords=libmagic1
- but replaced for trixie by libmagic1t64: https://packages.debian.org/search?suite=trixie&searchon=names&keywords=libmagic1

lebouquetin:
> just above 20MB gain => our image is far larger so no rush, keeping non slim version

This is pure optimisation, so you right: no need to go to slim for now.

> build_frontend node:16-bookworm: with Trixie we only have Node 22, 24, 26. So only updating to bookworm for now, and this souldn't affect anything as build_frontend is not used after build is done

I'm not sure what you mean. we do not want bookworm as the base image ; so you should use trixie + install a dedicated node 16 version.

@piRGoif

piRGoif:
> > build_frontend node:16-bookworm: with Trixie we only have Node 22, 24, 26. 
> 
> I'm not sure what you mean. we do not want bookworm as the base image

This image is used for the ~~build_backend~~ build_frontend stage only (here : https://github.com/tracim/tracim/blob/worktree-feat%2B6859_debian-13_python-3.13/tools_docker/Debian_Uwsgi/Dockerfile#L71).

 Final image relies on base stage, which use a trixie image.

lebouquetin:
> > >  build_frontend node:16-bookworm: with Trixie we only have Node 22, 24, 26.
> >  I'm not sure what you mean. we do not want bookworm as the base image
> This image is used for the build_backend stage only (here : https://github.com/tracim/tracim/blob/worktree-feat%2B6859_debian-13_python-3.13/tools_docker/Debian_Uwsgi/Dockerfile#L71).

You probably mean *build_frontend stage*.

Anyway, the goal is to ensure we know **how to build and run Tracim on a debian Trixie**.

Using Bookworm as base for frontend build will result in "not being able to run the build n a debian trixie" because you will adapt scripts and dependencies install to bookworm instead of trixie.

Setting up a manual install of the right version of node will ensure we can build manually the entire application on a development box.

Here is the way to install a dedicated version of nodejs on debian: https://nodejs.org/en/download
 
@piRGoif

piRGoif:
ok I understand, I will update the Dockerfile

piRGoif:
About mongrel / condure and zurl: both were pushpin dependencies. But both are replaced by the connmgr, which is condure merged into the project. The timeline extracted from the changelog (https://github.com/fastly/pushpin/blob/main/CHANGELOG.md):

- v1.31.0 (2020): Switched from Mongrel2 to Condure by default
- v1.37.0 (2023): Added ability to use Condure instead of Zurl for outgoing connections (opt-in, not removed)
- v1.38.0 (2024): Moved Condure into Pushpin as pushpin-condure
- v1.40.0 (2024): Renamed Condure to connmgr

The default config file only starts connmgr: https://github.com/fastly/pushpin/blob/v1.40.0/examples/config/pushpin.conf#L22    

The pushpin deps (https://packages.pushpin.org/deb/dists/pushpin-jammy/main/binary-amd64/Packages) no longer includes: 
- condure since 1.38
- zurl since 1.41

We have a pushpin config in `backend/pushpin_config/pushpin.conf` that is used by the backend docker-compose file (sets `PUSHPIN_CONFIG_DIR`). I've updated it to reflect current default pushpin config.    
Another config is in `backend/tracim_backend/tests/pushpin.conf` but don't seem to be used ?

lebouquetin:
> About condure and zurl: both were pushpin dependencies. But both are replaced by the connmgr, which is condure merged into the project.

Thanks for the explanation. Did you succeed in making TLM work on your dev environment?

piRGoif:
> Did you succeed in making TLM work on your dev environment?

Yes, locally with the pushpin container. I still have to generate and test a new Tracim image.

Also, I wasn't able to understand why TLM were working on my machine before... I made a lot of modifications in my Tracim backend venv, but Python 3.9 isn't present on my dev machine now (was it in the past ??)... 
But it was working when I was working in late june for #6852...

piRGoif:
About SETUPTOOLS_USE_DISTUTILS: explanations added in #6711

piRGoif:
Added PIP_BREAK_SYSTEM_PACKAGES env var : needed since Debian 12 to be able to process pip install without a venv. See [“Externally managed environments”: when PEP 668 breaks pip](https://pythonspeed.com/articles/externally-managed-environment-pep-668/)

piRGoif:
libapache2-mod-proxy-uwsgi package removal :

1. Apache 2.4.33 (2018): mod_proxy_uwsgi was added directly into Apache as a built-in submodule. The separate libapache2-mod-proxy-uwsgi package was turned into a transitional dummy pointing to apache2.
2. Bookworm (Debian 12): The transitional package was still shipped to give users time to update their package lists — but it had no content, just a dependency on apache2.
3. Apache 2.4.59-1 (April 2024 → Trixie): The transitional package was explicitly dropped per [[Debian bug #1032628](https://www.mail-archive.com/debian-apache@lists.debian.org/msg18425.html)](https://www.mail-archive.com/debian-apache@lists.debian.org/msg18425.html): "drop the transitional package after the release of bookworm, it has been released with buster and bullseye already." The package simply no longer exists in Trixie.

We are currently running Debian 11, which is bundled with Apache 2.4.67 (https://packages.debian.org/bullseye/apache2)

Package is present in system_packages/debian/server_packages.list. This file was created in commit 8f278bcb, where we also modified tools_docker/Debian_Uwsgi/Dockerfile : 
- switch hardcoded apt install to using server_packages.list file
- rename base_install to base
- base_install was debian:stretch, and base is debian:bullseye

Debian 9 Stretch was bundled with 2.4.25, so that would explain the presence of this package.

piRGoif:
Python package psycopg2 : it fails when building the image. 
We are asking for 2.9.9, but we can see in the changelog Python 3.13 compatibility was added in 3.9.10 : https://www.psycopg.org/docs/news.html#what-s-new-in-psycopg-2-9-10

piRGoif:
When running the image we have multiple warnings on container startup:

- Already present in previous images:
   - Deprecated config params (DEPOT_STORAGE_DIR, DEPOT_STORAGE_NAME, SMTP__USE_IMPLICIT_SSL) => created #6874
   - HTTP/HTTPS warnings => expected when running locally
   - Notification by email mechanism is disabled => expected
   - AH00558: Could not reliably determine the server's fully qualified domain name => cosmetic, needs ServerName in apache config
   - Restarting app server(s): uwsgi -> ! ! ! failed! => was already happening, doesn't seem to be an issue. entrypoint calls service uwsgi restart before uwsgi was ever started ?
   - subprocess 1263 exited with code 128 => was already happening, not investigated yet ?
- New logs:
   - init DB warnings : Document is empty / ink_file_open: cannot be opened => new warning after imagemagick update (from 6 to 7), stderr output when generating avatars on DB init (svg to jpg conversion), no impact as avatars are generated. Didn't find a solution to get rid of those errors :( A closed issue about that : https://github.com/ImageMagick/ImageMagick/issues/7462
   - Failed to wrap object of type 'PangoFT2FontMap' => no observed impact, referenced in open issue https://gitlab.com/inkscape/inkscape/-/work_items/5419

piRGoif:
software-properties-common : removed cause non existing in Trixie (see https://packages.debian.org/search?suite=all&searchon=names&keywords=software-properties-common).
No issue detected.

It seems it was used for pushpin, but it is now installed using curl + gpg.

piRGoif:
@lebouquetin and @alubert-algoo PR is ready for review, can you have a look ?

backend/requirements.txt
```diff
@@ -96,7 +96,7 @@ gitpython==3.1.18
     # via tracim_backend (setup.py)
 gripcontrol==4.1.0
     # via tracim_backend (setup.py)
-hapic[marshmallow]==0.95
+hapic[marshmallow] @ git+https://github.com/algoo/hapic.git@develop
```
lebouquetin: If it works well, then we just need to publish a stable release of hapic and to use this dependency.
Using a @develop dependency is broken by design (it is ok during development, but we must fix it now if tracim is workingwell

system_packages/debian/run_backend_packages.list
```diff
@@ -1,12 +1,12 @@
 ghostscript
 git
 imagemagick
-libmagickcore-6.q16-6-extra
+libmagickcore-?.q16-*-extra
```
lebouquetin: I suggest `libmagickcore-*-extra`. It would also includes `libmagickcore-7.q16hdri-10-extra` which is a series of extra codecs - I do not see a reason not to include it (except image size optimisation - but it is not the goal right now)

system_packages/README.md
```diff
@@ -8,3 +8,7 @@ This directory contains listing of system packages that can be used by Tracim.
 - `optional_preview_packages.list`: packages of external applications (inkscape, libreoffice…) that Tracim can use to preview contents.
 - `run_backend_packages.list`: minimal packages needed to run Tracim's backend.
 - `server_packages.list`: packages needed to run Tracim's backend with apache2 and uwsgi.
+
+> [!IMPORTANT]
+> The pushpin dependency must be installed manually, as currently there are no APT package for Debian 13 Trixie.
```
lebouquetin: Replace 

```
The pushpin dependency must be installed manually, as currently there are no APT package for Debian 13 Trixie.
```

With

```
Pushpin is not packaged in Debian 13 Trixie. You must add the official pushpin.org repo to source list in order to install pushpin.
```

tools_docker/Debian_Uwsgi/Dockerfile
```diff
@@ -42,14 +42,19 @@ RUN apt-get install -qy --no-install-recommends \
 # - installs optional dependencies
 ####
 FROM base AS build_backend
-ENV SETUPTOOLS_USE_DISTUTILS=stdlib
+# PIP_BREAK_SYSTEM_PACKAGES env var to workaround "This environment is externally managed" error
```
lebouquetin: I expect comments to start with a date, a type and initials. This should be in the Tracim coding rules (see https://algoo.tracim.fr/ui/workspaces/11/contents?folder_open=1194 )

Example:

```
# 2026-07-17 - INFO - D.A.
```

This helps maintain code and ask to the best coworker when there is a question.

Type of comment may be INFO / HACK / WARNING / TODO / FIXME

tools_docker/Debian_Uwsgi/Dockerfile
```diff
@@ -42,14 +42,19 @@ RUN apt-get install -qy --no-install-recommends \
...
 RUN apt-get install -qy $(cat "${PACKAGE_DIR}/build_backend_packages.list") \
-    # Go in backend folder
+    # Go in backend folder \
     && cd /tracim/backend \
-    # Install backend
-    && pip install --force-reinstall \
+    #--- pip install section \
+    # With Debian 12 Bullseye we were using `--force-reinstall` to force installing our Python packages instead of the system Python packages \
+    # This is no longer possible with Debian 13 Trixie : apt-installed Python packages are shipped without RECORD files (PEP 668) so pip cannot uninstall them \
+    # instead we are using `--ignore-installed` which writes everything fresh into `/usr/local/ instead`. \
```
lebouquetin, line 57: same remark: add a comment line with date, type and initials

tools_docker/Debian_Uwsgi/Dockerfile
```diff
@@ -68,10 +73,20 @@ RUN cp -r /tracim/frontend/dist/assets/font/Nunito/ /usr/share/fonts/nunito
...
+FROM debian:trixie AS build_frontend
 COPY --from=base /tracim /tracim
+# Install Node.js 16 via nvm (not in Trixie official repos)
+ENV NVM_DIR=/root/.nvm
+RUN apt-get update \
+    && apt-get install -qy curl bash \
+    && curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.5/install.sh | bash \
+    && . "$NVM_DIR/nvm.sh" \
+    && nvm install v16.20.2
+# Add nvm node 16 bin dir to PATH so all node binaries (node, npm, npx, yarn, corepack…) are found
```
lebouquetin: I think this comment is not usefull

tools_docker/Debian_Uwsgi/Dockerfile
```diff
@@ -68,10 +73,20 @@ RUN cp -r /tracim/frontend/dist/assets/font/Nunito/ /usr/share/fonts/nunito
 # - setups backend translation
 # - builds frontend
 ####
-FROM node:16-buster AS build_frontend
+FROM debian:trixie AS build_frontend
 COPY --from=base /tracim /tracim
+# Install Node.js 16 via nvm (not in Trixie official repos)
```
lebouquetin: > `# Install Node.js 16 via nvm (not in Trixie official repos)`

```
# Install Node.js 16 via nvm (nodejs is available in Trixie but this is not the v16 - which is required for Tracim build)
```

tools_docker/Debian_Uwsgi/Dockerfile
```diff
@@ -116,6 +134,12 @@ COPY --from=build_backend /tracim /tracim
 COPY --from=build_backend /usr/local /usr/local
 # backend font to generate avatars
 COPY --from=build_backend /usr/share/fonts/nunito /usr/share/fonts/nunito
+# Pre-build system font cache and create www-data's per-user cache/config dirs.
```
lebouquetin: Add

```
# INFO - 2026-07-20 - P.G.
```

tools_docker/pushpin/Dockerfile
```diff
@@ -9,14 +9,19 @@
 # works with mongrel2.
 
 # Pull the base image
-FROM debian:bullseye
+FROM debian:trixie-slim
```
lebouquetin: I do not understand. We add a discussion about slim or not slim and I told you explicitly not to use slim: the goal of this issue is to base tracim on debian, not to optimize it.

Especially: it removes man pages which are used by sysadmins. So at the moment, I prefer to keep `debian:trixie` and not `debian:trixie-slim`

lebouquetin (changes requested):

backend/requirements.txt
lebouquetin: I will publish a 1.0 release of hapic, so you can update the dependency

system_packages/debian/run_backend_packages.list
piRGoif: Indeed libmagickcore-*-extra will match (`apt-cache pkgnames 'libmagickcore-*-extra'`) : 
- libmagickcore-7.q16-10-extra ← what we need
- libmagickcore-7.q16hdri-10-extra

Switching to that, updating doc

tools_docker/Debian_Uwsgi/Dockerfile
piRGoif: Fixed

piRGoif: Shouldn't we move the coding rules to the Tracim repo doc folder ?

tools_docker/Debian_Uwsgi/Dockerfile
piRGoif: Fixed

tools_docker/Debian_Uwsgi/Dockerfile
piRGoif: Fixed

tools_docker/Debian_Uwsgi/Dockerfile
piRGoif: Fixed

tools_docker/Debian_Uwsgi/Dockerfile
lebouquetin: > Shouldn't we move the coding rules to the Tracim repo doc folder ?

We may ; we didn't for some reason at the moment:

- coding rules are for almost all our projects, not tracim-only
- documentation in the tracim repo is not ready to use - moving it now is not a good idea

tools_docker/pushpin/Dockerfile
piRGoif: We talked about the uwsgi Dockerfile, and I thought this pushpin Dockerfile is for dev so I kept the slim variant there. 

Removed.

lebouquetin:
By the way, this is a nice move to use raw strings: make things clear + remove test-purposed comments (which are now useless) :+1:

tools_docker/Debian_Uwsgi/Dockerfile
lebouquetin: This is a real challenge (I would be happy to have it in tracim repo - but this would result in duplicated data …) . To discuss on "how to manage this"

tools_docker/pushpin/Dockerfile
lebouquetin: >We talked about the uwsgi Dockerfile, 

You right

> and I thought this pushpin Dockerfile is for dev so I kept the slim variant there.

Hm … I misunderstood the docker use.

Nevertheless, I think optimising too early is generaly not a good idea (resulting in troubleshooting that start with apt install <some tools>...)

backend/requirements-devtool.txt
```diff
@@ -1,4 +1,5 @@
 # Devtools
+gnureadline
```
alubert-algoo: I am really surprised this module is necessary to run the application, since readline is already available on the system and manage completion since a very long time.

tools_docker/Debian_Uwsgi/Dockerfile
```diff
@@ -11,7 +11,7 @@ ARG CONTAINER_TRACIM_DIR="/tracim"
 # Base image for installing some base dependencies
 ####
 
-FROM debian:trixie-slim AS base
+FROM debian:trixie AS base
```
alubert-algoo, line 14: A reason to switch back to the non-slim version? There were missing packages?

tools_docker/Debian_Uwsgi/Dockerfile
```diff
@@ -73,10 +73,20 @@ RUN cp -r /tracim/frontend/dist/assets/font/Nunito/ /usr/share/fonts/nunito
...
+    && curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.5/install.sh | bash \
+    && . "$NVM_DIR/nvm.sh" \
+    && nvm install 16
+# Add nvm node 16 bin dir to PATH so all node binaries (node, npm, npx, yarn, corepack…) are found
+ENV PATH="$NVM_DIR/versions/node/v16.20.2/bin:$PATH"
 RUN \
-    cd /tracim/backend \
+    npm install -g yarn \
+    && cd /tracim/backend \
     && npm install "i18next-conv@<8" -g \
```
alubert-algoo: I am not sure this is the best way to work with Node v16 in this case. This is a layer, so the final image will not have the nvm installation and that a good choice.

But, in this situation, nvm will not be updated until we change the hard-coded version.

I know that the node:16 is not available with trixie, but maybe this change must occurs during the js modules upgrading?

tools_docker/Debian_Uwsgi/Dockerfile
lebouquetin: Yes: we upgrade from old stable to stable - there is no reason to go to slim at the same time.

We run into a lack of tools (vim, manpages, etc) more often than a lack of space. Building on top of slim image is an over-optimization and of-topic *at this time*: this is not required to upgrade from bookworm to trixie.

tools_docker/Debian_Uwsgi/Dockerfile
lebouquetin: I'm not sure what you mean.

The goal is to have a standard way to setup Tracim, through a docker container or through raw installation on a debian Trixie.

Basing the docker build process on another image may be more efficient for docker container build but it would be a distinct process than installing it on a raw trixie linux, resulting in potential specific debugging, different build behavior, etc.

lebouquetin: Also, it's not clear what your suggest. I would be interested in explanation of what you would have done here

alubert-algoo: In this modification, both the version of nvm and nodejs will be pinned. So, we need to check these versions from time to time for releases or security hotfixes. It is just an additional maintenance for us.

But since the version of node we use in Tracim is no longer available with a modern Debian image, I do not see another method to ensure the usage of node 16 on Debian 13.

Later, if we upgrade the JS dependencies to be fully functional with the latest version of node, I think it will be better to use the official node images with the trixie flavor to have less maintenance, instead of manually install nodejs.

lebouquetin: > Later, if we upgrade the JS dependencies to be fully functional with the latest version of node, I think it will be better to use the official node images with the trixie flavor to have less maintenance, instead of manually install nodejs.

An intersting thing here is **with the trixie flavor**. I asked @piRGoif not to use a dedicated image because it was bookworm-based.

Another point is : if we build from nodejs image, I feel like we loose the knowledge on "how to setup tracim on a raw debian trixie". Do you agree on that?

backend/requirements-devtool.txt
lebouquetin: let's remove gnureadline and re-add it if there are errors

tools_docker/Debian_Uwsgi/Dockerfile
alubert-algoo: Yes I agree. And if we really want to fully go to the Debian-only environment, the best way is to switch to the current Debian nodejs (v20) package when the JS modules are fully compatible.

backend/requirements-devtool.txt
piRGoif: See note about requirements-devtool.txt in https://github.com/tracim/tracim/issues/6859#issuecomment-5024991638 : if absent some warnings are emitted in the server log on startup

This deps file is intended to be installed on a dev environment only

tools_docker/Debian_Uwsgi/Dockerfile
piRGoif: I understand the concern about nvm fixed version. On https://github.com/nvm-sh/nvm install methods all expose a pinned version :/ @alubert-algoo do you have anything in mind ?

backend/requirements.txt
lebouquetin: @piRGoif hapic 1.0 is available - see https://pypi.org/project/hapic/

tools_docker/Debian_Uwsgi/Dockerfile
piRGoif: Discussed with Aurelien : we don't know of a way to keep this install method without depending on the whole nvm version.
Dependabot doesn't check for Dockerfile (see https://docs.github.com/en/code-security/reference/supply-chain-security/dependency-graph-supported-package-ecosystems#supported-package-ecosystems)

So yes indeed this is just temporary situation, we should rely on an official Debian Node image in the near future.

backend/requirements-devtool.txt
piRGoif: Discussed with Aurelien : adding a dependency only to solve one warning is too much. Removing the dependency.

tools_docker/Debian_Uwsgi/Dockerfile
lebouquetin: > So yes indeed this is just temporary situation, we should rely on an official Debian Node image in the near future.

This is not so easy - as I explained we need to know how to setup the full environment on a raw Debian distribution. 

By the way, building for production requires to pin versions so we ensure reproductible images. What I understand of your point is "we will get updates automatically" which make difficult to ensure we can reproduce an image (and a bug or unexpected behavior)

lebouquetin: I mark unresolved in order to ensure you read my comment. Then, lets resolve it again

piRGoif: Understood. We will at least need to be notified of nvm vulnerabilities.

docs/requirements.md
```diff
@@ -0,0 +1,13 @@
+# Tracim Compatibility
+
+## Python & OS versions
+
+> [!IMPORTANT]
+> We recommend running on Debian 13 with Python 3.11.
```
lebouquetin: Python 3.13 @piRGoif

lebouquetin: By the way, Why do you move the compatibilty to another file for less than 10 lines of explanation? In my point of view, it makes documentation harder to read because you need to clic to get extra (but small) information

piRGoif: Woops, thanks

About the new file : 
- previously requirements were only in docs.legacy/administration/installation/install_backend.md which is in the old directory
- I though adding it to docs/developer/setup_env/getting_started.md was a good idea
- I imagine there are other files where we need to add a mention to requirements
- having requirements at the doc directory root is nice I think ?

lebouquetin: Let's keep it like you did.

My primary goal is the new documentation match with the preliminary work - https://algoo.tracim.fr/ui/contents/155966

Issue #6876, opened 2026-07-22
✨ Feat: Improve the Kanban and Gantt


alubert-algoo:
#### required

- [x] Ce contenu a été modifié par quelqu’un. -> Ce contenu a été modifié.
- [x] Internationalize date format . example: `2026-08-10 → 2026-08-13` become in french `10/08/2026 → 13/08/2026`
- [x] Add standard margin around the Gantt view
- [x] In breadcrumb: replace `Gantt` with `Gantt view`
- [x] Allow to delete a dependency even if the dependency is in another column
- [x] Rules for gantt view (ie do not store the data but fill the json before to load gantt view)
  - [x] if only start date is filled; compute end date as `end date = start date + duration`
  - [x] if only end date is filled; compute start date as `start date = end date - duration`
  - [x] if start and end date date filled: ignore duration
  - [x] dependencies:
    - [x] if no start date defined, start date is computed from end date of all dependencies (ie the max end date)
    - [ ] bonus: if not complicated - if someone is affected to several cartes, then serialize cards.

#### important 

- [x] Do not set a default duration. A default duration of 1 day must be used by Gantt, BUT the value must not be stored. This will allow to hide duration if it is useless (case: all kanban that are not projects)
- [x] In the kanban view, do not show properties if no information is set. Examples:
  - [x] no duration set results in no duration shown
  - [x] no progress set results in no progress shown
  - [x] etc
- [x] dependency selection: a color circle left to the task title would help selecting right dependencies (same circle as the one shown on the read-only view of cards)
- [x] The dependencies of a card are related of the card column. 
  - [x] If we move a card in another column, this card will still show the dependencies, even if they are not available in the card editor.

#### Nice to have (may be postponed)

- [x] allow dependencies from other columns (in this case, select will show a first set of dependencies from current column, then a separator and dependencies from other columns)
- [x] make card edition consistant with view mode:
  - [x] dates are horizontal in view mode while vertical in edit mode
  - [x] progress and duration are not in an intuitive order
  - [x] 100% should probably be replaced by a "done" + check icon

#### Postponed (move to another issue)

- [ ] The dependencies cannot be easily sorted in the card editor.
  - [ ] To resolve this point, a new component must be created to manage the dependencies and allow to reorder them.
- [ ] The duration and advancement fields use TextInput to set their respective values, but having an NumberInput component can improve the interaction with the user.
[...]

Issue #6870, opened 2026-07-13
✨ Feat: Enrich kanban cards properties


lebouquetin:
### Feature description

In the goal to really help project management usage of Tracim, we must implement a gantt view (see #6867 meta issue) and this requires card extra properties.

Pre-requirements: #6869 

New expected properties:

- progress bar in percents. This will allow users to define the work progress state.
- Duration (in days). It is require to allow to build gantt dependency graph.
  - Authorized values must be float (like 1.5 day)
  - values used for gantt renderign must be rounded to higher it value (1.5 day -> 2 days)
  - default value should be 1 in the user interface
- dependencies: link to other cards to say « prerequired: finish these tasks ». 
  - The behavior should be a multi-select field (stored as a json list) ; the rendering is probably better as a text (1 issue per line), references based on uuid 
  - UI should show the card labels (not the UUID)
  - For the first release, dependency choose list must show first the cards of current column, then cars from other columns. A separator would help, like the one we use to select a template from current space vs other spaces
- A checkbox property labelled « closed » / « finished ». This status will override the 100% progress meaning « we dont care the real content of the card: now the task is finished »
  - default value: uncheck

### How it works

Adding properties is almost straight as the card structure is a schemaless JSON structure.

alubert-algoo:
The Kanban and the Card inside have been updated by the PR #6877 to add the wanted properties.

First, the Kanban board will show on each card the information related to Gantt:

<img width="600" height="436" alt="Image" src="https://github.com/user-attachments/assets/18b44d1c-036f-4712-9436-8351209fb62a" />

Under the start and due dates, two new fields have been added to show the progression (in percent) and the duration of the task (in days).

The list of dependencies to this card will be shown under by showing the title of the card and the color choose in the card.

<img width="600" height="458" alt="Image" src="https://github.com/user-attachments/assets/adfd0420-94c0-417f-ab46-2a01dcaea9f0" />

In the card editor, the new fields have been added next to the start and due dates. The progression can be marked as complete by clicking on the “Finished” button. 

<img width="449" height="49" alt="Image" src="https://github.com/user-attachments/assets/02ef57ae-eaf7-47de-8db6-506fba294d1d" />

When the button was clicked, the progression is shown as 100% and another button will be shown to reopen the task if necessary.

The list of dependencies are based on the current column and will shown the title of each cards inside.

[...]

PR #6877, opened 2026-07-22
feat(front): Enrich the card properties of the Kanban

alubert-algoo:
This PR add the new fields related to Gantt to the card from the Kanban view as specified by the issue #6870.

Some issues were spotted during the development and will be done in the issue #6876.


frontend_app_kanban/i18next.scanner/en/translation.json
```diff
@@ -56,5 +56,16 @@
   "Due date": "Due date",
   "Title": "Title",
   "Open field": "Open field",
-  "Assignment": "Assignment"
+  "Assignment": "Assignment",
+  "Advancement:": "Advancement:",
```
lebouquetin: `Progress` or `Progress status` would be better, I think

frontend_app_kanban/i18next.scanner/en/translation.json
```diff
@@ -56,5 +56,16 @@
...
+  "day": "day",
+  "days": "days",
+  "{{count}} day": "{{count}} day",
+  "{{count}} days": "{{count}} days",
+  "Dependencies:": "Dependencies:",
```
lebouquetin: `Depends on` would be better, I think

frontend_app_kanban/i18next.scanner/fr/translation.json
```diff
@@ -56,5 +56,16 @@
...
   "Title": "Titre",
   "Open field": "Champ libre",
-  "Assignment": "Affectation"
+  "Assignment": "Affectation",
+  "Advancement:": "Avancement :",
+  "day": "jour",
+  "days": "jours",
+  "{{count}} day": "{{count}} jour",
+  "{{count}} days": "{{count}} jours",
+  "Dependencies:": "Dépendances :",
```
lebouquetin: `Dépend de` or `Bloqué par` seems better, I think

frontend_app_kanban/i18next.scanner/en/translation.json
alubert-algoo: This key was incorrect since the trailing colon was not used. The final key is `Dependencies`.

frontend_app_kanban/i18next.scanner/fr/translation.json
alubert-algoo: Same as the english file, this is not the correct key.

frontend_app_kanban/i18next.scanner/en/translation.json
lebouquetin: I'm not sure to understand your answer. My point is to change the wording shown to the user.

Actually, I expect the UI to show `Depends on` or `Blocked by` instead of `Dependencies`

frontend_app_kanban/i18next.scanner/fr/translation.json
lebouquetin: > Same as the english file, this is not the correct key.

Same point as previous one: My feedback is about UI wording

frontend_app_kanban/i18next.scanner/en/translation.json
lebouquetin: I reopen this because in French `Avancement` is better while in English `Progress` is better than `Advancement` :-p

frontend_app_kanban/i18next.scanner/en/translation.json
alubert-algoo: It is because the behavior of this label is not the same as shown in this picture:
depends

In this case, the “Dependencies” was used as a placeholder and not like a label. We can add more information in this one, since it is inside the widget.

lebouquetin: The user need to see something like Depends on or Blocked By. 

Showing only `Dependancies` is not clear about the dependency direction.

Why not adding a label so that things are explicit for the end user?

alubert-algoo: Something like that?
depends-2

lebouquetin: This would be great, yes 👍

PR #6887, opened 2026-07-28
Fix UserConnectionStateMonitor broken due to a Python 3.11 change

### LINKED ISSUE #6865: Fix multiple TestLiveMessages failing with Python 3.11
For example : 

```
  ______________________________________ TestLiveMessages.test_api__user_live_messages_endpoint_with_GRIP_proxy__ok__favorite__async[sqlite] ______________________________________

  self = <backend.tracim_backend.tests.functional.test_live_messages.TestLiveMessages object at 0x7f06688e4a50>
  small_html_document_a = <tracim_backend.models.data.Content object at 0x7f065e2b9150>, pushpin = None, app_config = <tracim_backend.config.CFG object at 0x7f065e537b90>
  rq_database_worker = <Popen: returncode: None args: ['rq', 'worker', '-q', '-w', 'tracim_backend....>

      @pytest.mark.pushpin
      def test_api__user_live_messages_endpoint_with_GRIP_proxy__ok__favorite__async(
          self,
          small_html_document_a: Content,
          pushpin,
          app_config,
          rq_database_worker,
      ):
          with messages_stream_client() as client_events:
              add_favorite_request = requests.post(
                  "http://localhost:7999/api/users/1/favorite-contents",
                  auth=("admin@admin.admin", "admin@admin.admin"),
                  json={"content_id": small_html_document_a.content_id},
              )
              assert add_favorite_request.status_code == 200
  >           favorite_created_event = next(client_events)

  tracim_backend/tests/functional/test_live_messages.py:531:
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
  env/lib/python3.11/site-packages/sseclient/__init__.py:58: in events
      for chunk in self._read():
  env/lib/python3.11/site-packages/sseclient/__init__.py:48: in _read
      for chunk in self._event_source:
  env/lib/python3.11/site-packages/requests/models.py:753: in generate
      for chunk in self.raw.stream(chunk_size, decode_content=True):
  env/lib/python3.11/site-packages/urllib3/response.py:572: in stream
      for line in self.read_chunked(amt, decode_content=decode_content):
  env/lib/python3.11/site-packages/urllib3/response.py:764: in read_chunked
      self._update_chunk_length()
  env/lib/python3.11/site-packages/urllib3/response.py:694: in _update_chunk_length
      line = self._fp.fp.readline()
  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

  self = <socket.SocketIO object at 0x7f065e0dec80>, b = <memory at 0x7f065e436440>

      def readinto(self, b):
          """Read up to len(b) bytes into the writable buffer *b* and return
          the number of bytes read.  If the socket is non-blocking and no bytes
          are available, None is returned.

          If *b* is non-empty, a 0 return value indicates that the connection
          was shutdown at the other end.
          """
          self._checkClosed()
          self._checkReadable()
          if self._timeout_occurred:
              raise OSError("cannot read from timed out object")
          while True:
              try:
  >               return self._sock.recv_into(b)
  E               Failed: Timeout >30.0s

  ../../../../.pyenv/versions/3.11.14/lib/python3.11/socket.py:718: Failed
```

Real root cause: pubcontrol/utilities.py:47,67 (a transitive dependency of gripcontrol, used by LiveMessagesLib) does:
if isinstance(value, collections.Mapping):
collections.Mapping/collections.Iterable were deprecated aliases for collections.abc.Mapping/collections.abc.Iterable, and were removed entirely in Python 3.10. Your venv runs Python 3.11.14. Every single live-message publish call hits this AttributeError, gets wrapped as ValueError, and — since these tests mostly run in non-blocking mode — is silently swallowed, so the SSE client just waits forever for an event that will never arrive. This is a pubcontrol package compatibility bug with modern Python, unrelated to Pushpin, ZMQ, or IPv6/hostname resolution at all.

On CI these tests aren't part of the "backend-quick-tests" group therefore are only ran with 1 Python version.

Unfortunately pypubcontrol doesn't provide any changelog :( (nor [on Pypi](https://pypi.org/project/pubcontrol/) nor on [its Github repo](https://github.com/fanout/pypubcontrol))

piRGoif:
Fixes #6865 

Issue was causing TestLiveMessages.test_api__user_live_messages_endpoint_with_GRIP_proxy__err__too_many_online_users[sqlite] to fail

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### 🐛 Fixed Issues
 
 - Fix error when using "Download as PDF" action since 2026.07.00 [#6888](https://github.com/tracim/tracim/issues/6888)
-- Fix UserConnectionStateMonitor that was failing cause to `Enum.__format__()` change in Python 3.11
+- Fix UserConnectionStateMonitor that was failing cause to `Enum.__format__()` change in Python 3.11 [#6865](https://github.com/tracim/tracim/issues/6865)
 
 
 ## 2026.07.00
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### 🐛 Fixed Issues
 
 - Fix error when using "Download as PDF" action since 2026.07.00 [#6888](https://github.com/tracim/tracim/issues/6888)
+- Fix UserConnectionStateMonitor that was failing cause to `Enum.__format__()` change in Python 3.11
 
 
 ## 2026.07.00
--- a/backend/tracim_backend/lib/user_connection_state_monitor/monitor.py
+++ b/backend/tracim_backend/lib/user_connection_state_monitor/monitor.py
@@ -74,7 +74,7 @@ class UserConnectionStateMonitor:
         transaction.commit()
         logger.debug(
             self,
-            "Set connection status of user {} to {}".format(user_id or "*", status),
+            "Set connection status of user {} to {}".format(user_id or "*", status.value),
         )
 
     def add_to_pending_offline_users(self, user_id: int) -> None:


piRGoif:
Fix : in Python 3.11 `Enum.__format__()` behavior was changed, see https://docs.python.org/3/whatsnew/3.11.html#enum
IN monitor.py we were getting "online" and since Python 3.11 this same code returns "UserConnectionStatus.ONLINE"

alubert-algoo (approved):

PR #6889, opened 2026-07-28
Fix download as PDF

### LINKED ISSUE #6888: 🐛 Bug: Cannot download as PDF since Tracim 2026.07
### Tracim version

2026.07.00

### Issue description

Download as PDF a note document : a 400 error is thrown to the browser

💥 This is a regression caused by the marshmallow update in Tracim 2026.07.00 (#6859)

### How to reproduce

1. Open a note
2. In the contextual menu select "Download as PDF"

### Screenshots

_No response_

Firefox

Desktop

### Related PR

_No response_

piRGoif:
Fixes #6888

### DIFF
--- a/backend/tracim_backend/models/context_models.py
+++ b/backend/tracim_backend/models/context_models.py
@@ -609,7 +609,7 @@ class FileQuery(object):
 
     def __init__(self, force_download: int = 0, revision_id: int = None) -> None:
         self.force_download = force_download
-        # INFO - revision_id is only used by clients as a cache-busting value, it is not
+        # INFO - PGO - 2026-07-28 - clients only use revision_id as a cache-busting value, it is not
         # used server-side.
         self.revision_id = revision_id
 
@@ -622,7 +622,7 @@ class PageQuery(object):
     def __init__(self, force_download: int = 0, page: int = 1, revision_id: int = None) -> None:
         self.force_download = force_download
         self.page = page
-        # INFO - revision_id is only used by clients as a cache-busting value, it is not
+        # INFO - PGO - 2026-07-28 - clients only use revision_id as a cache-busting value, it is not
         # used server-side.
         self.revision_id = revision_id
 
@@ -864,7 +864,7 @@ class TranslationQuery:
         self.source_language_code = source_language_code
         self.target_language_code = target_language_code
         self.force_download = force_download
-        # INFO - revision_id is only used by clients as a cache-busting value, it is not
+        # INFO - PGO - 2026-07-28 - clients only use revision_id as a cache-busting value, it is not
         # used server-side.
         self.revision_id = revision_id
 
--- a/backend/tracim_backend/models/context_models.py
+++ b/backend/tracim_backend/models/context_models.py
@@ -607,7 +607,7 @@ class FileQuery(object):
     File query model
     """
 
-    def __init__(self, force_download: int = 0, revision_id: int = None) -> None:
+    def __init__(self, force_download: int = 0, revision_id: int | None = None) -> None:
         self.force_download = force_download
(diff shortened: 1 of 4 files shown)


backend/tracim_backend/models/context_models.py
```diff
@@ -607,18 +607,24 @@ class FileQuery(object):
     File query model
     """
 
-    def __init__(self, force_download: int = 0) -> None:
+    def __init__(self, force_download: int = 0, revision_id: int = None) -> None:
```
alubert-algoo: A reason to use `None` as value of the `revision_id` attribute since this is integer?

If this value is valid, you can use the union type `int | None` instead (since Python 3.10).

lebouquetin (commented):
As discussed, just need to check if the revision is not used or if it is only used for older revisions (like if you download default , then it takes the latests while the revision_id is used for other cases - when you download a given revision)

After this check (and eventually update of the comment), it is ok for me

backend/tracim_backend/models/context_models.py
lebouquetin: maybe set revision_id to 0 would be better in terms of consistency? (int or None is more complexe than int)

piRGoif: I like the idea that None means "not provided", but I'm unsure of the Tracim code habits ? What do you think @lebouquetin ?

piRGoif:
> just need to check if the revision is not used or if it is only used for older revisions

Indeed we are setting revision_id param value to the note revision : it is done in frontend_app_html-document/src/container/HtmlDocument.jsx L602 in the getDownloadPDFUrl function

I removed my comments

backend/tracim_backend/models/context_models.py
lebouquetin: let met check data in production and source code …

edit: it looks like we already use null value for parent_id of contents that have no parent.

In general, I like the id to keep type consistant - like setting 0 for "no id" - this is also safe because 0 is a falsy value in python. BUT: 

In core.models.data, we have this:

```
id = Column(Integer, Sequence("seq__content__id"), autoincrement=True, primary_key=True)
cached_revision_id = Column(
    Integer, ForeignKey("content_revisions.revision_id", ondelete="RESTRICT")
)  # this is the revision_id actually

[...]
workspace_id = Column(
        Integer, ForeignKey("workspaces.workspace_id"), unique=False, nullable=True
    )
    workspace = relationship("Workspace", remote_side=[Workspace.workspace_id])

    parent_id = Column(Integer, ForeignKey("content.id"), nullable=True, default=None)
```

So there is no clear rule. Let allow None as a value in the API then ensure in lower layers that we won't query the DB with a None

Note: in DB we have content revisions with null as parent_id.

lebouquetin: Most important @piRGoif: as you are working for an API, if you allow None, you should also allow 0 for API consistency: the user will set 0 or 1 for force_download, so the intuition will be to put 0 for "no revision_id"

piRGoif: Ok so keeping None, I've added the type hinting modification suggested by Aurelien

force_download seems more like a boolean value ?

piRGoif:
Note : I've tested with ODT and text file, both are working OK.
An image file has no download as pdf option.

No download as pdf option either for logbook, kanban and topic.

backend/tracim_backend/models/context_models.py
lebouquetin: > force_download seems more like a boolean value ?

Yes but if you have a boolean in the API and you set force_download=0 for a falsy value, you can intuite that "no revision" is 0

PR #6892, opened 2026-07-30
Fix connmgr.log perms

### LINKED ISSUE #6891: 🐛 Bug: After some time the "Tracim has a connection problem" error is always displayed even if Tracim is working correctly (except TLM)
### Tracim version

2026.07.00

### Issue description

The "Tracim has a connection problem" is displayed a few seconds after login. Every Tracim screen is working as expected, but TLM are not working.

### How to reproduce

Simply login

### Investigation

This appears after the Tracim 2026.07.00 upgrade and a container restart : everything was fine on first start, but after restart the problem appears.

When the problem arise in the container pushpin is working but no 7999 port is open. No suspect log.

We see though that when launching pushpin (`pushpin` command) with root we have no issue and a start log is added in `/var/tracim/logs/pushpin/connmgr.log`.
This file is 644 permission and www-data:www-data owner

If we start pushpin with the pushpin user we get an error when starting connmgr. Both other services (proxy, handler) are started well.

💥 This is a regression caused by the Pushpin update in Tracim 2026.07.00 (#6859)

piRGoif:
Fixes #6891 

Solve permissions issues with `/var/tracim/logs/pushpin/connmgr.log`

This file is a new one : Since 2026.07.00 we upgraded Pushpin and change its installation method. Condure + Zurl are replaced by connmgr (see #6859) which generates a new log file.

Seems like if the container is restarted we lost permissions on this file and so connmgr crashes on start.

### DIFF
--- a/tools_docker/Debian_Uwsgi/common.sh
+++ b/tools_docker/Debian_Uwsgi/common.sh
@@ -143,7 +143,7 @@ if [ ! -d /etc/tracim/custom_toolbox ]; then
     mkdir /etc/tracim/custom_toolbox -p
 fi
 
-# Create folder and logs file
+# Create folder and logs file (first install)
 if [ ! -d /var/tracim/logs ]; then
     mkdir /var/tracim/logs -p
     mkdir /var/tracim/logs/redis -p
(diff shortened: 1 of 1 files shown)


lebouquetin:
It's good for me. well done

PR #6896, opened 2026-07-31
Fix cannot edit in Collabora since 2026.07.00

### LINKED ISSUE #6895: 🐛 Bug: Cannot edit with Collabora Online since 2026.07.00
### Tracim version

2026.07.00

### Issue description

Since Tracim 2026.07.00 we cannot edit office documents with Collabora Online

### How to reproduce

Edit an existing office document or create a new one

Collabora loads but displays a "Failed to read document from storage" error

💥 This is a regression from the marshmallow update in Tracim 2026.07.00 (#6859)

### Screenshots

Image

_No response_

_No response_

### Related PR

_No response_

piRGoif:
Fixes #6895

### DIFF
--- a/backend/tracim_backend/applications/collaborative_document_edition/wopi/schema.py
+++ b/backend/tracim_backend/applications/collaborative_document_edition/wopi/schema.py
@@ -27,6 +27,13 @@ class WOPITokenQuerySchema(marshmallow.Schema):
     access_token = marshmallow.fields.String(
         required=True, metadata={"description": "Access token which uniquely identifies a user"}
     )
+    access_token_ttl = marshmallow.fields.Integer(
+        required=False,
+        metadata={
+            "description": "Expiration of the access token, in milliseconds since January 1, 1970 "
+            "UTC. 0 means the expiration time is unknown."
+        },
+    )
 
 
 class WOPICheckFileInfoSchema(marshmallow.Schema):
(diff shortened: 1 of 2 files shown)


alubert-algoo (approved):

lebouquetin:
I do not see a test case that will ensure we always send both parameters, right ?

The fix can be merged, but the test cases do not ensure we will identify if at a time we do not send both parameters anymore

lebouquetin:
In this kind of situation, my expectation is:

1. Fix the bug
2. Release the new release
3. Add/fix test cases if it is required

This allow to deliver fast the new release while improving the code for the future with the right test cases.

Of course, if the update of test cases is easy, you can do it at the same time as you fix the code

@piRGoif @alubert-algoo

piRGoif:
> I do not see a test case that will ensure we always send both parameters, right ?

I modified backend/tracim_backend/tests/functional/test_wopi.py to add the extra field : every tests in that class are failing without the fix

Do you have something else in mind @lebouquetin ?

lebouquetin:
> Do you have something else in mind @lebouquetin ?

Imagine our integration miss to send the 2 parameters. These tests will success (because they manage the right case) but the behaviour will fail because the frontend do not send the two parameters. Do you confirm this or I misunderstand the case ?

lebouquetin:
Maybe I misunderstand the subject. Let's merge and we'll take 5 minutes on next monday @piRGoif

piRGoif:
The issue was caused by Collabora sending 2 params to the endpoint (access_token, access_token_ttl) while the validation schema only included one (access_token)

Ok go for the merge then, and discuss about it later

PR #6908, opened 2026-08-03
Auto test Debian 13

### LINKED ISSUE #6878: Debian 13 support in auto tests
We updated Tracim to support Debian 13 in #6859, we now need to support this OS also in auto tests

piRGoif:
Fixes #6878


piRGoif:
Currently branch has linter issue fixes but they will be pushed with #6921 PR, so the branch will be rebased interactively

concourse/scripts/backend-full-tests
```diff
@@ -1,11 +1,11 @@
 #!/bin/bash
 set -e
 
-python_version=${1:-3.13.14}
+script_dir="$(realpath $(dirname $0))"
+python_version=${1:-$(cat "$script_dir/../PYTHON_VERSION")}
```
alubert-algoo: I do not thinks this is the best way to deal with this version.

It is better to use environment variable in this case. This script is loaded by the Docker container, so we can set this variable in the Dockerfile.

It will also fix the hard-coded version from `tools_docker/concourse/Dockerfile`.

piRGoif:
> Currently branch has linter issue fixes but they will be pushed with #6921 PR, so the branch will be rebased interactively

PR was merged to maintenance then develop branches. Our PR branch is now rebased on top of those changes, and no longer contains linter issue commits

concourse/scripts/backend-full-tests
alubert-algoo: I have checked the commit where you switch to the environment variable and it seems fine to me. Did you tried it locally?

piRGoif: yes, by running image with the -e parameter to provide env var
It is running smoothly on the CI too (run >= 12 on Concourse)

piRGoif: ok closing conversation and preparing merge !

PR #6918, opened 2026-08-04
fix(frontend): prevent > character from appearing before mentions (Fixes #6914)

### LINKED ISSUE #6914: 🐛 Bug: The greater-than character appears above a mention if a note started with this same mention
### Tracim version

<=2026.07.04

### Issue description

When we started the content of a note with a mention. Every time we use this mention in the rest of the document, it will be prefixed with the greater-than character.

If we save the document again, the character will be add again.

This problem do not occurs on another mention, if this is the first element of a line.

### How to reproduce

1. In a space, click on “Write a note” in the right sidebar
2. Add a title to the note and click on “Validate and create”
3. The editor will be open, start this document by mentioning someone (with the @ character)
4. Add a new line, and write a text with the mention (for example: “`Foo @bar`”)
5. Click on validate

### Screenshots

If we write this content and validate:
Image

We will obtain this document:
Image

_No response_

_No response_

### Related PR

_No response_

waterWang:
## Problem

When a note starts with a mention (e.g. `@bar`), every subsequent use of that same mention in the document gets prefixed with the `>` character.

## Root Cause

The `searchMention` function uses the regex `/(?:^|\s|\W)@([a-zA-Z0-9_.-]+)\b/g` which matches `>` (non-word character) as a valid boundary before `@`. This causes `>@bar` to be matched as a mention, and the `>` gets included in the rendered output.

## Fix

Changed the regex to `/\B@([a-zA-Z0-9_.-]+)\b/g` using `\B` (non-word boundary) instead of `(?:^|\s|\W)`. This ensures mentions are only matched at word boundaries, not after non-word characters like `>`.

Also updated the corresponding dynamic regex in `searchMentionAndReplaceWithTag`.

Fixes #6914

### DIFF
--- a/frontend_lib/src/mentionOrLinkOrSanitize.js
+++ b/frontend_lib/src/mentionOrLinkOrSanitize.js
@@ -99,7 +99,7 @@ export const searchMention = (text) => {
   // Regex explanation: https://regex101.com/r/hHosBa/11
   // Match (@XXX part): '@XXX', ' @XXX ', '@XXX-', ':@XXX:', '(@XXX)', '!@XXX!', ...
   // Don't match: 'XXX@XXX', '@<span>XXX</span>'
-  const mentionRegex = /(?:^|\s|\W)@([a-zA-Z0-9_.-]+)\b/g
+  const mentionRegex = /\B@([a-zA-Z0-9_.-]+)\b/g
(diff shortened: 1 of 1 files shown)


lebouquetin:
I close this as this is SPAM - see https://github.com/tracim/tracim/pull/6919

PR #6919, opened 2026-08-04
fix(frontend_lib): prevent > character from being captured by mention regex

### LINKED ISSUE #6914: 🐛 Bug: The greater-than character appears above a mention if a note started with this same mention

waterWang:
### Issue
Fixes #6914

### Root Cause
The `searchMention` regex `/(?:^|\s|\W)@([a-zA-Z0-9_.-]+)\b/g` uses `\W` which matches `>` (greater-than) as a valid preceding character. When a note starts with a mention like `@user`, and the same mention appears later in the content, the `>` character from blockquote markup gets captured as the leading character (`mention[0]`), causing `>` to appear before the mention in the rendered output.

### Fix
Replace `\W` with `[^\w>]` in the regex to exclude `>` from the character class while keeping all other non-word characters.

Changes:
- `frontend_lib/src/mentionOrLinkOrSanitize.js`: Updated both `searchMention()` and `searchMentionAndReplaceWithTag()` regex patterns

### DIFF
--- a/frontend_lib/src/mentionOrLinkOrSanitize.js
+++ b/frontend_lib/src/mentionOrLinkOrSanitize.js
@@ -99,7 +99,7 @@ export const searchMention = (text) => {
   // Regex explanation: https://regex101.com/r/hHosBa/11
   // Match (@XXX part): '@XXX', ' @XXX ', '@XXX-', ':@XXX:', '(@XXX)', '!@XXX!', ...
   // Don't match: 'XXX@XXX', '@<span>XXX</span>'
-  const mentionRegex = /(?:^|\s|\W)@([a-zA-Z0-9_.-]+)\b/g
+  const mentionRegex = /(?:^|[^\w>])@([a-zA-Z0-9_.-]+)\b/g
   const mentionList = text.match(mentionRegex)
   return mentionList || []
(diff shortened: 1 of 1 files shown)


lebouquetin:
@waterWang maybe you fix the issue but you create another one. By the way, you did not update the tests, you also sent a series of PR - some being inappropriate.

I reported this PR with the SPAM - I'm pretty sure you are vibe-coding on a series of projects without to understand what it is about - and even without to test your own fix.

Please stop.

PR #6922, opened 2026-08-06
Fix linter behavior difference between pre-commit and CI

### LINKED ISSUE #6921: 🐛 Bug: pre-commit isn't aligned anymore with requirements-devtool.txt
### Tracim version

2026.07.00

### Issue description

While working on #6878, Concourse failed due to linter issues on the following files:
backend/tracim_backend/lib/crud_hook/hookspec.py
backend/tracim_backend/lib/utils/cors.py
backend/tracim_backend/lib/utils/translation.py
backend/tracim_backend/lib/webdav/__init__.py
backend/tracim_backend/tests/models/test_content.py

After fixing them locally the pre-commit hook failed, reverting the modifications

=> CI and pre-commit don't behave the same way.
We should have the exact same linter behavior between both pre-commit and Concourse.

### How to reproduce
Apply the linter fix and try to commit

### Expected

Update pre-commit versions (local git hook) to the one used in requirements-devtools.txt (Concourse), as the later were updated in #6759 to be compatible with Python 3.13.

### Investigation
Regression brought by #6859: we updated requirements-devtool.txt (used on Concourse) but forgot pre-commit config file (used locally as a git hook)
- Flake8 was updated from 6.0.0 to 6.1.0
- Black from 23.3.0 to 24.10.0

Also note that flake8 isn't launched in pre-commit: `entry: bash -c "cd backend"` just return passed.

During bug resolution, after updating the versions the Concourse build failed again on pycodestyle E704 errors. 
This was due to a known Flake8/Black conflict, both adopting different behaviors. Flake8 handled that by adding E704 in its default ignore list... but we were overriding it in the config ! 
So I changed the option to extend ignore list instead of overriding it. W503 was part of our list, but is already in the default ignore list both in Flake8 [previous (6.0.0)](https://github.com/pycqa/flake8/blob/6.0.0/src/flake8/defaults.py) and [current (6.1.0)](https://github.com/pycqa/flake8/blob/6.1.0/src/flake8/defaults.py) versions so I removed it.

About E704 different behavior see:
- Black 24.1 change was made with ref 3796: https://github.com/psf/black/blob/main/CHANGES.md#highlights-10
- Flake8 about working with other tools on E704: https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html#e701-e704

The isort dependency was left apart as it is blocked by pre-commit version and doesn't create issue for now. It will be done in #6923.

piRGoif:
Fix #6921


alubert-algoo (approved):

lebouquetin (changes requested):
I do not understand the PR.

The issue is: precommit and CI are not aligned.

What I expect to read: precommit aligned to CI OR CI aligned to precommit

This is the reason why I asked to add a part in the description, somthing like this, straight, clear, factual:

```
## Expected

- upgrade black xxx
- disable flake8 xxx
- ...
```

Could you put this section in the issue?  Otherwise, like told in my other comment: *the proper resolution of a bug is open to interpretation.*

.pre-commit-config.yaml
```diff
@@ -13,8 +13,8 @@ repos:
           - id: debug-statements
           - id: check-merge-conflict
     # backend specific
-    - repo: https://github.com/ambv/black
-      rev: 23.3.0
+    - repo: https://github.com/psf/black
+      rev: 24.10.0
```
lebouquetin, line 18: As discussed with @alubert-algoo at the moment - and in relation with our internal thread https://algoo.tracim.fr/ui/workspaces/162/contents/thread/192944, in order to ensure consistency between files, a way to make things work well is to put a link inline.

Instead of

```
      rev: 24.10.0
```

We will put something like:

```
      rev: 24.10.0  # INFO - P.G. - See #xxx
```

Where `#xxx` is an issue which explain which files must be updated at the same time.

Another way would be (we probably have something like this in tracim source code)

```
      rev: 24.10.0  # INFO - D.A. - search for all instances of BLK-VER-CONSISTENCY in the project sources
```

Which do not use an external issue but allow direct access to all instances of the same version of the dependency.

cc @piRGoif

piRGoif: Pushed fa93f0c0 to add a comment in pre-commit config

piRGoif: Discussed with Aurelien and Damien: hard to match between pre-commit and requirements* as isort has different names. We whose to add a link to internal doc.

PR #6929, opened 2026-08-13
add real-time updating in the Kanban (#6872)

### LINKED ISSUE #6872: ✨ Feat: allow to update kanban partially
## Currently

Kanban is stored as a full content, ie if you edit a card, then the enire kanban is re-stored

## Expected

Using columns and card uuids, we can imagine a strategy to patch content and allow several edit at the same time by only saving the edit-located part of the content

## Work

- Identify the possibility in order to implement a patch strategy with pros/cons
- defined a plan to implement it and strategy to manage edge cases if required

First ideas: 
- a PATCH strategy which store only edited card - but what if the card is moved from a column to another one? ( refactor the content in order to store in columns only reference to cards and store cards as independent entities ?)
- a DIFF strategy identifying what has changed (cards, columns) and apply the diff if no conflict? What about the conflicts?
- implement cards as sub-contents - but what if card is moved from a coloumn to another one?

### How it works

The idea is first to identify the possible strategies

### Related PR

_No response_

alubert-algoo:
Resolves #6872 

## Checkpoints

**For developers**

**For code reviewers**

**For testers**


frontend_app_kanban/i18next.scanner/fr/translation.json
```diff
@@ -70,5 +70,8 @@
   "No card": "Aucune carte",
   "Reopen the card": "Réouvrir la carte",
   "Show Gantt view": "Afficher le Gantt",
-  "Today": "Aujourd’hui"
+  "Today": "Aujourd’hui",
+  "This content has been modified.": "Ce contenu a été modifié.",
+  "Reload it!": "Recharger le contenu !",
+  "If you still want to use your current work, simply check the cross and the validation will be restored.": "Si vous souhaitez tout de même publier vos éditions, cliquez sur l’icône en forme de croix et la validation sera de nouveau possible."
```
piRGoif: When dev will be over, it will be worth opening Weblate to check those new translation keys status both in EN and FR ?

backend/setup.py
```diff
@@ -17,6 +17,8 @@
     "sqlakeyset",
     # API
     "hapic[marshmallow]>=1.0",
+    # API file patching
+    "jsonpatch==1.33",
```
piRGoif: Maybe don't pin the version here so we can get security alerts and fix propositions ?
```suggestion
    "jsonpatch>=1.33",
```

backend/tracim_backend/applications/content_file/controller.py
```diff
@@ -246,6 +255,158 @@ def download_file(self, context, request: TracimRequest, hapic_data=None):
...
+
+        if (revision_id := hapic_data.query.current_revision_id) != content.revision_id:
+            raise ConflictingNewerRevWhilePatching(
+                f"the revision sent with the patch ({revision_id}) do not match "
+                f"the current revision of the content ({content.revision_id}). "
+                "This patch will be ignored."
+            )
+
+        match content.file_mimetype:
+            case "application/json":
```
piRGoif: I'm a big fan of exit conditions placed at the function top O:)

So I'm suggesting to replace this match/case by an if placed just below the content var init, just before revision_id test

backend/tracim_backend/applications/content_file/controller.py
```diff
@@ -246,6 +255,158 @@ def download_file(self, context, request: TracimRequest, hapic_data=None):
...
+                patch_file = hapic_data.files.files
+                if patch_file.type != "application/json":
+                    raise UnvalidJsonFile(
```
piRGoif: I was surprised by this exception name (Unvalid instead of Invalid), but this is legacy code (submitted back in 2022)

The exception isn't listed as hapic.handle_exception function decorator though, is it normal ?

backend/tracim_backend/exceptions.py
```diff
@@ -74,6 +74,10 @@ class TagAlreadyExistsError(TracimError):
     error_code = ErrorCode.TAG_ALREADY_EXISTS
 
 
+class ConflictingNewerRevWhilePatching(TracimError):
```
piRGoif: What do you think about this name instead: PatchRevisionOlderThanContentRevision ?
Same suggestion for error code O:)

backend/tracim_backend/applications/content_file/controller.py
```diff
@@ -246,6 +255,158 @@ def download_file(self, context, request: TracimRequest, hapic_data=None):
...
+            except CannotGetDepotFileDepotCorrupted as exc:
+                raise TracimFileNotFound(
+                    "file related to revision {} of content {} not found in depot.".format(
+                        revision.revision_id, revision.content_id
+                    )
+                ) from exc
+
+        else:
+            raise ContentTypeNotAllowed(
+                f"the content_type {from_revision.file_mimetype} is not available with this route."
```
piRGoif: The exception isn't declared in the function decorator ? There is a dedicated http status : 415

Also a message change suggestion:

```suggestion
                f"Content type '{from_revision.file_mimetype}' isn't supported"
```

backend/tracim_backend/applications/content_file/controller.py
```diff
@@ -246,6 +255,158 @@ def download_file(self, context, request: TracimRequest, hapic_data=None):
...
+            session=request.dbsession,
+            config=app_config,
+        )
+
+        from_revision = api.get_one_revision(
+            hapic_data.query.from_revision_id,
+            content_id=hapic_data.path.content_id,
+        )
+
+        if from_revision.file_mimetype == "application/json":
```
piRGoif: Same suggestion as in upload_patch_file: invert condition so that exit condition stays at the very top

backend/tracim_backend/applications/content_file/controller.py
```diff
@@ -246,6 +255,158 @@ def download_file(self, context, request: TracimRequest, hapic_data=None):
...
+                        new_mimetype=content.file_mimetype,
+                    )
+                    api.save(content)
+
+                return {"new_revision": content.revision_id}
+
+            case _:
+                raise ContentTypeNotAllowed(
+                    f"the content_type {from_revision.file_mimetype} is not "
+                    "available with this route."
```
piRGoif: Same comment as in download_patch_file

backend/tracim_backend/applications/content_file/controller.py
```diff
@@ -246,6 +255,158 @@ def download_file(self, context, request: TracimRequest, hapic_data=None):
...
+            hapic_data.query.from_revision_id,
+            content_id=hapic_data.path.content_id,
+        )
+
+        if from_revision.file_mimetype == "application/json":
+            to_revision = api.get_one_revision(
+                hapic_data.query.to_revision_id,
+                content_id=hapic_data.path.content_id,
+            )
+
```
piRGoif, line 326: shouldn't we add a safeguard to disallow too big revision difference ? Like allowing 10 revisions but not 10000 ?
Maybe this should be implemented in the corresponding schema ?

backend/tracim_backend/applications/content_file/controller.py
```diff
@@ -246,6 +255,158 @@ def download_file(self, context, request: TracimRequest, hapic_data=None):
...
+                    open(to_file.file_object._file_path, "rb") as to_buffer,
+                ):
+                    patch = make_patch(load(from_buffer), load(to_buffer))
+                    return {
+                        "from_revision": hapic_data.query.from_revision_id,
+                        "to_revision": hapic_data.query.to_revision_id,
+                        "patch_content": patch,
+                    }
+
+            except CannotGetDepotFileDepotCorrupted as exc:
```
piRGoif: exception not declared in function decorator ?

frontend_app_kanban/src/component/KanbanCard.jsx
```diff
@@ -104,7 +104,10 @@ function KanbanCard (props) {
         {showConfirmPopup && (
           <ConfirmPopup
             onCancel={() => setShowConfirmPopup(false)}
-            onConfirm={() => props.onRemoveCard(props.card)}
+            onConfirm={() => {
+              setShowConfirmPopup(false)
+              props.onRemoveCard(props.card)
+            }}
```
piRGoif, line 114: Why was it necessary ? Maybe add a comment to explain why this was done ?

piRGoif (changes requested):

frontend_app_kanban/i18next.scanner/fr/translation.json
lebouquetin: development should deliver EN+FR then a weblate overview is important to make things consistant and update missing stuff

backend/tracim_backend/applications/content_file/controller.py
lebouquetin: exit conditions at top is avery good practice when exit conditions are easy to read. This make maintainance easier as you do not needto read the entire stuff for simple cases.

This can be worded like this:

- simple case, fast read
- complexe case (or nominal case), full read

backend/tracim_backend/applications/content_file/controller.py
lebouquetin: You can rename it by the way if its 2 or 3 renamings

backend/tracim_backend/applications/content_file/controller.py
lebouquetin: There is no reason to limit things. Something slow is better than something not working at all.

A good practice is to put limitations only when facing issues.

backend/tracim_backend/applications/content_file/controller.py
lebouquetin: @alubert-algoo it is not expected to have this in the controller, you must move this in ContentApi in order we can reuse the feature in CLI or anywhere else.

lebouquetin (changes requested):
@alubert-algoo important refactor is expected (not complex but important):

- All the intelligence must be in some ContentApi methods, not in the controller. Otherwise, we cannot reuse it.
- Controller implementation must match with something like Adaptor design pattern

backend/tracim_backend/applications/content_file/controller.py
alubert-algoo: In this case, this is the `TracimFileNotFound` which must be declared, since this is the one which will be raised. And this is the case in the decorator.

backend/tracim_backend/applications/content_file/controller.py
piRGoif: I made the suggestion because the task is quite CPU consuming, and any reader access user can call this API. I guess this would ease to overwhelm a Tracim instance ?

lebouquetin: We would live patch if needed. We have no idea of the real limitation that would be useful. So better not to spend time now

frontend_app_kanban/src/component/KanbanCard.jsx
alubert-algoo: The comment was in the dedicated commit, but I have forget this PR will be squash, I will add this comment in the file.

backend/tracim_backend/applications/content_file/controller.py
piRGoif: Decision made, closing conversation

backend/tracim_backend/applications/content_file/controller.py
alubert-algoo: This exception was removed from this method with the latest patches.

backend/tracim_backend/lib/core/content.py
```diff
@@ -2599,3 +2621,123 @@ def _get_content_property(
...
+            content_id (int): The identifier of the content.
+            content_type (str): The type of the content.
+            patch_file_content (str): The content of the patch to apply.
+            patch_mimetype (str): The mimetype of the patch.
+
+        Returns:
+            int: The latest revision of the content after the application of the patch.
+
+        Raises:
+            ContentTypeNotAllowed: If the specified content and patch are not JSON files.
```
piRGoif: There are multiple exceptions raised in this function, not just one :/

backend/tracim_backend/lib/core/content.py
```diff
@@ -2599,3 +2621,123 @@ def _get_content_property(
...
+
+        content = self.get_one(content_id, content_type)
+        if revision_id != content.revision_id:
+            raise ConflictingNewerRevWhilePatching(
+                f"the revision sent with the patch ({revision_id}) do not match "
+                f"the current revision of the content ({content.revision_id}). "
+                "This patch will be ignored."
+            )
+
+        if content.file_mimetype != "application/json":
```
piRGoif: I would move this test above the one on revision_id, so that it is closer to the patch_mimetype test

backend/tracim_backend/tests/library/test_content_api.py
```diff
@@ -3580,6 +3582,298 @@ def test_unit__get_all_query__with_user(
...
+                content_type_slug=content_type_list.File.slug,
+                workspace=workspace,
+                parent=None,
+                label="json_file",
+                do_save=False,
+            )
+            api.update_file_data(text_file, "json_file", "application/json", b"[]")
+
+        api.save(text_file, ActionDescription.CREATION)
+        assert text_file.revision_id == 1
```
piRGoif, line 3612: Can this be factorized to a pytest setup method equivalent ?

backend/tracim_backend/tests/library/test_content_api.py
```diff
@@ -3580,6 +3582,298 @@ def test_unit__get_all_query__with_user(
...
+
+        with new_revision(session, transaction.manager, content=text_file):
+            api.update_file_data(
+                text_file, text_file.label, text_file.file_mimetype, b'[{"foo": "bar"}]'
```
piRGoif, line 3783: Can the values be extracted to a variable, so that it would appear it is really the same between submission and retrieval ?

piRGoif (changes requested):
Bravo, great work ! 
Just a few suggestions, but they are not blockers

backend/tracim_backend/tests/library/test_content_api.py
alubert-algoo: I just tried and since the final method will have a lot of arguments (the same as the testing method), this refactoring will not improve the length of the test. The only gain is to have a method to call all the procedure at once, but since the other tests from this file do not do it, I think I will pass on this one.

backend/tracim_backend/tests/library/test_content_api.py
alubert-algoo: In this method, this value only appear once. The verification will be done on a Python object (and not a bytes), so if we use a variable, we need to decode this one and use the `json.load` method to convert the decoded string to an array.

backend/tracim_backend/lib/core/content.py
alubert-algoo: I have changed to message and add the missing exception.

backend/tracim_backend/tests/library/test_content_api.py
piRGoif: I see the same value L3783 (update file) and L3793 (assert) ?

backend/tracim_backend/exceptions.py
piRGoif: was done, closing conversation

backend/tracim_backend/tests/library/test_content_api.py
piRGoif: I was just thinking of extracting the technical non test related stuff out of the test function.
But ok I understand, closing conversation

backend/tracim_backend/applications/content_file/controller.py
piRGoif: Message modification implemented.

Decorator for ContentTypeNotAllowed exception is present on download_patch_file but not on upload_patch_file

piRGoif: Fixed ! Closing

piRGoif (approved):

PR #6935, opened 2026-08-14
Github issues templates : remove "related pr" section + set issue type

piRGoif:
Github can link the PR using the "development" section, and this can be done when creating the PR by adding a reference in its description ([Linking a pull request to an issue - GitHub Docs](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue))
As it is something we are doing every time now, I think we can get rid of the "related pr" part in the issue description ?

The "type" field ([Managing issue types in an organization - GitHub Docs](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/managing-issue-types-in-an-organization)) is also present, and not filed yet. This PR set the field value for Bug and Feature.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -2,6 +2,7 @@ name: Bug report
 description: Report a bug you detected using tracim
 title: "🐛 Bug: "
 labels: ["to sort", "bug"]
+type: "Bug"
 body:
   - type: markdown
     attributes:
@@ -72,13 +73,3 @@ body:
         - Smartphone
         - Tablet
         - Other
-  - type: textarea
-    id: related-pr
-    attributes:
-      label: Related PR
-      description: |
-        Let us know if this is related to an open pull request.
-        Leave empty if there is no related PR or if you don't know.
-      placeholder: PR #0000
-    validations:
-      required: false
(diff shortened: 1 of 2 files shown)


piRGoif:
@alubert-algoo this is a small modification in the issues template, do you agree with me ?

alubert-algoo:
> @alubert-algoo this is a small modification in the issues template, do you agree with me ?

There is a reason to remove the `type: textarea` block in the yaml files? I do not understand why you replace it with the `type` field.

lebouquetin:
Why do you remove the PR part? This part allows to link straight to the associated PR. How do you go to the associated PR if not used? (issues are currently linked altogether it is sometime difficult to find the one in charge of fixing the issue)

piRGoif:
Sorry I though you were sharing my concerns :/ I've added explanations in the description @alubert-algoo @lebouquetin

lebouquetin:
> Sorry I though you were sharing my concerns

As I explained a few weeks ago:

- when you suggest a change, I expect at least that you explain the why. You ask for a review, so we need to know *why* you suggest a change (this does not preclude the relevance - its just required to understand *why/what* you want to fix or improve and then allow us to review the *how*)
- Issues are not used by developers only, so I expect you ask me too for review of changes in issue templating. Maybe I won't be available, but the default is to ask me too.

---

With the updated description, I'm ok with your suggestion. I let you discuss with @alubert-algoo about the body part.

@piRGoif

alubert-algoo:
Seems good to me. These textareas were redundant with the quick keywords.

piRGoif:
Thanks to both of you for the review, sorry for the description, I'll do my best to do better next time !

PR #6936, opened 2026-08-14
Fix/6931 frontend unit tests errors

### LINKED ISSUE #6931: 🐛 Bug: errors in frontend-unit-tests
### Tracim version

2026.07.00

### Issue description

Since Concourse run were restored (#6878) we are seeing some frontend unit tests failing.

### How to reproduce

See Concourse runs

### Screenshots

_No response_

_No response_

_No response_

### Related PR

_No response_

piRGoif:
Fixes #6931


piRGoif:
All tests are passing now, removing draft flaf

frontend_app_agenda/test/babelRegister.js
alubert-algoo, line 1: I am surprised that the extensions value cannot be set in the `.mocharc.js`. There is a reason to do a double importation to set the extensions?

frontend_app_agenda/test/Agenda.spec.js
```diff
@@ -85,7 +85,7 @@ describe('<Agenda />', () => {
               }
             }
           }
-          wrapper.setState({ userWorkspaceList: userWorkspaceList })
+          wrapper.setState({ agendaList: userWorkspaceList })
```
lebouquetin, line 88: I do not understand: why `userWorkspaceList` replaced with `agendaList`?

frontend_app_agenda/test/Agenda.spec.js
```diff
@@ -1,7 +1,7 @@
 import React from 'react'
 import { shallow } from 'enzyme'
 import { expect } from 'chai'
-import { Agenda } from '../src/container/Agenda.jsx'
+import { Agenda } from '../src/container/Agenda.tsx'
```
lebouquetin: This seems strange to my.

In my comprehension of open-calendar (and in the instructions I gave on open-calendar development), the open-calendar source-code is written in typescript but the integration in tracim should be in JS (this is a standard dependency - or at least it should).

Doesn't it the case? (or maybe I misunderstand the meaning of `.tsx` extension?)

lebouquetin (commented):
A few questions to help us understand and evaluate the fix

frontend_app_agenda/test/babelRegister.js
piRGoif: Perfectly right, this new file wasn't necessary ! I've removed it with c1a07687

frontend_app_agenda/test/Agenda.spec.js
piRGoif: The js to ts was done in ac2e698

- Agenda.tsx is called from frontend_app_agenda/src/index.tsx but without any extension
- index.tsx itself is referenced in webpack.standalone.config.js and webpack.servdev.config.js
- both webpack files are used as source for transpilation 

=> client side, both on production and serdev we are not seeing any ts code as it is transpiled.

About the import in the test, I see I can also use no extension, I've pushed that modification with 523af6c5

lebouquetin: I think you fix the symptom. Lets speak live

frontend_app_agenda/test/Agenda.spec.js
piRGoif: The regression was introduced in 1e3f6683, in this commit Agenda props were modified:
- removed userWorkspaceList 
- added agendaList and addressBookList

The tested functions (handleSharedspaceModified / handleUserModified) are never reading addressBookList so I just replaced userWorkspaceList by agendaList

lebouquetin (approved):

alubert-algoo (approved):

PR #6938, opened 2026-08-17
allow to enable gantt at runtime

### LINKED ISSUE #6937: 🐛 Bug: the gantt feature can only be activate during build time
### Tracim version

build_192

### Issue description

The Gantt feature is currently disable and can only be activated with an environment variable. The current implementation only use this variable during the built of the container image, which means we have to build a dedicated version with this feature.

To use the variable during runtime, we need to put this one in the system configuration sent through the API from the route `/api/system/config`.

By adding this variable into the response, we will be able to retrieve the variable at runtime instead of build time.

For the frontend, we need to:
- Revert the changes on `frontend/webpack.standalone.config.js` and `frontend/src/util/helper.js` (from the commit 57cd46b3067ea4e5e83d93b71054215de5c25ad9)
- Use the value from the config instead of the static built value in `frontend/src/util/appFactory.js`

### How to reproduce

When we build the image and the environment variable was not set, the container using this image will not have access to the Gantt feature.

### Screenshots

_No response_

_No response_

_No response_

alubert-algoo:
Resolves #6937.

### DIFF
--- a/backend/tracim_backend/config.py
+++ b/backend/tracim_backend/config.py
@@ -739,6 +739,10 @@ class CFG(object):
         except ValueError:
             raise ConfigurationError("The value of ui.notes.code_sample_languages is malformed")
 
+        # INFO - A.L - 2026-08-17 - This variable is temporary and will be removed
+        # in the future when the Gantt feature is fully implemented.
+        self.GANTT_VIEW__ENABLED = asbool(self.get_raw_config("gantt_view.enabled", "False"))
+
     def __load_uploaded_files_config(self) -> None:
         default_depot_storage_path = self.here_macro_replace("%(here)s/depot")
         self.DEPOT_STORAGE_DIR = self.get_raw_config(
--- a/backend/tracim_backend/lib/core/system.py
+++ b/backend/tracim_backend/lib/core/system.py
@@ -104,6 +104,7 @@ class SystemApi(object):
             limitation__max_non_guest_users=self._config.LIMITATION__MAX_NON_GUEST_USERS,
             limitation__max_guest_users=self._config.LIMITATION__MAX_GUEST_USERS,
             limitation__max_guest_user_space_nb=self._config.LIMITATION__MAX_GUEST_USER_SPACE_NB,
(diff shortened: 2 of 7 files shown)


backend/tracim_backend/config.py
```diff
@@ -739,6 +739,10 @@ def _load_global_config(self) -> None:
         except ValueError:
             raise ConfigurationError("The value of ui.notes.code_sample_languages is malformed")
 
+        # INFO - A.L - 2026-08-17 - This variable is temporary and will be removed
+        # in the future when the Gantt feature is fully implemented.
+        self.GANTT_VIEW__ENABLED = asbool(self.get_raw_config("gantt-view.enabled", "False"))
```
lebouquetin: Config is almost always based on underscore instead of hyphen.

So `gantt-view.enabled` would be better set as `gantt_view.enabled`

lebouquetin (changes requested):

backend/tracim_backend/config.py
alubert-algoo: I have updated the commit to use the underscore.

lebouquetin (approved):

backend/tracim_backend/config.py
```diff
@@ -739,6 +739,10 @@ def _load_global_config(self) -> None:
         except ValueError:
             raise ConfigurationError("The value of ui.notes.code_sample_languages is malformed")
 
+        # INFO - A.L - 2026-08-17 - This variable is temporary and will be removed
+        # in the future when the Gantt feature is fully implemented.
```
piRGoif, line 743: The phrase is quite strange to me, maybe change it to something like:

```suggestion
        # INFO - A.L - 2026-08-17 - Enable the Gantt view in Kanban app
        # Temporary, will be removed when feature dev will is complete
```

piRGoif (approved):
Just a little comment rephrase suggestion but I approve nonetheless

backend/tracim_backend/config.py
alubert-algoo: We have seen together to ignore this suggestion and to merge this pr.

PR #6939, opened 2026-08-18
Fix Cypress before hook

### LINKED ISSUE #6930: 🐛 Bug: end to end auto tests errors
### Tracim version

2026.07.00

### Issue description

Since Debian 13 migration (#6859), auto tests are broken.

Expected: 
Able to launch the tests locally and on CI to check version 

### How to reproduce

Launch Cypress tests 
=> almost every tests are failing in before all hook

### Issues fixed

> [!NOTE]
> Apart the Chromium version issue (develop only), all below issues fixes were merged to maintenance/2026-07 then to develop

#### Tests trigger list

The tests are launched only if at least one of the modified files matches a pattern configured in concourse/scripts/end-to-end-cypress-tests. This config was lacking functionnal_tests dir.

#### Before hook

The fixtures were sent as is to the API, which was rejecting with 400 error because of marshmallow validation having unknown fields.
Solution implemented (PR #6939) was to remove extra fields from data sent to the Tracim API.

#### Content lists click

By default Cypress will click in the center of the target element. Some tests were failing because in contents list, the click happens in the content author column, so opening the author homepage instead of the content details. 

A similar issue was already fixed years ago in 413e3f37 by adding `.content__item` selector filtering, so I applied the same.

#### Search

Actually my local config was set to advanced search, and the tests are only compatible with simple search. I added a comment explaining that.

##### iFrame

Since switch to OpenDavCal we migrate from an iframe to a div (commit ac2e6989) but the test wasn't updated accordingly.

##### OpenDavCal Document.parseHTMLUnsafe VS Chromium 118

Before PR Cypress was running embedded Electron/Chromium version 118. It is incompatible with OpenDavCal see algoo/open-calendar#31 (we must use Chrome >= 124).

We are already running latest 13.* Cypress version (13.17.0), and this branch only has 118 browser version.   
Version 14.* update Electron and have a 130 version both for 14.0.0 and latest 14.5.4. But... updating to Cypress 14 means updating Node as it only supports 18+ - see https://docs.cypress.io/app/references/migration-guide#Nodejs-18-support. About updating node see #6942.

So instead I've added Chromium to the Concourse image (version 151 as the one bundled in Debian 13) and added a policy for clipboard access (solve "NotAllowedError: Failed to execute 'writeText' on 'Clipboard'" error).

piRGoif:
This solves most of the Cypress errors, but there are still remaining, that will be fixed in another PR

The base branch is maintenance as Cypress was totally unusable on that branch

### DIFF
--- a/functionnal_tests/cypress/support/db_commands.js
+++ b/functionnal_tests/cypress/support/db_commands.js
@@ -85,7 +85,7 @@ Cypress.Commands.add('createRandomUser', (profile = 'users') => {
 Cypress.Commands.add('createUser', (fixturePath = 'baseUser') => {
   return cy
     .fixture(fixturePath)
-    .then(userJSON => cy.request('POST', '/api/users', userJSON))
+    .then(({ user_id: _userId, ...userJSON }) => cy.request('POST', '/api/users', userJSON))
     .then(handleUndefinedResponse)
 })
 
@@ -107,7 +107,7 @@ Cypress.Commands.add('createWorkspace', (fixturePath = 'baseWorkspace') => {
   let url = '/api/workspaces'
   return cy
     .fixture(fixturePath)
-    .then(workspaceJSON => cy.request('POST', url, workspaceJSON))
+    .then(({ workspace_id: _workspaceId, ...workspaceJSON }) => cy.request('POST', url, workspaceJSON))
     .then(handleUndefinedResponse)
 })
 
--- a/functionnal_tests/cypress/support/db_commands.js
+++ b/functionnal_tests/cypress/support/db_commands.js
@@ -85,7 +85,7 @@ Cypress.Commands.add('createRandomUser', (profile = 'users') => {
 Cypress.Commands.add('createUser', (fixturePath = 'baseUser') => {
   return cy
     .fixture(fixturePath)
-    .then(({ user_id: _userId, ...userJSON }) => cy.request('POST', '/api/users', userJSON))
+    .then(({ user_id, ...userJSON }) => cy.request('POST', '/api/users', userJSON))
     .then(handleUndefinedResponse)
 })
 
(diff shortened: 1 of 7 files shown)


functionnal_tests/cypress/support/db_commands.js
```diff
@@ -85,7 +85,7 @@ Cypress.Commands.add('createRandomUser', (profile = 'users') => {
 Cypress.Commands.add('createUser', (fixturePath = 'baseUser') => {
   return cy
     .fixture(fixturePath)
-    .then(userJSON => cy.request('POST', '/api/users', userJSON))
+    .then(({ user_id: _userId, ...userJSON }) => cy.request('POST', '/api/users', userJSON))
```
alubert-algoo: Where the `_userId` come from? If this is just to have an `undefined` value, I think you can directly use `undefined`.

functionnal_tests/cypress/support/db_commands.js
```diff
@@ -107,7 +107,7 @@ Cypress.Commands.add('createWorkspace', (fixturePath = 'baseWorkspace') => {
   let url = '/api/workspaces'
   return cy
     .fixture(fixturePath)
-    .then(workspaceJSON => cy.request('POST', url, workspaceJSON))
+    .then(({ workspace_id: _workspaceId, ...workspaceJSON }) => cy.request('POST', url, workspaceJSON))
```
alubert-algoo: Same question here for `_workspaceId`.

functionnal_tests/cypress/support/db_commands.js
piRGoif: We are excluding user_id and workspace_id from the received results, so that we can pass an array without those key to the cy.Request call.

The _userId var will be initialized from the results values.

alubert-algoo: Ok, I see. So this part can be write `{ user_id, ...userJSON }`, since this value will be not used in the final request.

piRGoif: Ok thanks, this is better indeed ! Pushed bdd52ad2 as a fix.

alubert-algoo (approved):

PR #6941, opened 2026-08-18
Fix Cypress error with OpenDavCal

### LINKED ISSUE #6930: 🐛 Bug: end to end auto tests errors

piRGoif:
Refs #6930

Only goal of this PR is to run the remaining fixes on the CI O:)

- test trigger dirs
- run_dev_backend:: allow to pass spec parameter
- documentation
- Chromium 118 VS OpenDavCal Document.parseHTMLUnsafe

### DIFF
--- a/docs.legacy/development/test/concourse.md
+++ b/docs.legacy/development/test/concourse.md
@@ -27,7 +27,7 @@ fly --target algoo login --team-name algoo --concourse-url https://ci.algoo.fr:4
 
 List all available images to look for their ids
 ```
-fly builds
+fly -t algoo builds
 ```
 
 Connect to a specific docker image by its id
@@ -39,7 +39,7 @@ fly -t algoo intercept -b build_id
 
 List the first <number_image> available image from algoo CI
 ```
-./fly builds -t algoo -c <number_image>
+./fly -t algoo builds -c <number_image>
 ```
 
 Create a file where you want to retrieve the image on local
@@ -54,16 +54,16 @@ Recuperate the image to the file created from it
 
 ### Where
 
-- <number_image>: number of image wanted, by default is 50
-- <team_name>: `algoo`
-- <build_id>: first column of `./fly builds -t algoo`
-- <step_name>: `end-to-end-cypress-tests`
-  - other available values are
-    - end-to-end-cypress-tests
-    - pull-request
-    - tracim-status-update
-- <imgae_location>: failed test in concourse will display the location after "(Screenshots)"
-- <local_file_location>: local file to put the screenshot in
+- `number_image`: number of image wanted, by default is 50
+- `team_name`: `algoo`
+- `build_id`: first column of `./fly builds -t algoo`
+- `step_name`: `end-to-end-cypress-tests`
+  - available values are:
+    - `end-to-end-cypress-tests`
+    - `pull-request`
+    - `tracim-status-update`
+- `image_location`: failed test in concourse will display the location after "(Screenshots)"
+- `local_file_location`: local file to put the screenshot in
 
 ### Example
 
--- a/docs.legacy/development/test/concourse.md
+++ b/docs.legacy/development/test/concourse.md
@@ -56,7 +56,7 @@ Recuperate the image to the file created from it
 
 - `number_image`: number of image wanted, by default is 50
 - `team_name`: `algoo`
-- `build_id`: first column of `./fly builds -t algoo`
+- `build_id`: first column of `./fly -t algoo builds`
 - `step_name`: `end-to-end-cypress-tests`
   - available values are:
     - `end-to-end-cypress-tests`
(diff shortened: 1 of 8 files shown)


piRGoif:
End to end tests are green \o/
Switching PR out of draft ready for review !

piRGoif:
PR is ready to review
Documentation is in issue #6930 "OpenDavCal Document.parseHTMLUnsafe VS Chromium 118"

docs.legacy/development/test/concourse.md
```diff
@@ -54,16 +54,16 @@ Recuperate the image to the file created from it
...
-- <step_name>: `end-to-end-cypress-tests`
-  - other available values are
-    - end-to-end-cypress-tests
-    - pull-request
-    - tracim-status-update
-- <imgae_location>: failed test in concourse will display the location after "(Screenshots)"
-- <local_file_location>: local file to put the screenshot in
+- `number_image`: number of image wanted, by default is 50
+- `team_name`: `algoo`
+- `build_id`: first column of `./fly builds -t algoo`
```
alubert-algoo: You have made this update above `./fly -t algoo builds -c <number_image>`. Do the change must be copied here?

tools_docker/concourse/Dockerfile
```diff
@@ -17,7 +17,13 @@ RUN \
...
 
 FROM base_install AS main
-    # tested python versions
+
+# HACK - PGO - 2026-08-19
+# Chromium version used to run Cypress, kept fixed so CI runs stay reproducible.
+#    see concourse/scripts/end-to-end-cypress-tests that forces to use Chromium intead of bundled Electron
+#    Our current Cypress 13 version has v118 that is incompatible with OpenDavCal (#6930)
+# Debian 13 is by default on Chromium 151.0.7922.137-1~deb13u1
+ARG CHROMIUM_VERSION=151.0.7922.137-1~deb13u1
```
alubert-algoo, line 26: This version will evolved quickly on Debian (even on stable). I hope it will not broke too soon :)

(since it is just for the building of the image, it will be OK)

alubert-algoo (approved):
This seems ok to me!

tools_docker/concourse/Dockerfile
piRGoif: Either we have a fixed version or we depend on a package provider right ? Having a fixed version is more stability, and we can update whenever we want to which version better suits us O:)

Do you agree ?

docs.legacy/development/test/concourse.md
piRGoif: Good catch ! Fixed in 384e8ba3

tools_docker/concourse/Dockerfile
alubert-algoo: Like I said, this is for the building of the image, the problem may occurs during the build, but it is not something we will be doing everyday, so it is OK.

We only have to remember this part can potentially break when building :)

piRGoif: Thanks ! 
Closing conversation and merging !

Issue #6945, opened 2026-08-20
🐛 Bug: Patch are applied on the Kanban board when visualize a previous revision


alubert-algoo:
### Tracim version

build_194

### Issue description

When the user click on a revision, in the timeline sidebar, the board will show to him the state of the board for this specific revision.

Since the latest changes on the Kanban to have realtime modification, the patch are loading when a TLM is received by a client. But when the board is in `revision` mode, these patches must be ignored.

The code did this verification, but it seems that the component found a way to bypass the `revision` node checking.

The board is in read-only when the `revision` mode is enable, so there is no risk to loose data.

### How to reproduce

1. In the Kanban board, I click on a revision in the timeline sidebar
2. The board will shown the state of the board
3. In another tab from the web browser, on the same Kanban board but with latest revision, I did a modification on a card/column
4. The first tab will shown the board with a modification

PR #6946, opened 2026-08-21
fix(kanban): do not reload board when a TLM occurs in revision mode.

alubert-algoo:
This patch will block the real-time reloading when an older revision is currently shown to the user.

Refs #6945

### DIFF
--- a/frontend_app_kanban/src/component/Kanban.jsx
+++ b/frontend_app_kanban/src/component/Kanban.jsx
@@ -109,7 +109,7 @@ export class Kanban extends React.Component {
     console.debug('%c<Kanban> component did update', 'color: gold', state, props, prevProps)
 
     if (props.mode === APP_FEATURE_MODE.REVISION) {
-      if (prevProps.content.current_revision_id !== props.content.current_revision_id) {
+      if (props.isSelectedRevision && prevProps.selectedRevisionId !== props.selectedRevisionId) {
         this.loadBoardContent()
       }
     } else {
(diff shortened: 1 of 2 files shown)


piRGoif (changes requested):
Nice comment, I understand your code
"previousRevision" seems to me not very understandable so I would rename it. It is the revision we had when the component was loaded right ?

alubert-algoo:
> Nice comment, I understand your code "previousRevision" seems to me not very understandable so I would rename it. It is the revision we had when the component was loaded right ?

The revision used for the Kanban will used `currentContentRevisionId` and `current_revision_id` (from `props.content`). I think `previousRevision` is logical in this case since the board will shown a previous state of the content.

piRGoif:
Ok, thanks.
Maybe then userSelectedRevisionId ?

alubert-algoo:
@piRGoif I have force-push to ensure there is no missing modification after the rename.

- `isPreviousRevision` was renamed as `isSelectedRevision`
- `previousContentRevisionId` was renamed as `selectedContentRevisionId`
- `previousRevisionId` was renamed as `selectedRevisionId`

piRGoif (approved):

piRGoif:
There is a frontend-tests issue but seems like a temporary network problem ?

piRGoif:
Might me worse trying to create a corresponding Cypress test ? O:)

alubert-algoo:
I have restarted the frontend-tests. We will see if this is a temporary problem ;)

piRGoif:
In end-to-end, workspace/add_member.cy.js failed again :( Can you have a look ?

alubert-algoo:
> In end-to-end, workspace/add_member.cy.js failed again :( Can you have a look ?

@piRGoif The test is ok (run it on my machine from multiple branches). Maybe it is the random number used to choose the role to assign which break in the CI?

```js
    it('should pre-select and save same role as workspace default role when adding new member in workspace', () => {
        const randomSelection = Math.floor(Math.random()*4)
```

piRGoif:
ok
debugging this test is another issue

PR #6948, opened 2026-08-26
fix(logbook): use the correct markup with the internal link.

### LINKED ISSUE #6934: 🐛 Bug: instance contents links aren't showed in logbook
### Tracim version

2026.07.00

### Issue description

In a logbook event description, inserting an internal link isn't interpreted when rendering the content

### How to reproduce

1. In a logbook event description, add an internal link by typing the `#` character followed by the content id, eg `#123`
2. => When typing the link the menu is shown, displaying corresponding content id+title
3. Validate the event
4. => the event is displayed with a raw `#...` instead of the content title with a link

### Screenshots

_No response_

_No response_

_No response_

### Related PR

_No response_

alubert-algoo:
The internal links, like `#1`, were not shown as link when the logbook was rendered.

By using the `searchContentAndReplaceWithTag` function, these links will be replaced with the correct markup and properly shown to the user.

Fixes #6934.

### DIFF
--- a/frontend_app_logbook/src/component/LogbookEntryEditor.jsx
+++ b/frontend_app_logbook/src/component/LogbookEntryEditor.jsx
@@ -7,6 +7,7 @@ import {
   IconButton,
   TextInput,
   TinyEditor,
+  searchContentAndReplaceWithTag,
   sendGlobalFlashMessage,
   getLocalStorageItem,
   setLocalStorageItem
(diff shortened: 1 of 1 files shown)


piRGoif (approved):
Seems ok to me but still not a React expert O:)

lebouquetin (approved):

PR #6952, opened 2026-08-27
fix(frontend): correctly sort the Timeline

### LINKED ISSUE #6894: 🐛 Bug: thread revisions and comments not in the right order
### Tracim version

v2026.07.00+build_187

## current status

Using a script, I add a comment to a thread then change the thread status.

Revision dates and ids in the database are:

for the comment:

```
revision_id       | 5580
created           | 2026-07-31 06:03:00.325142
updated           | 2026-07-31 06:03:00.325149
```

for the thread status update:

```
revision_id       | 5581
created           | 2026-07-17 14:37:07.535904
updated           | 2026-07-31 06:03:00.601796
```

The rendering:

Image

## Expected

According to the dates and revisions ids, the status change should be shown below the comment

### How to reproduce

Just go the the galae support space  `/ui/workspaces/28/contents/thread/3652`

### Screenshots

_No response_

_No response_

_No response_

### Related PR

_No response_

alubert-algoo:
Resolves #6894

### DIFF
--- a/frontend_lib/src/component/Timeline/Timeline.jsx
+++ b/frontend_lib/src/component/Timeline/Timeline.jsx
@@ -109,6 +109,19 @@ export class Timeline extends React.Component {
       return null
     }
 
+    props.timelineData.sort((first, second) => {
+      const firstCreated = new Date(first.created_raw)
+      const secondCreated = new Date(second.created_raw)
+      if (firstCreated < secondCreated) return -1
+      if (firstCreated > secondCreated) return 1
+
+      const firstRevisionId = first.revision_id || first.current_revision_id
+      const secondRevisionId = second.revision_id || second.current_revision_id
+      if (firstRevisionId < secondRevisionId) return -1
+      if (firstRevisionId > secondRevisionId) return 1
+      return 0
+    })
+
     const timelineDataGrouped = groupTimelineData(props.timelineData)
 
     return (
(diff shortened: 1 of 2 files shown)


frontend_lib/src/component/Timeline/Timeline.jsx
```diff
@@ -109,6 +109,19 @@ export class Timeline extends React.Component {
       return null
     }
 
+    props.timelineData.sort((first, second) => {
+      const firstCreated = new Date(first.created_raw)
+      const secondCreated = new Date(second.created_raw)
```
lebouquetin: if `created_raw` is a ISO 8601 datetime, then you may avoid conversion to date: ISO8601 datetime can be alphabeticaly sorted, ie:

```
first.created_raw > second.created_raw
```

gives the same result as

```
Date(first.created_raw) > Date(second.created_raw)
```

and it avoids conversion (this is probably faster?)

lebouquetin (approved):
Minor speed improvment? I'm not sure this is important - but in a sorting algorithm it may be …

This is not a requirement, so
 
- if you think its a good idean, then change
- if you think it does really not improve speed, skip the suggestion

frontend_lib/src/component/Timeline/Timeline.jsx
alubert-algoo: Good idea! The format used by the date is `YYYY-MM-DD`, so the sorting process will also works with this string.

The gain will not be very visible on small timeline, but it is always less CPU time wasted for the client.

alubert-algoo: I have implemented your proposition.

lebouquetin (approved):

lebouquetin:
@alubert-algoo this is ok for me :+1:

PR #6956, opened 2026-09-02
fix(kanban): cannot delete a card with an unknown dependency.

alubert-algoo:
If the card have a dependency to a card which been removed, the Board will crash by trying to retrieve the information about the deleted card.


lebouquetin:
Just to be sure: this part of the fix will also be merged into develop, right?

alubert-algoo:
> Just to be sure: this part of the fix will also be merged into develop, right?

Yes, the fix will be mandatory for the work in progress on the kanban/gantt improvment.

lebouquetin (approved):

piRGoif:
As it seems there are no corresponding issue I've added this PR to the 2026.10.00 project

PR #6962, opened 2026-09-14
fix: pin the version 1.41 for Pushpin in the container.

### LINKED ISSUE #6961: 🐛 Bug: The build of the container failed on pushpin installation
### Tracim version

all

### Issue description

The latest version of Pushpin (1.42) cannot be installed with Debian, since the package from the Pushpin repository want to install the library `libqt6network6t64` which is not available.

### How to reproduce

1. Try to build the container from this repository.
2. The build will failed on the pushpin installation.

### What was done
The Pushpin version was pinned to 1.41 directly in the Dockerfile.

alubert-algoo:
The new version 1.42 wants to install the package `libqt6network6t64` which does not exists on Debian 13.

This is a quick fix to force the installation of the version 1.41 instead of the latest one.

Resolves #6961.


lebouquetin (approved):
Please add an issue in order to go back to compile-from-sources installation process (with a pinned version of the sources).

Then this fix is ok for me

piRGoif:
There is also a docker in docker possibility

Also, we might contact the Pushpin maintainer by opening an issue ? (for reference we reported the Ubuntu Noble Pushpin package issue in https://github.com/fastly/pushpin/issues/48396#issuecomment-5241439290)

lebouquetin:
> There is also a docker in docker possibility

As I already explained: we need to be able to install Tracim on a raw debian system (VM or HW node), so we must avoid docker-specific deployment.

PR #6964, opened 2026-09-16
Refactoring of the Gantt functions

### LINKED ISSUE #6955: 🐛 Bug: gantt tasks sequence computation
# Expected work

## Refactor the workflow

- The Kanban format is a dictionnary: a set of columns containing cards.
- The Gantt format is also a dictionnary.

In order to sort and process card dependencies, we need to work on a task list instead of a dictionnary. All data computation must be done on this list of tasks. Then the list 

- can be transformed back to Kanban structure thanks to cards UUID
- can be transformed to Gantt format for rendering.

The list of tasks must be the pivot format.

Work to do: Implement two pre-processing steps that do the following:

1. [ ] transform the kanban JSON (a set of `Columns` including `Cards`) into a set of `Tasks`. The naming is important for understanding. this function will be named `convertKanbanToSetOfCards(kanbanStruct)`. Naming may be suggested but need to be validate together
2. [ ] a function will implement all the business rules, taking a list of Tasks as input and returning a (new) list of Tasks. This function will be name `applyRulesToTasks(taskList) -> taskList`.

All the rules will be implemented sequentially in the `applyRulesToTasks()` function in order to make things easy to understand, debug, evolve.

## The big picture

Here is the primary algorithm - and it must be written somewhere like this:

```
kanbanData = getKanbanData()
taskList = convertKanbanToSetOfCards(kanbanData)
processedTaskList = applyRulesToTasks(taskList, <parameters>)
ganttData = convertTaskListToGantt(processedTaskList)
return ganttData
```

Why?

- it allows to split by concern
- it makes easy to insert some rules or processing or debug or anything else
- it allows to reuse things without to rewrite

KISS

Tasks are setup with date only, no hour. Start and end date are « task included ». If a task duration is 1 day, then start=end. If a task B is after task A, then `taskB.start = taskA.end + 1`. **We do not need hours for this processing**.

**If hours are required for rendering**, then it will be computed at rendering time with the following strategy:

- start time is 0:01
- end time is 23:59

## Make things parametric

The `applyRulesToTasks()` will allow easy parameterization with the following parameters:

- `ignoreWeekendDays`: True/False, will allow to exclude weekend days from computation.
  - Example: a 3 days task that start on friday will end on tueday if ignoreWeekendDays is True, it will end on sunday if False)
  - If a task finishes on friday, then next tasks start on monday
- `weekendDays`: a list of weekend days index . Default is [0, 6] (sunday, saturday0 but it must be parametric (in middleeast, week-end is friday, saturday for example)

## Process all tasks / set default values

- [ ] tasks that have no start / end date and no dependency must start today. This will ensure users do not forget tasks
- [ ] if a task has no duration and no start/end date, consider it lasts 1 day

## Make the algorithm more robust

- [ ] circular dependencies must be identified. The claude-generated code I shared did it - it must be re-used. (see below)

## Compute full duration at the very end

- [ ] In order to ensure the period contains all tasks, the min/max must be computed at the very end of processing.
- [ ] The processing includes star/end date computation for all the tasks. This is not optionnal: at the end of processing, every task has a start date, an end date and possibly a duration.

## Processing Rules

Note: these rules must be clearly identified in the source code. The goal of this implementation is not to be efficient in speed but to process data as expected. Speed consideration will be taken into account in a future development.

The original set of rules is found in the attached JS file: [helper2.js](https://github.com/user-attachments/files/31918878/helper2.js)

I expect this code to be imported "as is" in Tracim in order to ensure the rules are applied (this will also avoid to rewrite specs and to re-implement it). 

For traceability, here are the main strategy and rules (taken from helper2.js source code):

```
 * STRATEGY (current scope — forward-only):
 * - depends = predecessors: a card can't start before ALL its dependencies end.
 *   With multiple dependencies, the constraint is start >= max(end of each dependency).
 * - Dates only ever flow forward through the dependency graph (from predecessors
 *   to dependents). A card's own start/end/duration is never inferred from
 *   something that depends on it (no backward pass yet).
 * - Circular dependencies and references to unknown ids are reported as errors,
 *   not thrown — the resolver still returns whatever it could resolve.
 *
 * RESOLUTION RULES (per card, in this priority order):
 * 1. No duration given -> defaults to 1 (day).
 * 1b. Duration is an INCLUSIVE day count, not an offset: duration 1 (or
 *     less) means kickoff === deadline (a same-day task). Duration 3 spans
 *     3 calendar days total (start + 2 more), or 3 business days when
 *     ignoreWeekend is set.
 * 2. Explicit kickoff present -> that's the start, UNLESS it's earlier than
 *    the dependency end date, in which case the start is recomputed to the
 *    dependency end date. If the explicit kickoff is not earlier, it's kept
 *    exactly as given.
 * 3. No explicit kickoff, but an explicit deadline is present -> start is
 *    computed by walking the duration back from the deadline. If that
 *    computed start is on or after the dependency end date, it's kept. If
 *    it's earlier, BOTH start and end are recomputed from the dependency
 *    end date instead.
 * 4. No explicit kickoff, no explicit deadline, but the card has dependencies
 *    -> start = dependency end date.
 * 5. No explicit kickoff, no explicit deadline, no dependencies at all ->
 *    start = today. (This is a last-resort fallback, not an override — a
 *    dependency-less card with an explicit kickoff or deadline still uses
 *    rule 2 or 3 above; "today" only fires when there's truly nothing else
 *    to anchor on.)
 * - Original input values are always preserved on `kickoff_original` /
 *   `deadline_original` before kickoff/deadline are overwritten with the
 *   computed result, whether or not they actually changed.
 * - `ignoreWeekend` option (passed to computeCardDates / resolveGanttCards):
 *   when true, Saturdays and Sundays are skipped whenever duration is used
 *   to compute a date (both forward, start->end, and backward, end->start),
 *   AND no resolved start or end date is ever allowed to land on a weekend
 *   at all — a date that would fall on Saturday/Sunday (whether explicit,
 *   dependency-derived, or the "today" fallback) is pushed forward to the
 *   following Monday before it's used.
 */
```

### refactor in order to clarify naming between kanban and gantt:
 
- [ ] kanban contains columns and columns contain cards
- [ ] gantt contains projects and projects contain tasks

### Logging for debuging and understanding purpose

Log all processing rules in console.log. They do not need to be in the user interface but we must be able to filter lgos from console. Eg a keywork like `KanbanToGantt` should allow to filter all transformation done to data.

So every computation must show as a log line:

```
KanbanToGantt: Card f97b7995-99b0-4a2e-862d-af3f2f2f0bff [Ticket 123] - Apply RULE xxx : set kickoff date to 2026-08-31
KanbanToGantt: Card f97b7995-99b0-4a2e-862d-af3f2f2f0bff [Ticket 123] - Apply RULE yyy : set deadline to 2026-08-31
```

This will allow to debug without to do step-to-step debugging

### Compute colors after all date computation

Because dependencies and weekend may modify start/end dates, the color of a task must be computed after all start/end/duration stuff.

- [ ] compute color at the very end of the preprocessing

### Weekend days

- [ ] Ensure weekend days are ignored even if they are in the middle of a task. This should be taken into account by the claude-generated code.

### Improve user experience

Currently, we need to have focus on the header in order to scroll using the arrows. This is not easy to use.

- [ ] add a scrollbar if possible so it is fast to horizontally scroll. If not possible, then we need to have a talk in order to decide what to implement
- [ ] optionnal: add left/right buttons to scroll 1 week or 1 month

alubert-algoo:
Refs #6955


frontend_app_kanban/src/gantt.ts
```diff
@@ -0,0 +1,336 @@
...
+
+/* INFO - A.L - 2026-09-15 - These two functions are wrapper around the
+   date-fns functions. If we wants to use custom weekend days, these
+   functions must be altered to calc the delta based on these custom days. */
+const add = (date: Date, days: number, excludeWeekendDays: boolean): Date => {
+  return excludeWeekendDays ? addBusinessDays(date, days) : addDays(date, days)
+}
+const sub = (date: Date, days: number, excludeWeekendDays: boolean): Date => {
+  return excludeWeekendDays ? subBusinessDays(date, days) : subDays(date, days)
+}
```
alubert-algoo, line 110: @lebouquetin This is where the weekend management is currently done.

The `weekendDays` option is not currently used, since this value cannot be set in date-fns and frappe-gantt:
- https://github.com/date-fns/date-fns/blob/main/pkgs/core/src/addBusinessDays/index.ts
- https://github.com/frappe/gantt/blob/master/src/index.js#L126

If we wants to manage custom weekend days, the `{add,sub}BusinessDays` functions must be replaced.

lebouquetin: I suggest to put both links in your comment so the source code includes the full picture on "what are the limitations and how to override them in case we want weekends not on saturday/sunday @alubert-algoo

alubert-algoo: I have put these links in the latest commit.

frontend_app_kanban/package.json
```diff
@@ -65,11 +66,21 @@
     "style-loader": "^0.23.1",
     "stylus": "^0.54.5",
     "stylus-native-loader": "^1.1.0",
+    "ts-loader": "^9.5.2",
+    "ts-standard": "^12.0.2",
+    "typescript": "^5.8.3",
```
piRGoif, line 71: Shouldn't this be added in the root package.json so that we can have Typescript in any app ?

frontend_app_kanban/tsconfig.json
```diff
@@ -0,0 +1,25 @@
+{
```
piRGoif, line 1: same comment, shouldn't this file be added at the root level ?

frontend_app_kanban/webpack.standalone.config.js
```diff
@@ -24,12 +24,13 @@ module.exports = {
       use: 'standard-loader',
       exclude: [/node_modules/, /frontend_lib/]
     }, {
-      test: [/\.js$/, /\.jsx$/],
+      test: [/\.[tj]s$/, /\.[tj]sx$/],
```
piRGoif, line 27: same comment as for package.json for adding inside kanban app VS globally

frontend_app_kanban/test/.mocharc.js
```diff
@@ -1,5 +1,9 @@
 'use strict'
 
+require('@babel/register')({
+  extensions: ['.js', '.jsx', '.ts', '.tsx']
+})
```
piRGoif, line 5: Same TS comment : add it globally instead ?

frontend_app_kanban/src/helper.js
```diff
@@ -34,89 +32,33 @@ export const localStorageFieldIdBuilder = (entryId) => {
...
-
-  // Ensure to use the beginning and the end of the day to have section
-  // using the full width of the available tasks.
-  if (minStart !== undefined) minStart.setHours(0, 0, 0)
-  if (maxEnd !== undefined) maxEnd.setHours(23, 59, 59)
 
-  return [minStart, maxEnd]
-}
+      if (!done[item.id]) done[item.id] = []
+      done[item.id].push(item.parent)
```
piRGoif:
```suggestion
      if (!done[item.id]) { done[item.id] = []}
      
      done[item.id].push(item.parent)
```

frontend_app_kanban/src/helper.js
```diff
@@ -34,89 +32,33 @@ export const localStorageFieldIdBuilder = (entryId) => {
...
-          minStart = kickoff
-        }
-      }
-
-      if (card.deadline?.length > 0) {
-        const deadline = new Date(card.deadline)
-        if (maxEnd === undefined || deadline > maxEnd) {
-          maxEnd = deadline
-        }
+export const nested = (items, id = null, done = {}) => {
```
piRGoif: Function name is not very self speaking ? Either change name or add a JSDoc ? 

Also the param names are very generic. May we rename them like:
 items -> tasks
id -> taskId

frontend_app_kanban/src/component/KanbanGantt.styl
```diff
@@ -12,12 +12,16 @@ barRadius = 3px
   &__left__top
     min-height auto
 
+.kanban__contentpage__wrapper__board
+  & > div
+    height 100% !important
```
piRGoif, line 17: `!important` in CSS are breaking cascade :( Can we get rid of those present in this file ? Maybe by working on selectors specificity ?

frontend_app_kanban/src/component/KanbanGantt.jsx
```diff
@@ -88,19 +90,9 @@ export class KanbanGantt extends React.Component {
...
-    console.debug('%c<KanbanGantt> start second preprocessing', 'color: chartreuse')
-    const computedBar = computeDependenciesFromGantt(bars, state.dependencies)
-    console.debug('%c<KanbanGantt> end second preprocessing', 'color: chartreuse', computedBar)
-
-    // INFO - A.L - 2026-09-03 - Only use the bar with the start and end dates
-    // correctly specified to have a working Gantt.
-    return computedBar.filter((bar) => bar.start && bar.end)
+    let projects = getTasksListFromKanbanCards(this.props.columns, defaultOptions)
+    projects = applyBusinessRulesToProjects(projects, defaultOptions)
+    return convertTasksListToGantt(projects)
```
piRGoif, line 95: I really like those lines: it is simple, nice ! And having the corresponding code in a TS file was a perfect move, congrats !

frontend_app_kanban/src/gantt.ts
```diff
@@ -0,0 +1,353 @@
...
+
+interface Options {
+  excludeWeekendDays: boolean
+  weekendDays: number[]
+}
+
+export const colors = {
+  FINISHED_TASK: '#C0DD97',
+  OVERDUE_TASK: '#FFF1F1',
+  OVERDUE_TASK_OVERLAY: '#F7C1C1'
```
piRGoif, line 87: Shouldn't the value be moved to CSS vars ?

frontend_app_kanban/src/gantt.ts
```diff
@@ -0,0 +1,353 @@
...
+
+/* INFO - A.L - 2026-09-15 - These two functions are wrapper around the
+   date-fns functions. If we wants to use custom weekend days, these
+   functions must be altered to calc the delta based on these custom days.
+   ---
+   The two projects date-fns and frappe-gantt have the saturday/sunday weekend
+   days hard-coded in their sources:
+   - https://github.com/date-fns/date-fns/blob/main/pkgs/core/src/addBusinessDays/index.ts
+   - https://github.com/frappe/gantt/blob/master/src/index.js#L126
+*/
```
piRGoif: Clear comment, congrats

frontend_app_kanban/src/gantt.ts
```diff
@@ -0,0 +1,353 @@
...
+const add = (date: Date, days: number, excludeWeekendDays: boolean): Date => {
+  return excludeWeekendDays ? addBusinessDays(date, days) : addDays(date, days)
+}
+const sub = (date: Date, days: number, excludeWeekendDays: boolean): Date => {
+  return excludeWeekendDays ? subBusinessDays(date, days) : subDays(date, days)
+}
+
+/* Apply the rules used to represent the tasks in the Gantt view */
```
piRGoif: Use the [JSDoc format](https://jsdoc.app/about-getting-started) instead ? 

```suggestion
/** 
 * Apply the rules used to represent the tasks in the Gantt view 
 */
```

frontend_app_kanban/src/gantt.ts
```diff
@@ -0,0 +1,353 @@
...
+  duration: number
+  progress: number
+  finished: boolean
+  start: Date
+  end: Date
+  depends: string[]
+  _card: KanbanCard
+}
+
+interface Options {
```
piRGoif, line 79: This is quite a generic naming. Maybe replace with TaskCalendarOptions ?

frontend_app_kanban/src/gantt.ts
```diff
@@ -0,0 +1,353 @@
...
+
+      recursiveDependencies(task.depends, dependencies)
+        .filter((id) => id !== task.id)
+        .forEach((id) => {
+          const parentTask = tasksById[id]
+          if (parentTask && (!start || parentTask.end > start)) {
+            const parentTaskEnd = parentTask.end
+            parentTaskEnd.setHours(0, 0, 0)
+
+            // Move the task just after the found parent
```
piRGoif: I really dislike inline comments as they are hard to maintain in the long run
Here I found the code is self speaking so I would remove this line ?

frontend_app_kanban/src/gantt.ts
```diff
@@ -0,0 +1,353 @@
...
+    project.tasks = project.tasks
+      .map((task: Task) => {
+        return tasksById[task.id] ? tasksById[task.id] : task
+      })
+      .sort((first: Task, second: Task) => {
+        if (first.start < second.start) return -1
+        if (second.start < first.start) return 1
+        return 0
+      })
+  })
```
piRGoif, line 169: This block can be extracted to a function with a naming reflecting what it does, that would reduce number of lines in this function and remove the need for an inline comment

frontend_app_kanban/src/gantt.ts
```diff
@@ -0,0 +1,353 @@
...
+
+  return projects
+}
+
+/* Convert all the projects as a structure usable by Frappe-Gantt */
```
piRGoif:
```suggestion
/**
 * Convert all the projects as a structure usable by Frappe-Gantt 
 */
```

frontend_app_kanban/src/gantt.ts
```diff
@@ -0,0 +1,353 @@
...
+      end: new Date(Date.now()),
+      color: project.color,
+      custom_class: 'gantt-section'
+    },
+    ...project.tasks.map((task: Task) => {
+      const [color, colorProgress] = getColorsForTask(task)
+
+      // INFO - A.L - 2026-09-15 - Since the card in the Kanban only use
+      // date and not datetime, the kickoff will start at midnight and the
+      // deadline at 23h59.
```
piRGoif, line 193: I would move this to the function JSDoc, as this would explain the function output (Date in input, Datetime in output because of frappe-gantt)

frontend_app_kanban/src/gantt.ts
```diff
@@ -0,0 +1,353 @@
...
+        duration: `${task.duration}d`,
+        dependencies: task.depends,
+        progress: task.finished ? 100 : task.progress,
+        _card: task._card
+      }
+    })
+  ]))
+}
+
+/* Retrieve all the dependencies available from the list of projects */
```
piRGoif:
```suggestion
/**
 *  Retrieve all the dependencies available from the list of projects 
 */
```

frontend_app_kanban/src/gantt.ts
```diff
@@ -0,0 +1,353 @@
...
+        }
+        if (!dependencies[dependId].includes(task.id)) {
+          dependencies[dependId].push(task.id)
+        }
+      })))
+
+  return dependencies
+}
+
+/* Retrieve the colors based on the status of the specified task */
```
piRGoif:
```suggestion
/**
 * Retrieve the colors based on the status of the specified task 
 */
```

frontend_app_kanban/src/gantt.ts
```diff
@@ -0,0 +1,353 @@
...
+export const getColorsForTask = (task: Task): string[] => {
+  if (task.finished || task.progress === 100) {
+    return [colors.FINISHED_TASK, colors.FINISHED_TASK]
+  } else if (!isToday(task.end) && isPast(task.end)) {
+    return [colors.OVERDUE_TASK, colors.OVERDUE_TASK_OVERLAY]
+  }
+  return ['', '']
+}
+
+/* Retrieve all the tasks available from the list of projects */
```
piRGoif:
```suggestion
/**
 * Retrieve all the tasks available from the list of projects 
 */
```

frontend_app_kanban/src/gantt.ts
```diff
@@ -0,0 +1,353 @@
...
+export const getTasksByIdentifier = (projects: Project[]): IdentifiedTasks => {
+  console.debug('%c<Gantt> retrieve all the tasks from the list of projects', 'color: chartreuse', projects)
+  const identifiers = {}
+  projects.forEach((project: Project) => project.tasks.forEach((task: Task) => {
+    identifiers[task.id] = task
+  }))
+  return identifiers
+}
+
+/* Convert a structure from react-kanban to a list of projects */
```
piRGoif:
```suggestion
/**
 * Convert a structure from react-kanban to a list of projects 
 */
```

frontend_app_kanban/src/gantt.ts
```diff
@@ -0,0 +1,353 @@
...
+          progress: parseInt(card.progress) || 0,
+          finished: card.finished,
+          _card: card
+        }
+      })
+    )
+  }))
+}
+
+/* Prepare the start and end dates */
```
piRGoif:
```suggestion
/**
 * Prepare the start and end dates 
 */
```

frontend_app_kanban/src/gantt.ts
```diff
@@ -0,0 +1,353 @@
...
+
+  // Avoid to having an ending date before the starting date
+  if (start > end) {
+    end = add(start, duration - 1, excludeWeekendDays)
+  }
+
+  return [start, end]
+}
+
+/* Sort the list of tasks by their dependencies */
```
piRGoif:
```suggestion
/**
 * Sort the list of tasks by their dependencies 
 */
```

frontend_app_kanban/src/gantt.ts
```diff
@@ -0,0 +1,353 @@
...
+const sub = (date: Date, days: number, excludeWeekendDays: boolean): Date => {
+  return excludeWeekendDays ? subBusinessDays(date, days) : subDays(date, days)
+}
+
+/* Apply the rules used to represent the tasks in the Gantt view */
+export const applyBusinessRulesToProjects = (
+  projects: Project[],
+  { excludeWeekendDays, weekendDays }: Options = defaultOptions
+): Project[] => {
+  console.debug('%c<Gantt> apply rules on all the projects', 'color: chartreuse')
```
piRGoif, line 119: May we add a wrapper for this file log, so that if we change our minds it would be easier to modify level and/or color, prefix etc ?

frontend_app_kanban/src/gantt.ts
piRGoif: It is not very clear why we have 2 values returned. Document what is returned like with a [`@returns` JS doc tag](https://jsdoc.app/tags-returns) ?

frontend_app_kanban/src/gantt.ts
```diff
@@ -0,0 +1,353 @@
...
+    end = add(start, duration - 1, excludeWeekendDays)
+    console.debug(
+      '<GanttRule> apply rule on %s “%s”: set dates as today', card.id, card.title
+    )
+  } else {
+    start = new Date(start)
+    end = new Date(end)
+  }
+
+  // Avoid to having an ending date before the starting date
```
piRGoif:
```suggestion
  // Avoid ending date value before the starting date value
```

frontend_app_kanban/src/gantt.ts
```diff
@@ -0,0 +1,353 @@
...
+  console.debug(
+    '%c<Gantt> sort the list of tasks by their dependencies', 'color: chartreuse', tasks
+  )
+  const tasksDependencies: IdentifiedDependency[] = []
+  tasks.sort((first: Task, second: Task): number => {
+    if (first.depends.length === 0 && second.depends.length > 0) return -1
+    if (second.depends.length === 0 && first.depends.length > 0) return 1
+    if (first.depends.includes(second.id)) return 1
+    if (second.depends.includes(first.id)) return -1
+    return 0
```
piRGoif, line 349:
```suggestion
    if (first.depends.length === 0 && second.depends.length > 0) { return -1}
    if (second.depends.length === 0 && first.depends.length > 0) { return 1}
    if (first.depends.includes(second.id)) { return 1}
    if (second.depends.includes(first.id)) { return -1}
    return 0
```

frontend_app_kanban/test/gantt.spec.js
```diff
@@ -0,0 +1,400 @@
+import assert from 'node:assert'
```
piRGoif, line 1: So cool you added tests ! Thanks !

piRGoif (changes requested):
Congrats for your wonderful work !!

I just have some suggestions:
- moving ts support to the root if possible so that any app can use it
- extracting some logic to wrappers or private functions
- use JSDoc more

frontend_app_kanban/package.json
alubert-algoo: If we do that, we must do it for the other application which used Typescript (the agenda for example). We can do it on another branch.

frontend_app_kanban/src/helper.js
alubert-algoo: This function will be removed in my next patch.

frontend_app_kanban/src/helper.js
alubert-algoo: This function will be removed in my next patch.

frontend_app_kanban/src/gantt.ts
alubert-algoo: Since these values must be set in the Gantt JSON, this is not possible without having to write multiple CSS class and override the Gantt generation process.

frontend_app_kanban/src/gantt.ts
alubert-algoo: Since this interface is not exportable, there is no problem with the current name, since it means `Options for Gantt.ts`.

frontend_app_kanban/src/gantt.ts
alubert-algoo: If we move this block in a dedicated function, it means this behavior can be use by another function in the future. But in this case, this block must only be used inside `applyBusinessRulesToProjects`, so there is no need to move it.

frontend_app_kanban/src/gantt.ts
alubert-algoo: Already tried, but it was not easy as I though (since the arguments are not the same).

Also, these line of debug will be removed in the future.

frontend_app_kanban/src/gantt.ts
alubert-algoo: If I put brackets, I will add new lines. The goal of not using the brackets in this case is to avoid new lines :)

frontend_app_kanban/src/component/KanbanGantt.styl
alubert-algoo: In this case, this class is share between the Kanban and the Gantt. Even with a subclass, the value will be override without the `!important` flag and the render will not be correct with Gantt.

frontend_app_kanban/src/component/KanbanGantt.jsx
```diff
@@ -88,19 +90,9 @@ export class KanbanGantt extends React.Component {
...
-    console.debug('%c<KanbanGantt> end first preprocessing', 'color: chartreuse', bars)
-
-    console.debug('%c<KanbanGantt> start second preprocessing', 'color: chartreuse')
-    const computedBar = computeDependenciesFromGantt(bars, state.dependencies)
-    console.debug('%c<KanbanGantt> end second preprocessing', 'color: chartreuse', computedBar)
-
-    // INFO - A.L - 2026-09-03 - Only use the bar with the start and end dates
-    // correctly specified to have a working Gantt.
-    return computedBar.filter((bar) => bar.start && bar.end)
+    let projects = getTasksListFromKanbanCards(this.props.columns, defaultOptions)
```
lebouquetin, line 93: the naming make things unclear: what is the structure returned by getTasksListFromKanbanCards() a list of tasks ? or a list of projects containing tasks?

frontend_app_kanban/package.json
piRGoif: ok I've added #6969. Closing conversation.

frontend_app_kanban/webpack.standalone.config.js
piRGoif: See answer above

frontend_app_kanban/tsconfig.json
piRGoif: See answer above

frontend_app_kanban/test/.mocharc.js
piRGoif: See answer above

frontend_app_kanban/src/component/KanbanGantt.styl
piRGoif: classes are low in the [specificity score](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Cascade/Specificity#how_is_specificity_calculated), I was more thinking of adding elements or id ? But ok let's keep it if it is complicated to change

frontend_app_kanban/src/gantt.ts
piRGoif: I was thinking of replacing `"#C0DD97"` by `var(--kanban_task_finished)` for example ? This is stil text that can be injected as is in the html/css

frontend_app_kanban/src/gantt.ts
piRGoif: ok, closing conversation

frontend_app_kanban/src/gantt.ts
piRGoif: ok
Not having brackets leads to awful bugs, I saw it multiple times in my career :(

frontend_app_kanban/src/gantt.ts
piRGoif: "Gantt Options" is still too generic in my opinion
Those 2 options are dedicated to task start/end computation

frontend_app_kanban/src/gantt.ts
```diff
@@ -0,0 +1,394 @@
...
+
+/**
+ * INFO - A.L - 2026-09-15 - These two functions are wrapper around the
+ * date-fns functions. If we wants to use custom weekend days, these
+ * functions must be altered to calc the delta based on these custom days.
+ * ---
+ * The two projects date-fns and frappe-gantt have the saturday/sunday weekend
+ * days hard-coded in their sources:
+ * - https://github.com/date-fns/date-fns/blob/main/pkgs/core/src/addBusinessDays/index.ts
+ * - https://github.com/frappe/gantt/blob/master/src/index.js#L126
```
piRGoif, line 103: Use the power of JSDoc O:)



```suggestion
 * Our dependencies have hard-coded weekend days (saturday/sunday):
 * @link https://github.com/date-fns/date-fns/blob/main/pkgs/core/src/addBusinessDays/index.ts date-fns code with hardcoded weekend days
 * @link https://github.com/frappe/gantt/blob/master/src/index.js#L126 frappe-gantt code with hardcoded weekend days
```

piRGoif (changes requested):
Just a small JSDoc suggestion

piRGoif (approved):
Discussed with Aurelien : we are merging for the build, and the remaining tasks will be done in the next days or so

frontend_app_kanban/src/component/KanbanGantt.jsx
alubert-algoo: This information is available in the `gantt.ts` file:
```ts
export const getTasksListFromKanbanCards = (
  kanban: KanbanColumn[],
  { excludeWeekendDays, weekendDays }: Options = defaultOptions
): Project[] => {
```

lebouquetin: So

- the naming do not match what it does (it do not return a list of tasks: it returns a list of projects),
- the implementation does not match the spec defined in the issue (it was asked to work on flat task lists)

This is not a problem to make things differently - but it need to be discussed before review.

PR #6965, opened 2026-09-16
fix: pin the version in the container pushpin

alubert-algoo:
This container will be used by the CI, so this version must be pinned too.

Refs #6961.


piRGoif:
The PR title doesn't reflect the modification right ? This is about minio and not pushpin ?

alubert-algoo:
@piRGoif Yes, this is no longer the correct title, since the Minio image is no longer available. I will move this PR as draft to avoid confusion.

